PR #16, opened 2025-01-01
Use explicit prototype for Fiber

mikearnaldi:


mikearnaldi:
using explicit prototype is up for discussion but I fear setting fields in a `static {}` block may hurt monomorphism

tim-smart:
We can merge this and see how it goes :)

PR #49, opened 2025-02-26
Introduce Named Scopes

mikearnaldi:
cc @IMax153 @tim-smart

### DIFF
--- a/src/Scope.ts
+++ b/src/Scope.ts
@@ -128,11 +128,13 @@ export const unsafeMake: (options?: { strategy?: "sequential" | "parallel" | und
  * @since 4.0.0
  * @category combinators
  */
-export const provide: {
-  <I>(tag: Context.Tag<I, Scope>): {
-    (value: Scope): <A, E, R>(self: Effect<A, E, R>) => Effect<A, E, Exclude<R, I>>
-    <A, E, R>(self: Effect<A, E, R>, value: Scope): Effect<A, E, Exclude<R, I>>
-  }
+export const provide: <I>(
+  tag: Context.Tag<I, Scope>
+) => {
+  <From>(value: Context.Tag<From, Scope>): <A, E, R>(self: Effect<A, E, R>) => Effect<A, E, Exclude<R, I>>
+  <A, E, R, From>(self: Effect<A, E, R>, value: Context.Tag<From, Scope>): Effect<A, E, Exclude<R, I>>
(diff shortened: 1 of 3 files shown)


IMax153:
I must say - I am a big fan of this change.

I particularly like the idea of being able to attach semantic meaning to scopes in the environment so you can actually understand what is likely being tracked by a scope. Especially because, at the moment, it can get a bit difficult to know what resources will be finalized when providing a `Scope` for a deeply nested program.

IMax153:
@tim-smart - I feel like this would also partially solve the footguns we were worried about in the other PR. Types that we expose can introduce named scopes into the environment (e.g. `HttpConnectionScope` or whatever).

mikearnaldi:
> @tim-smart - I feel like this would also partially solve the footguns we were worried about in the other PR. Types that we expose can introduce named scopes into the environment (e.g. `HttpConnectionScope` or whatever).

yeah that's exactly the use case, things that should be locally erased should be explicit, and with this doing `Scope.scoped(HttpConnectionScope)(effect)` (maybe aliased to `HttpClient.scopedConnection`) won't close anything else

mattiamanzati:
LOVE THIS!

Also ties to the argument that would be really nice in the future to emit named spans for both scope life (as of now) and for time taken to shutdown. This is definetely a step towards having more control over that by having different scope names.

mikearnaldi:
> LOVE THIS!
> 
> Also ties to the argument that would be really nice in the future to emit named spans for both scope life (as of now) and for time taken to shutdown. This is definetely a step towards having more control over that by having different scope names.

not sure if this goes in the same direction (also currently no span is emitted, even for "scope life"), Named scopes are plain scopes they just have a different tag so per se they don't add tracing. Imho tracing should be done in user-land adding spans to finalizers

tim-smart:
After the conf and cluster it would be good to see how these scope changes feel in a project. Maybe start bringing in the rest of the monorepo.

mikearnaldi:
> After the conf and cluster it would be good to see how these scope changes feel in a project. Maybe start bringing in the rest of the monorepo.

Named scopes are for locally erased resources, like the HttpConnection on a fetch call, not for resources in general, you wouldn't see many explicit scopes in a user project

mikearnaldi:
After a lot of thinking I am no longer sure this reduces complexity, the key issue is that two different scopes will close independently and resources allocated in one and the other may be interdependent, leading to potential incorrect closing order

mikearnaldi:
Closing due to the issue described, after a lot of testing I do agree with @tim-smart and @IMax153 on explicit scope

PR #50, opened 2025-03-04
Experiment making STM a plain Effect

mikearnaldi:


tim-smart:
Had a play around here https://github.com/Effect-TS/effect-smol/pull/51

Seems pretty good, but if we compare to STM the only difference is the allowance of async transactions. I don't think that is an issue though.

mikearnaldi:
> Had a play around here #51
> 
> Seems pretty good, but if we compare to STM the only difference is the allowance of async transactions. I don't think that is an issue though.

indeed, but it's actually one of the reasons I decided to try, I wanted to have async transactions

PR #77, opened 2025-04-01
try rewriteRelativeImportExtensions

tim-smart:
This allows importing directly from github in deno


mikearnaldi:
Importing from github would not be a good idea as it would end up generating duplicate packages when using libraries that import from npm. I think the Deno story needs to be forcing users to use "npm:..."

lishaduck:
Using `rewriteRelativeImportExtensions` allows using Node's native TS support as well as Deno's (and would make #80 slightly easier).

I do agree that `npm:` is much better than using `esm.sh/gh/`, that'd lead to lots of duplication.

fubhy:
Implemented here with some other recent improvements: https://github.com/Effect-TS/effect-smol/pull/316

fubhy:
@lishaduck Is there anything in there now that would simlify / improve the Deno setup based on this?

lishaduck:
> @lishaduck Is there anything in there now that would simlify / improve the Deno setup based on this?

You can drop `sloppy-imports` from this:

https://github.com/Effect-TS/effect-smol/blob/e3181e8c199e8d26aec3fe7ee1f1287299bdf542/deno.json#L3

Oh, and if you're in the mood for another tedious but semi-automated PR, it'd be a lot simpler to enable `noImplictOverride` instead of fragilely hacking through Deno's tsconfig resolution bugs.

fubhy:
@lishaduck I've done so here: https://github.com/Effect-TS/effect-smol/pull/327 

I've also enforced this in our general typescript base config and improved the exclude patterns a bit further.

lishaduck:
> @lishaduck I've done so here: #327
> 
> I've also enforced this in our general typescript base config and improved the exclude patterns a bit further.

Awesome! At this point, I think the only thing left is to add `@effect/platform-deno` (which I think is doable now)!

PR #80, opened 2025-04-13
Deno Support

### LINKED ISSUE #79: Deno Support
### What is the problem this feature would solve?

For those of us crazy people who use Deno, we'd very much like Deno support to be built-in in Effect v4.
Monorepo support, npm compat, and BYONM is much more mature than when https://github.com/Effect-TS/effect/issues/1806 was filed.

### What is the feature you are proposing to solve the problem?

- Add Deno to test matrix from start (I've got Vitest working for everything except code coverage, and there seems to be some progress on that front).
- Add in `@effect/platform-deno`
- Clearly communicate that Effect is meant to be used via `npm:` specifiers, and won't be published to `jsr:`
  (Could `@effect/platform-deno` be on JSR though, so it can depend on `@std`?)

### What alternatives have you considered?

Keep (half) maintaining [`effect-platform-deno`](https://github.com/lishaduck/effect-utils).

lishaduck:
## Type

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add in support for running the test suite and typechecking with Deno.

## Related

- Closes #79.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,0 +1,29 @@
+{
+    "nodeModulesDir": "manual",
+    "unstable": [
+        "sloppy-imports",
+        "bare-node-builtins",
+        "node-globals"
+    ],
+    "workspace": [
+        "./packages/*"
+    ],
+    "compilerOptions": {
+        "noImplicitOverride": false
+    },
+    "exclude": [
+        "**/node_modules/",
+        "**/.tsbuildinfo/",
+        "**/build/",
+        "**/dist/",
+        "**/bundle/",
+        "scratchpad/",
+        "scripts/",
+        "packages/effect/dtslint/**/*.ts",
+        "packages/platform-bun/",
+        "packages/platform-node/",
+        "packages/platform-node-shared/",
+        "test-schemapart.mjs",
+        "vitest.*.ts"
+    ]
+}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -26,6 +26,20 @@ jobs:
         uses: ./.github/actions/setup
       - run: pnpm check
 
+  types-deno:
+    name: Types on Deno
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install deno
+        uses: denoland/setup-deno@v2
+      - name: Install dependencies
+        uses: ./.github/actions/setup
+      - run: deno check
+
   bundle:
     name: Bundle
     if: github.event_name == 'pull_request'
@@ -104,6 +118,21 @@ jobs:
         run: pnpm vitest
         env:
           NODE_OPTIONS: --max_old_space_size=8192
+
+  test-deno:
+    name: Test on Deno
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install deno
+        uses: denoland/setup-deno@v2
+      - name: Install dependencies
+        uses: ./.github/actions/setup
+      - name: Test
+        run: deno task test
 
   jsdoc-analysis:
     name: JSDoc Analysis
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -42,5 +42,7 @@
   "editor.suggestSelection": "recentlyUsed",
   "editor.wordBasedSuggestions": "matchingDocuments",
   "editor.parameterHints.enabled": true,
-  "files.insertFinalNewline": true
+  "files.insertFinalNewline": true,
+  "deno.enable": true,
+  "deno.lint": false
 }
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@rollup/plugin-node-resolve": "^16.0.1",
     "@rollup/plugin-replace": "^6.0.2",
     "@rollup/plugin-terser": "^0.4.4",
+    "@types/node": "^24.0.12",
     "@types/jscodeshift": "^17.3.0",
     "@vitest/browser": "^3.2.4",
     "@vitest/coverage-v8": "^3.2.4",
--- a/packages/effect/src/Cron.ts
+++ b/packages/effect/src/Cron.ts
@@ -934,18 +934,18 @@ const parseSegment = (
 }
 
 const splitStep = (input: string): [string, number | undefined] => {
-  const seperator = input.indexOf("/")
-  if (seperator !== -1) {
-    return [input.slice(0, seperator), Number(input.slice(seperator + 1))]
+  const separator = input.indexOf("/")
+  if (separator !== -1) {
+    return [input.slice(0, separator), Number(input.slice(separator + 1))]
   }
 
   return [input, undefined]
 }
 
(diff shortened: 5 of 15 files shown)


deno.json
```diff
@@ -0,0 +1,40 @@
+{
+    "nodeModulesDir": "manual",
+    "unstable": [
+        "sloppy-imports"
```
lishaduck, line 4: Alternatively, #77 would obviate this.
This only **effect**s this internally, the built version is fine.

Except for the stack overflow.

deno.json
```diff
@@ -0,0 +1,40 @@
...
+    "compilerOptions": {
+        "strict": true,
+        "exactOptionalPropertyTypes": true,
+        "lib": [
+            "ESNext",
+            "DOM",
+            "DOM.Iterable"
+        ],
+        "types": [
+            "node"
```
lishaduck, line 15: Ehh. I couldn't get it admit that `@types/node` existed, so I vendored it.

src/internal/tracer.ts
```diff
@@ -1,5 +1,11 @@
 import type * as Tracer from "../Tracer.js"
 
+declare global {
+  class ErrorConstructor {
+    stackTraceLimit: number
+  }
+}
```
lishaduck, line 7: ↑ here

src/Cron.ts
```diff
@@ -682,18 +682,18 @@ const parseSegment = (
 }
 
 const splitStep = (input: string): [string, number | undefined] => {
-  const seperator = input.indexOf("/")
-  if (seperator !== -1) {
-    return [input.slice(0, seperator), Number(input.slice(seperator + 1))]
+  const separator = input.indexOf("/")
+  if (separator !== -1) {
+    return [input.slice(0, separator), Number(input.slice(separator + 1))]
```
lishaduck, line 687: Unrelated. Feel free to ask me to revert it.

src/internal/core.ts
```diff
@@ -504,7 +504,7 @@ export const YieldableError: new(
   message?: string,
   options?: ErrorOptions
 ) => Cause.YieldableError = (function() {
-  class YieldableError extends globalThis.Error {
+  class YieldableError extends (globalThis.Error as any) {
```
lishaduck, line 507: Otherwise it fails with `'YieldableError' does not override 'message'` or something like that.

lishaduck:
@mikearnaldi, is this non-disruptive enough?

lishaduck:
@IMax153, CI is failing due to the bundle size commenter, because `pull_request` is running without secrets. I'd advise _against_ switching to `pull_request_trigger` (due to security reasons), instead you can either disable it when running on forks (suboptimal), or [use the `workflow_run` trigger](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/).

I'd be happy to PR to fix it, with the normal CI caveat of "I can't test it, hopefully it works."

pixeleet:
https://deno.com/blog/v2.3#local-npm-packages
So this landed, but I haven't had the opportunity to try yet

lishaduck:
Works for monorepo as well with no more overhead than \*cough\* project references \*cough\*.

lishaduck:
I'm spinning up a PR to fix the comment-to-fork issue real quick, then I think this is good.

packages/platform-node-shared/test/NodeFileSystem.test.ts
```diff
@@ -64,7 +66,7 @@ describe("FileSystem", () => {
       expect(after).toEqual("")
     })))
 
-  it("should track the cursor position when reading", () =>
+  it.skipIf(isDeno)("should track the cursor position when reading", () =>
```
tim-smart, line 69: What's the issue here?

lishaduck: The test fails. It's a bug in Deno, we hit it in effect-platform-deno as well: https://github.com/lishaduck/effect-utils/blob/eb3bdb3b35b6ca6fe448472e37c4440df5be7bb3/packages/platform-deno/tests/DenoFileSystem.test.ts#L75-L77

I'm 99% sure it's a bug in Deno, but I didn't see any open issues the last time I checked and I didn't understand the issue well enough to file one.

deno.json
```diff
@@ -25,17 +25,15 @@
...
         "**/node_modules/",
         "**/build/",
         "**/dist/",
         "**/bundle/",
         "scratchpad/",
         "scripts/",
-        "vitest.setup.ts"
+        "vitest.setup.ts",
+        "packages/effect/test/Schema.test.ts",
+        "packages/effect/test/SchemaToSerializer.test.ts"
```
lishaduck, line 37: Unlike `pnpm check`, both tsserver and `deno check` error with this:

```
TS2417 [ERROR]: Class static side 'typeof MyError' incorrectly extends base class static side 'typeof ErrorConstructor'.
  Types of property 'prototype' are incompatible.
    Property 'message' is missing in type 'ErrorConstructor' but required in type 'Error'.
      class MyError extends Error {}
```

deno.json
lishaduck: Oh, I did eventually figure out _why_ this wasn't working—Deno's loading this lazily, so it requires an explicit import of `node:*` to add the global augmentations for those modules, and then it doesn't let you import node:globals (b/c it doesn't exist), so, as far as I could tell, you can't get `@types/node` to work for globals.

lishaduck:
@gcanti, you might want to CP 228a710fd920cd37e177326b6cdeb8d04615c3c4, that's unrelated to Deno (beyond that I didn't turn off `deno check` for the bench directory.)

lishaduck:
Node globals are still weird (going with @ts-ignore), can't convince it to turn off `noImplictOverride` with `references`, and I still can't figure out the weird Schema Error errors... but otherwise good :)
Oh, and the LSP is broken. Oh, in other news, it stopped segfaulting! (remind me why I'm doing this... jk)

deno.json
lishaduck: Still happening, not sure why.

lishaduck: 🤦‍♂️, forgot I ignored these. Still not sure why it's erroring (some default .d.ts differences methinks), but test typechecks aren't a complete blocker.

packages/platform-node-shared/test/NodeFileSystem.test.ts
lishaduck: (Still need to fix tests not to run, only did tsc)

lishaduck: Fixed it! (https://github.com/vitest-dev/vitest/issues/8280)

deno.json
lishaduck: I did eventually find this in the docs: https://docs.deno.com/runtime/reference/ts_config_migration/#type-declaration-semantics:~:text=It%20is%20unpredictable%20how%20Deno%20will%20handle%20ambient/global%20declarations.

lishaduck:
@tim-smart, this passes checks locally. Does it address your concerns?

lishaduck:
🤞 this time...

~~(Need to check it still works after rebase, and remember to reinstall pkgs this time lol..., but it did work)~~ EDIT 2: Ok, now it works... I hope 😅

tim-smart (approved):

PR #92, opened 2025-05-02
Schema v4

gcanti:
More details here: https://github.com/Effect-TS/effect-smol/blob/schema-v4/packages/effect/SCHEMA.md


gcanti:
going to merge this and start sending smaller PRs now that the schema modules are being used in other parts of the codebase

PR #149, opened 2025-06-04
Work on fn

mikearnaldi:


mikearnaldi:
looks like we can't really do anything about compacting pipe... this is unsafe on `testInference` it is supposed to raise an error

PR #181, opened 2025-06-26
backport from effect

tim-smart:


packages/effect/src/Cause.ts
```diff
@@ -45,6 +45,24 @@ export const isCause: (self: unknown) => self is Cause<unknown> = core.isCause
  */
 export type Failure<E> = Fail<E> | Die | Interrupt
 
+/**
+ * @since 4.0.0
+ * @category guards
+ */
+export const failureIsFail: <E>(self: Failure<E>) => self is Fail<E> = core.failureIsFail
```
gcanti, line 52: maybe we can just use `isFail` / `isDie` / `isInterrupt` here?

PR #207, opened 2025-06-30
Redacted: the ability to override the default label was added

KhraksMamtsov:
+ port tests

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

```ts
const redacted = Redacted.make("redacted", "MY_LABEL")
strictEqual(Redacted.label(redacted), "MY_LABEL")
strictEqual(redacted.toString(), "<MY_LABEL>")
strictEqual(JSON.stringify(redacted), "\"<MY_LABEL>\"")

assertTrue(Redacted.unsafeWipe(redacted))
throws(() => Redacted.value(redacted), new Error("Unable to get redacted value with label: \"MY_LABEL\""))
```

I considered adding a label `field` to the `interface Redacted { }` — but it might be confusing.
For example, it could give the impression that `Redacted.getEquivalence` compares labels, which it doesn't.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/chilly-doors-go.md
+++ b/.changeset/chilly-doors-go.md
@@ -1,5 +1,0 @@
----
-"effect": patch
----
-
-In the Redacted module, the ability to override the default label was added.
--- a/.changeset/chilly-doors-go.md
+++ b/.changeset/chilly-doors-go.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+In the Redacted module, the ability to override the default label was added.
--- a/packages/effect/src/Redacted.ts
+++ b/packages/effect/src/Redacted.ts
@@ -92,10 +92,10 @@ const Proto = {
   label: undefined,
   ...PipeInspectableProto,
   toJSON() {
-    return this.label ?? "<redacted>"
+    return `<${this.label ?? "redacted"}>`
   },
   toString() {
-    return this.label ?? "<redacted>"
+    return `<${this.label ?? "redacted"}>`
   },
   [Hash.symbol]<T>(this: Redacted<T>): number {
     return Hash.cached(this, () => Hash.hash(redactedRegistry.get(this)))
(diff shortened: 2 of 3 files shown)


KhraksMamtsov:
@mikearnaldi, could you share what you think about this?
It seems like it could be useful in error messages to better understand exactly what we're dealing with. :)
If needed, I can backport this and support it in the Schema and Config.

.changeset/chilly-doors-go.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+In the Redacted module, the ability to override the default label was added.
```
tim-smart, line 5: Don't need changesets at this stage

packages/effect/src/Redacted.ts
```diff
@@ -120,10 +127,28 @@ export const value = <T>(self: Redacted<T>): T => {
   if (redactedRegistry.has(self)) {
     return redactedRegistry.get(self)
   } else {
-    throw new Error("Unable to get redacted value")
+    throw new Error(`Unable to get redacted value with label: "${label(self)}"`)
```
tim-smart, line 130: I think this should only be the output if a custom label was provided.

packages/effect/src/Redacted.ts
```diff
@@ -120,10 +127,28 @@ export const value = <T>(self: Redacted<T>): T => {
...
+ *
+ * const API_KEY = Redacted.make(123, "API_KEY")
+ *
+ * assert.equal(Redacted.label(API_KEY), "API_KEY")
+ * ```
+ *
+ * @since 3.3.0
+ * @category getters
+ */
+export const label = (self: Redacted<any>): string => (self as any).label
```
tim-smart, line 150: You can add the label to the Redacted interface

tim-smart (changes requested):

packages/effect/src/Redacted.ts
```diff
@@ -84,9 +89,13 @@ const Proto = {
   [TypeId]: {
     _A: (_: never) => _
   },
+  label: undefined,
   ...PipeInspectableProto,
   toJSON() {
-    return "<redacted>"
+    return this.label ?? "<redacted>"
```
tim-smart, line 95: I think we still want the <> around the label

PR #253, opened 2025-07-08
add Config module

tim-smart:


packages/effect/src/config/ConfigProvider.ts
```diff
@@ -0,0 +1,152 @@
...
+ * @since 4.0.0
+ * @category Constructors
+ */
+export const fromEnv = (options?: {
+  readonly pathDelimiter?: string | undefined
+  readonly environment?: Record<string, string | undefined> | undefined
+}): ConfigProvider => {
+  const env = options?.environment ?? {
+    ...process.env,
+    ...(import.meta as any).env
```
IMax153, line 92: Won't these fail if the globals don't exist?

IMax153 (commented):
I really like the new encoding, particularly how simple `ConfigProvider` is now! Nice work Tim

packages/effect/src/config/ConfigProvider.ts
lishaduck: 🤦‍♂️ yup

PR #269, opened 2025-07-11
Config.Redacted - use name as label by default

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #


tim-smart:
Not sure about this as it will potentially leak environment variable names

KhraksMamtsov:
> Not sure about this as it will potentially leak environment variable names

I also don't know how critical this is — it would definitely be convenient though.
Okay, in any case, this can be done now and more explicitly.
If there's a request later, we can always come back to it.

PR #328, opened 2025-07-18
add bun vitest

fubhy:
Still blocked by https://github.com/oven-sh/bun/issues/4145 but looks like that's finally happening

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -140,20 +140,20 @@ jobs:
       - name: Test
         run: deno task test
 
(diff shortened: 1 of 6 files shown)


.github/workflows/check.yml
```diff
@@ -38,6 +40,8 @@ jobs:
         uses: denoland/setup-deno@v2
```
lishaduck, line 40: I think this is redundant now?

fubhy: Yes indeed

PR #361, opened 2025-07-28
POC: add Config.schema API

gcanti:
## Goal

The goal is to provide a single API that leverages the Schema module for all validation and decoding needs.

## How it works

1. Reads the keys from the struct fields
2. Uses these keys to load the corresponding values (as strings) from the `ConfigProvider`
3. Collects all key/value pairs into a `Serializer.StringLeafJson` struct
4. Generates a deserializer schema using `Serializer.stringLeafJson`
5. Calls `ToParser.decodeUnknownEffect`

**Example**

```ts
import { Config, ConfigProvider } from "effect/config"
import { Schema } from "effect/schema"

const config = Config.schema({
  a: Schema.String,
  b: Schema.Finite,
  c: Schema.Int,
  d: Schema.BigInt,
  e: Schema.Trim
})

const r = Effect.runSyncExit(
  config.parse(
    ConfigProvider.fromEnv({ environment: { a: "a", b: "1.2", c: "1", d: "2", e: "  e  " } }).context()
  )
)

if (r._tag === "Success") {
  console.log(r.value)
} else {
  console.dir(r.cause, { depth: null })
}
// => { a: 'a', b: 1.2, c: 1, d: 2n, e: 'e' }
```

### DIFF
--- a/bundle/config-schema.ts
+++ b/bundle/config-schema.ts
@@ -1,0 +1,5 @@
+import * as Config from "#dist/effect/config/Config"
+import * as Effect from "#dist/effect/Effect"
+import * as Schema from "#dist/effect/schema/Schema"
+
+Effect.runFork(Config.schema({ ENV: Schema.String }).asEffect())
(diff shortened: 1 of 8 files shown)


gcanti:
Closing in favor of https://github.com/Effect-TS/effect-smol/pull/384

PR #426, opened 2025-08-22
chore: enable type tests

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR makes it possible to run the type tests again.

@gcanti Seems like the errors you saw were caused by the `rewriteRelativeImportExtensions` option. It works fine when `tsc` is compiling, but language service throws "This import path is unsafe to rewrite..." when it tries to handle `references`. TSTyche is using the language service API.

The solution could be to move `rewriteRelativeImportExtensions` (added in TS 5.7) and `erasableSyntaxOnly` (added in TS 5.8) to `tsconfig.src.json` files. Probably these are not so important for `tsconfig.test.json`. That way you can also test types targeting TypeScript `>=5.4`. The unknown options are out of the way.

Some test files import using `.ts` extension. To make that work, I enabled `allowImportingTsExtensions` (added in TS 5.0) to `tsconfig.test.json`.

---

Might be there is cleaner way to organise the TSConfig files, of course.

### DIFF
--- a/tstyche.config.json
+++ b/tstyche.config.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://tstyche.org/schemas/config.json",
   "checkSuppressedErrors": true,
+  "checkSourceFiles": false,
   "testFileMatch": ["packages/*/dtslint/**/*.tst.*"]
 }
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           node-version: latest
       - run: pnpm check
-      # - run: pnpm test-types --target '>=5.4'
+      - run: pnpm test-types --target '>=5.4'
 
   types-deno:
     name: Types on Deno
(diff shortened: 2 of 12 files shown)


tstyche.config.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "https://tstyche.org/schemas/config.json",
   "checkSuppressedErrors": true,
+  "checkSourceFiles": false,
```
mrazauskas, line 4: Much better performance when this is disabled. The next major will not have this option, the implementation is not good. It simply type-checks all files I the program. Each and every `.ts` file. That is already done by `tsc`. So.. That was bad idea ;D

mrazauskas:
Ah! Alternative solution could be to set `"tsconfig": "ignore"` in `tstyche.config.json`. In this case it would ignore TSConfig files and uses the [defaults](https://tstyche.org/project/compiler-options#default-compiler-options). These are rather strict. And the setup would be less fragile, perhaps?

PR #427, opened 2025-08-22
chore: update TSTyche config

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

An alternative to #426.

The `tsconfig: "ignore"` tells to TSTyche to use its [default compiler options](https://tstyche.org/project/compiler-options#default-compiler-options) instead of resolving the nearest `tsconfig.json`. The default are rather strict.

This way type testing would be independent from any options set in the TSConfig files. And that will always work between TypeScript versions.

## Related

- Closes #426

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           node-version: latest
       - run: pnpm check
-      # - run: pnpm test-types --target '>=5.4'
+      - run: pnpm test-types --target '>=5.4'
 
   types-deno:
     name: Types on Deno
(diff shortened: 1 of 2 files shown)


gcanti:
Using the default compiler options seems like a good idea :+1:

mrazauskas:
Yes, I also like this solution better. It does not limit other tolling anyhow. And that is good.

gcanti:
Thanks so much for the quick heads up

mrazauskas:
Welcome! I never saw that error before, learned something new.

By the way, locally this setup spins faster than #426. Running `pnpm test-types --target '>=5.4'` takes some 4 seconds less. That’s very good.

PR #456, opened 2025-08-31
[WIP] add memory file system

fubhy:
Just experimenting atm. Not sure yet where this will go or whether it's a good idea. Could be handy for testing and for browser use-cases.


nikelborm:
https://github.com/nounder/effect-memfs

PR #495, opened 2025-09-06
Structural Equality By Default

mikearnaldi:
## Summary

- Implement deep structural equality for plain objects and arrays by default
- Add comprehensive support for special numeric values (NaN, Infinity, -Infinity)  
- Introduce instance equality opt-out mechanism for performance-critical cases
- Maintain full backward compatibility with existing Equal/Hash implementations

This enhancement makes `Equal.equals()` work intuitively for plain JavaScript objects and arrays while preserving all existing behavior for custom Equal implementations and other data types.

---

The last version handles deep equality on custom classes, meaning most likely we can get rid of everything in the `Data` module


packages/effect/src/internal/equal.ts
```diff
@@ -0,0 +1,79 @@
...
+  ) {
+    return false
+  }
+
+  // Include all other objects for structural comparison
+  return true
+}
+
+/** @internal */
+export const instanceEqualityRegistry = new WeakMap<object | Function, true>()
```
gcanti, line 79:
```suggestion
export const instanceEqualityRegistry = new WeakMap<object, true>()
```

tim-smart:
There is some pretty big perf footguns if we do this, i.e. if you have structures with large Uint8Array's and don't implement Hash/Equal

mikearnaldi:
> There is some pretty big perf footguns if we do this, i.e. if you have structures with large Uint8Array's and don't implement Hash/Equal

depends if you want to consider them mutable or not, if you consider them mutable with `Equal.byReference` the perf is the same no?

mikearnaldi:
Merging this, we can iterate further down the line

PR #526, opened 2025-09-16
remove ShardManager

tim-smart:
- [x] Remove ShardManager
- [x] Add RunnerStorage
- [x] Add runner health check singleton


afonsomatos:
Incredible if this works!

PR #583, opened 2025-09-25
add xmlEncoder

gcanti:
`Serializer.xmlEncoder` lets you serialize values to XML.

**Example**

```ts
import { Option } from "effect/data"
import { Schema, Serializer } from "effect/schema"

  a: Schema.String,
  b: Schema.Array(Schema.NullOr(Schema.String)),
  c: Schema.Struct({
    d: Schema.Option(Schema.String),
    e: Schema.Date
  }),
  f: Schema.optional(Schema.String)
})

// const encoder: (t: {...}) => Effect<string, Schema.SchemaError, never>
const xmlEncoder = Serializer.xmlEncoder(schema)

console.log(
  Effect.runSync(
    xmlEncoder({
      a: "",
      b: ["bar", "baz", null],
      c: { d: Option.some("qux"), e: new Date("2021-01-01") },
      f: undefined
    })
  )
)
/*
<root>
  <a></a>
  <b>
    <item>bar</item>
    <item>baz</item>
    <item/>
  </b>
  <c>
    <d>
      <_tag>Some</_tag>
      <value>qux</value>
    </d>
    <e>2021-01-01T00:00:00.000Z</e>
  </c>
  <f/>
</root>
*/
```


packages/effect/src/schema/Serializer.ts
```diff
@@ -264,3 +265,152 @@ export const goEnsureArray = memoize((ast: AST.AST): AST.AST => {
   }
   return out
 })
+
+type XmlEncoderOptions = {
+  /** Root element name for the returned XML string. Default: "root" */
+  readonly rootName?: string | undefined
```
IMax153, line 271: Should we wrap the whole thing in the `<xml>` tag instead of `root`? i.e. `<?xml version="1.0" encoding="UTF-8"?>`

gcanti: The XML Declaration is a processing instruction, it's `<?xml ... ?>` and must be self-contained

```xml
<?xml version="1.0" encoding="UTF-8"?>  <!-- Declaration only -->
<root>                                   <!-- Document content starts here -->
  <child>content</child>
</root>
```

PR #589, opened 2025-09-26
feat: cli/prompt migration

voffie:
## Type

- [X] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds all the functionality of prompts from effect/cli. It is still missing some important parts, such as tests. Since this module is heavily dependent on the `Terminal` module, it can't be merged before it is implemented and updated within this module.

### DIFF
--- a/packages/effect/src/unstable/cli/Prompt.ts
+++ b/packages/effect/src/unstable/cli/Prompt.ts
@@ -1,0 +1,3140 @@
+/**
+ * @since 4.0.0
+ */
+
+import type { NoSuchElementError } from "../../Cause.ts"
+import * as Arr from "../../collections/Array.ts"
+import * as Data from "../../data/Data.ts"
+import * as Option from "../../data/Option.ts"
+import * as Redacted from "../../data/Redacted.ts"
+import * as Effect from "../../Effect.ts"
+import { dual, identity, pipe } from "../../Function.ts"
+import * as Pipeable from "../../interfaces/Pipeable.ts"
+import { YieldableProto } from "../../internal/core.ts"
+import * as Match from "../../match/Match.ts"
+import * as FileSystem from "../../platform/FileSystem.ts"
+import * as Path from "../../platform/Path.ts"
+import * as Terminal from "../../platform/Terminal.ts"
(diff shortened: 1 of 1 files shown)


packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -0,0 +1,3140 @@
...
+  return Effect.gen(function*() {
+    const fs = yield* FileSystem.FileSystem
+    const path = yield* Path.Path
+    const currentPath = yield* resolveCurrentPath(state.path, options)
+    const selectedPath = state.files[state.cursor]
+    const resolvedPath = path.resolve(currentPath, selectedPath)
+    const info = yield* Effect.orDie(fs.stat(resolvedPath))
+    if (info.type === "Directory") {
+      const files = yield* getFileList(resolvedPath, options)
+      const filesWithoutParent = files.filter((file) => file !== "..")
```
voffie, line 1999: This filter/comparison is wrong since the `getFileList` function returns a boolean array instead of a string array

packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -0,0 +1,3140 @@
...
+          Effect.flatMap(FileSystem.FileSystem, (fs) =>
+            // Ensure the user provided starting path exists
+            Effect.orDie(fs.exists(path)).pipe(
+              Effect.filterOrDieMessage(
+                identity,
+                `The provided starting path '${path}' does not exist`
+              ),
+              Effect.as(path)
+            ))
+      }),
```
voffie, line 1828: This entire flatMap has a few issues:
1. FileSystem.FileSystem can't be the first argument in flatMap due to `Argument of type 'Key<FileSystem, FileSystem>' is not assignable to parameter of type 'Effect<unknown, never, FileSystem>'. Property '[TypeId]' is missing in type 'Key<FileSystem, FileSystem>' but required in type 'Effect<unknown, never, FileSystem>'.`
2. `filterOrDieMessage()` does not exist currently. Either this is on purpose or it has not been migrated yet. I could include the filterOrDieMessage functionality if needed

PR #600, opened 2025-09-30
migrate terminal module

voffie:
This PR adds the current implementation of the Terminal module. I've updated it to fit with the new structure and moved everything into one file

### DIFF
--- a/packages/effect/src/testing/index.ts
+++ b/packages/effect/src/testing/index.ts
@@ -26,6 +26,8 @@ export * as TestClock from "./TestClock.ts"
 export * as TestConsole from "./TestConsole.ts"
 
 /**
+ * Testing utilities for Schema validation and assertions.
+ *
  * @since 4.0.0
  */
 export * as TestSchema from "./TestSchema.ts"


packages/effect/src/platform/Terminal.ts
```diff
@@ -0,0 +1,95 @@
...
+import type { PlatformError } from "./index.ts"
+import type * as Option from "../data/Option.ts"
+import * as Data from "../data/Data.ts"
+
+/**
+ * A `Terminal` represents a command-line interface which can read input from a
+ * user and display messages to a user.
+ *
+ * @since 4.0.0
+ * @category models
```
IMax153, line 16: Let's improve the categories for consistency so that things get nicely grouped in the output API documentation.

- `models` / `model` -> `models`
- `tag` -> `services`

We also may want to consider adding a `TypeId` to the `Terminal` interface like to do with `FileSystem`, so that we can also add a type guard (i.e. `isTerminal`). This would also necessitate us adding a `make` constructor to the model that automatically adds the `TypeId` (so we can keep the `TypeId` internal). See the `FileSystem` module for what I mean.

IMax153 (changes requested):

packages/effect/src/platform/Terminal.ts
```diff
@@ -0,0 +1,95 @@
+/**
+ * @since 4.0.0
+ */
+
+import { Effect, Scope, ServiceMap } from "../index.ts"
```
tim-smart, line 5: We avoid using the re-export files in library code, even if it is type only. So these should be `import * as Effect from "../Effect.ts" etc

tim-smart (approved):

PR #602, opened 2025-10-01
add Differ module

gcanti:
## Using the Differ Module for Type-Safe JSON Patches

The `Differ` module lets you compute and apply JSON Patch (RFC 6902) changes for any value described by a `Schema`. You give it a schema once, then use the returned differ to produce a patch from an old value to a new value, and to apply that patch.

**Example** (Compare two values and apply the patch)

```ts
import { Differ, Schema } from "effect/schema"

// Describe the shape of your data
  id: Schema.Number,
  name: Schema.String,
  price: Schema.Number
})

// Build a differ tied to the schema
const jsonPatchDiffer = Differ.makeJsonPatch(schema)

// Prepare two values to compare
const oldValue = schema.makeUnsafe({ id: 1, name: "a", price: 1 })
const newValue = schema.makeUnsafe({ id: 1, name: "b", price: 2 })

// Compute a JSON Patch document (an array of operations)
const jsonPatch = jsonPatchDiffer.diff(oldValue, newValue)
console.log(jsonPatch)
/*
[
  { op: 'replace', path: '/name', value: 'b' },
  { op: 'replace', path: '/price', value: 2 }
]
*/

// Apply the patch to the old value to get the new value
const patched = jsonPatchDiffer.patch(oldValue, jsonPatch)
console.log(patched)
// { id: 1, name: 'b', price: 2 }
```

### How it works

The idea is simple: if you have a `Schema` for a type `T`, you can serialize any `T` to JSON and back. That lets us compute and apply JSON Patch on the JSON view, while keeping the public API typed as `T`.

- `diff(oldValue, newValue)`
  1. Encode `oldValue: T` and `newValue: T` to JSON with the schema serializer.
  2. Compute a JSON Patch document between the two JSON values.
  3. Return that patch (an array of `"add" | "remove" | "replace"` operations).

- `patch(oldValue, patch)`
  1. Encode `oldValue: T` to JSON.
  2. Apply the JSON Patch to the JSON value.
  3. Decode the patched JSON back to `T` using the schema.

This approach keeps patches independent from TypeScript types and uses the schema as the guardrail when turning JSON back into `T`.

### DIFF
--- a/packages/effect/SCHEMA.md
+++ b/packages/effect/SCHEMA.md
@@ -4657,8 +4657,8 @@ const schema = Schema.Struct({
 const jsonPatchDiffer = Differ.makeJsonPatch(schema)
 
 // Prepare two values to compare
-const oldValue = schema.makeUnsafe({ id: 1, name: "a", price: 1 })
-const newValue = schema.makeUnsafe({ id: 1, name: "b", price: 2 })
+const oldValue = { id: 1, name: "a", price: 1 }
+const newValue = { id: 1, name: "b", price: 2 }
 
 // Compute a JSON Patch document (an array of operations)
 const jsonPatch = jsonPatchDiffer.diff(oldValue, newValue)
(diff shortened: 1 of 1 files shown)


packages/effect/src/optic/Optic.ts
```diff
@@ -368,8 +368,11 @@ function make(ast: AST.AST): any {
   }
 }
 
-function shallowCopy(s: object): any {
-  return Array.isArray(s) ? s.slice() : { ...s }
+/** @internal */
+export function cloneShallow<T>(x: T): T {
+  if (Array.isArray(x)) return x.slice() as T
+  if (typeof x === "object" && x !== null) return { ...x } as T
```
nikelborm, line 374: How will this play out in the case of `.prototype`? I think it's reasonable to preserve it here

gcanti: @nikelborm it's intended to be used with POJOs: https://github.com/Effect-TS/effect-smol/blob/f23760590b838e309a46338d9875bedfdd064c55/packages/effect/SCHEMA.md#problem

gcanti: > Or maybe some runtime check that would reject objects that have a prototype other than object or null?

sure

PR #613, opened 2025-10-05
Opticable trait

KhraksMamtsov:
# Conflicts:
#	packages/effect/src/optic/Optic.ts

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Hi! I’d like to propose an extension point like this — so that it would be possible to use optics to modify custom objects.

When traversing the AST, Optic checks whether an object implements this trait and uses the result of cloning with possible modification.

The trait implementation should ensure invariants.

Here’s an example of how the trait could be implemented for Option.
I hope this idea makes sense — it would be really cool :)

## Related

- Related Issue #
- Closes #


KhraksMamtsov:
@gcanti hi
could you please take a look

KhraksMamtsov:
The trait can also be implemented in the Data, Result, and other modules.

fubhy:
So... Like ... Optic cable? As in ... Fiber optic? You missed your chance here for the perfect naming.

KhraksMamtsov:
> So... Like ... Optic cable? As in ... Fiber optic? You missed your chance here for the perfect naming.

If you name it differently, it’ll work slower than the speed of light ;)
Nevertheless, I’m open to your creativity)

gcanti:
@KhraksMamtsov the idea behind `ToOptic.ts` is precisely to avoid both a trait-based system and an excess of built-in optics, enabling direct work with plain objects

KhraksMamtsov:
> @KhraksMamtsov the idea behind `ToOptic.ts` is precisely to avoid both a trait-based system and an excess of built-in optics, enabling direct work with plain objects

Sorry, I hadn’t looked at `ToOptic`.
Now I see that it’s meant to use optics together with the schema and its isomorphism in the POJO.
Yeah, that’s much more powerful.

Well then, I’m winding my "optic cable" back onto the spool...

PR #615, opened 2025-10-05
prompt migration

voffie:
This PR migrates cli/prompt from v3. It is a duplicate of #589 since it got auto-closed by #609 . I haven't added tests for the Prompt module since they rely on a lot of v3-specific code. I think it is better to do a PR that looks over all of the cli tests and updates them

### DIFF
--- a/packages/effect/src/unstable/cli/Prompt.ts
+++ b/packages/effect/src/unstable/cli/Prompt.ts
@@ -1,0 +1,3137 @@
+/**
+ * @since 4.0.0
+ */
+
+import type { NoSuchElementError } from "../../Cause.ts"
+import * as Arr from "../../collections/Array.ts"
+import * as Data from "../../data/Data.ts"
+import * as Option from "../../data/Option.ts"
+import * as Redacted from "../../data/Redacted.ts"
+import * as Effect from "../../Effect.ts"
+import { dual, identity, pipe } from "../../Function.ts"
+import * as Pipeable from "../../interfaces/Pipeable.ts"
+import { YieldableProto } from "../../internal/core.ts"
+import * as Match from "../../Match.ts"
+import * as FileSystem from "../../platform/FileSystem.ts"
+import * as Path from "../../platform/Path.ts"
+import * as Terminal from "../../platform/Terminal.ts"
+import * as EffectNumber from "../../Number.ts"
+import * as Queue from "../../Queue.ts"
+import * as Ansi from "./internal/ansi.ts"
+import type * as Primitive from "./Primitive.js"
+
+const TypeId = "~@effect/cli/Prompt"
+
+/**
+ * @since 4.0.0
+ * @category models
+ */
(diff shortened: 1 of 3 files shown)


packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -0,0 +1,3137 @@
...
+    filter: options.filter ?? (() => Effect.succeed(true))
+  }
+  const initialState: Effect.Effect<
+    FileState,
+    never,
+    Prompt.Environment
+  > = Effect.gen(function* () {
+    const path = Option.none<string>()
+    const currentPath = yield* resolveCurrentPath(path, opts)
+    const files = yield* getFileList(currentPath, opts)
```
voffie, line 731: Here is an issue where the FileState wants the file variable to be a list of strings. However, getFileList returns a boolean array. I haven't changed anything from v3.

packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -0,0 +1,3137 @@
...
+    const confirm = Confirm.Hide()
+    return { cursor: 0, files, path, confirm }
+  })
+  return custom(initialState, {
+    render: handleFileRender(opts),
```
voffie, line 736: All of the keys/functions within the custom block has issues. Render & Clear has issues related to the Environment type where it probably isn't using all of Path, FileSystem & Terminal.

Process is instead having issue due to the files being a boolean array instead of a string array

packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -0,0 +1,3137 @@
...
+    ...options
+  }
+  const initialState: NumberState = {
+    cursor: 0,
+    value: "",
+    error: Option.none()
+  }
+  return custom(initialState, {
+    render: handleRenderInteger(opts),
+    process: handleProcessInteger(opts),
```
voffie, line 833: Same issue here as with handleProcessFloat

packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -0,0 +1,3137 @@
...
+function resolveCurrentPath(
+  path: Option.Option<string>,
+  options: FileOptions
+): Effect.Effect<string, never, FileSystem.FileSystem> {
+  return Option.match(path, {
+    onNone: () =>
+      Option.match(options.startingPath, {
+        onNone: () => Effect.sync(() => process.cwd()),
+        onSome: (path) =>
+          Effect.flatMap(FileSystem.FileSystem, (fs) =>
```
voffie, line 1815: This entire flatMap has a few issues and I'm not sure on how to migrate this fully to v4

tim-smart:
@voffie back over to you :)

IMax153:
@voffie - looks like a few issues in CI need to be resolved

PR #627, opened 2025-10-07
feat: add Command prompt constructor

voffie:
This PR was a follow-up to #600 . It was adding the Terminal module to cli/Command since it was merged before the Terminal module was migrated.

EDIT: Due to changes to main, this branch has diverged into adding the prompt constructor to `Command` as it was decided upon in the comments of this PR.

### DIFF
--- a/packages/effect/src/unstable/cli/Command.ts
+++ b/packages/effect/src/unstable/cli/Command.ts
@@ -122,7 +122,10 @@ export interface Command<Name extends string, Input, E = never, R = never> exten
    * The method which will be invoked with the command-line input and path to
    * execute the logic associated with the command.
    */
-  readonly handle: (input: Input, commandPath: ReadonlyArray<string>) => Effect.Effect<void, E | CliError.CliError, R>
+  readonly handle: (
+    input: Input,
+    commandPath: ReadonlyArray<string>
+  ) => Effect.Effect<void, E | CliError.CliError, R | Environment>
 }
 
 /**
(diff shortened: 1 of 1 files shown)


IMax153:
@voffie - should we also take this opportunity to perhaps add back the `Command.prompt` constructor?

tim-smart:
You will need to add NodeTerminal / BunTerminal first :)

voffie:
Is the NodeTerminal / BunTerminal needed for the `Command.prompt` constructor, or is it a separate concern?

IMax153:
@voffie - I think we can add it as part of this PR, as otherwise there will be no concrete implementations of `Terminal` available for use.

packages/effect/src/unstable/cli/Command.ts
IMax153: We need to fix the signature of `handle` in the `Command` interface to reflect the environment services that are provided by the `run` methods:
```ts
  /**
   * The method which will be invoked with the command-line input and path to
   * execute the logic associated with the command.
   */
  readonly handle: (
    input: Input,
    commandPath: ReadonlyArray<string>
  ) => Effect.Effect<void, E | CliError.CliError, R | Environment>
```

IMax153: We can also subsequently update the signature here in `makeCommand`:
```ts
  readonly handle?:
    | ((input: Input, commandPath: ReadonlyArray<string>) => Effect.Effect<void, E, R | Environment>)
    | undefined
```

IMax153: And here inside `makeCommand`:
```ts
  const handle = (
    input: Input,
    commandPath: ReadonlyArray<string>
  ): Effect.Effect<void, CliError.CliError | E, R | Environment> =>
```

Issue #408, opened 2025-08-14
Add `ensure*` helpers to the Layer module


IMax153:
### What is the problem this feature would solve?

We should add the same `ensure*` helpers we added to `Effect` to `Layer`.

### What is the feature you are proposing to solve the problem?

- `Layer.ensureSuccess`
- `Layer.ensureFailure`
- `Layer.ensureContext`

tim-smart:
We should rename ensureContext to ensureServices

IMax153:
Closed via #644

PR #644, opened 2025-10-12
`Effect.ensure*` -> `Effect.satisfies*`

KhraksMamtsov:
I’d like to suggest changing the name for this family of functions.
```
Effect.ensureSuccessType 
->
Effect.satisfiesSuccess
```
It seems to be a clear and well-established term (`satisfies`) in TypeScript, and it would be better to use it.

## Type

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/Layer.ts
+++ b/packages/effect/src/Layer.ts
@@ -1690,7 +1690,7 @@ export const satisfiesSuccessType =
  * ```ts
  * import { Layer } from "effect"
  *
- * declare const ErrorLayer: Layer<never, Error, never>
+ * declare const ErrorLayer: Layer.Layer<never, Error, never>
  * declare const TypeErrorLayer: Layer.Layer<never, TypeError, never>
  * declare const StringLayer: Layer.Layer<never, string, never>
  *


gcanti:
+1 for changing `ensure` to `satisfies`

I would keep the `Type` postfix though

IMax153:
Agree with `ensure` -> `satisfies` and keeping the `Type` postfix. 

@KhraksMamtsov if you wouldn't mind making that change, we can get this merged.

KhraksMamtsov:
@IMax153 done

KhraksMamtsov:
BTW, I think `Schedule.ensure*` could be aligned as well

IMax153:
@KhraksMamtsov - perhaps we could also address #408 and add the satisfies combinators to `Layer` as part of this PR?

KhraksMamtsov:
@IMax153 Of course, with pleasure!
I also wanted to clarify — would it be possible to name the parameter responsible for context uniformly across all modules?
Right now, in `Effect` it’s called `Requirements`,
in `Schedule` it’s `Services`,
and in `Layer` it will also be `Services`.
What do you think?

IMax153:
I'm not sure about the parameter renaming, as we tend to use R everywhere for effect. But we are leaning into the services terminology for 4.0. 

Let's defer that to a separate potential PR but I'll still ping @mikearnaldi to give his 2 cents here.

KhraksMamtsov:
@IMax153 I added type constraints for `Layer`

PS. I also noticed that `Effect` uses `satisfiesError`, and Layer uses `satisfiesFailure`.

IMax153:
We can standardize this to satisfiesErrorType I think

IMax153:
@KhraksMamtsov looking through the changes, let's make one last adjustment and standardize around satisfiesRequirementsType for now. 

We can change the naming layer if we decide services is better. But since Context doesn't exist in 4.0 I think using requirements for now makes the most sense. 

Then we can get this merged.

KhraksMamtsov:
@IMax153 done
Do we need to do the same with `Schedule.satisfiesServicesType`?

IMax153:
> Do we need to do the same with Schedule.satisfiesServicesType?

@KhraksMamtsov - you know what, I'm so sorry, but can we standardize around what you originally suggested? `ensureServicesType` for everything?

I'm realizing we've modified the names of other combinators that accessed context to services, so that's probably the most consistent.

I'm really sorry for the back and forth. Thanks for your help on this.

KhraksMamtsov:
@IMax153 no problem — that’s part of the workflow :)
so just to clarify,
we want to use `satisfiesServicesType` (`satisfies` — I think that’s what you meant) for all three modules — `Effect`, `Schedule`, and `Layer`, right?

IMax153:
> we want to use `satisfiesServicesType` (`satisfies` — I think that’s what you meant) for all three modules

Sigh - it is, my apologies. That's what I get for responding when I'm over tired 😅

KhraksMamtsov:
It seems to be ready

packages/effect/src/Layer.ts
```diff
@@ -1644,3 +1644,101 @@ const makeUnimplemented = (error: globalThis.Error) => {
...
+ *
+ * This function provides compile-time type checking to ensure that the
+ * requirements (context) type of an layer conforms to a specific type constraint.
+ *
+ * @example
+ * ```ts
+ * import { Layer } from "effect"
+ *
+ * declare const FortyTwoLayer: Layer<never, never, 42>
+ * declare const StringLayer: Layer<never, never, string>
```
IMax153:
```suggestion
 * declare const FortyTwoLayer: Layer.Layer<never, never, 42>
 * declare const StringLayer: Layer.Layer<never, never, string>
```

IMax153 (approved):

PR #722, opened 2025-11-12
fix NonEmptyReadonlyArray in covariant position

gcanti:


packages/effect/src/stream/Channel.ts
```diff
@@ -691,7 +691,7 @@ export const fromChunk = <A>(chunk: Chunk.Chunk<A>): Channel<A> => fromArray(Chu
 export const fromIteratorArray = <A, L>(
   iterator: LazyArg<Iterator<A, L>>,
   chunkSize = DefaultChunkSize
-): Channel<Arr.NonEmptyReadonlyArray<A>, never, L> =>
+): Channel<Arr.NonEmptyArray<A>, never, L> =>
```
tim-smart: Hmm not sure about this. We want to ensure stream chunks are non-empty, and making them mutable means one .pop() / mutation could break things.

PR #800, opened 2025-12-09
add Standard module

gcanti:
**Motivations**

- A sort of standard, introspectable way to represent a schema (so no functions)
- The idea came from wanting to dogfood and streamline our internals. Right now we have a bunch of ad hoc converters from Schema to X, which we can simplify with a standard AST
- And we could easily add more, like https://x.com/rohovskoi_/status/1996859980348379451
- Once the pieces are in place, there is no reason users cannot leverage these tools, even for a subset of schemas, as requested here https://x.com/jonas/status/1996670334561669385
- Another advantage: users have often complained about how hard it is to extract annotations from the AST (and in v4 the AST is even more complex). The standard representation should be a bit easier to work with

**Immediate Goal**

The goal is to make a strict subset of schemas transferable over a wire by serializing them to JSON and then deserializing them back into runtime schema values.

The tactic is as follows:

`Schema` -> `StandardAST` representation -> JSON through `toSerializerJson` -> and back.

Strategically, the `StandardAST` representation can support several compilers:

1. `StandardAST` -> runtime generation
2. `StandardAST` -> code generation
3. `StandardAST` -> types generation

To make a schema transferable over a wire we only need 1), while 2) and 3) can be used in a refactoring of the `FromJsonSchema` module, where the new implementation will be:

`JSONSchema` -> normalized `JSONSchema` -> conversion to the `StandardAST` representation -> various generations.

```mermaid
flowchart TD
    Schema -->|fromAST|StandardAST{"Standard.Document: { schema, definitions }"}
    JS["JSONSchema (draft-07, draft-2020-12, openapi-3.0, openapi-3.1)"] -->|normalize|NJS
    NJS --> |denormalize|JS
    NJS["JSON Schema (draft-2020-12)"] -->|fromJsonSchema|StandardAST
    StandardAST --> |toJson|JSON
    JSON --> |fromJson|StandardAST
    StandardAST --> |toJsonSchema|NJS
    StandardAST --> |toSchema|Schema
    StandardAST --> |toCode|Code["{ code, types: { Type, Encoded, DecodingServices, EncodingServices }, imports }"]
    Schema --> |toArbitrary|Arbitrary
    Schema --> |toEquivalence|Equivalence
    Schema --> |toFormatter|Formatter
```

**Example**

```ts
import { Schema, Standard } from "effect/schema"

  a: Schema.String.annotate({
    description: "my description",
    customAnnotation: 1n
  })
})

// Encode the schema into standard Document (includes the AST and definitions).
const document = Standard.fromAST(schema.ast)

// Encode the document into a JSON-compatible representation.
const json = Standard.toJson(document)

// This output is safe to send over the network or store on disk.
console.log(JSON.stringify(json, null, 2))
/*
{
  "schema": {
    "_tag": "Objects",
    "propertySignatures": [
      {
        "name": "a",
        "type": {
          "_tag": "String",
          "annotations": {
            "description": "my description",
            "customAnnotation": "1"
          },
          "checks": []
        },
        "isOptional": false,
        "isMutable": false
      }
    ],
    "indexSignatures": []
  },
  "definitions": {}
}
*/

// Decode the JSON-compatible representation back into a standard document.
const decodedDocument = Standard.fromJson(json)

// Convert the standard document into a runtime schema.
const decodedSchema = Standard.toSchema<typeof schema>(decodedDocument)

// You can read annotations from the decoded schema, since they were preserved through the round trip.
console.log(decodedSchema.fields.a.ast.annotations)
// { description: 'my description', customAnnotation: 1n }
```

### DIFF
--- a/packages/effect/SCHEMA.md
+++ b/packages/effect/SCHEMA.md
@@ -6510,6 +6510,22 @@ This approach keeps patches independent from TypeScript types and uses the schem
 
 The `Standard` module provides a way to encode and decode schemas to and from JSON. This is useful for sending schemas over the wire or storing them on disk.
 
+```mermaid
+flowchart TD
+    Schema -->|fromAST|StandardAST{"Standard.Document: { schema, definitions }"}
+    JS["JSONSchema (draft-07, draft-2020-12, openapi-3.0, openapi-3.1)"] -->|normalize|NJS
+    NJS --> |denormalize|JS
+    NJS["JSON Schema (draft-2020-12)"] -->|fromJsonSchema|StandardAST
+    StandardAST --> |toJson|JSON
+    JSON --> |fromJson|StandardAST
+    StandardAST --> |toJsonSchema|NJS
+    StandardAST --> |toSchema|Schema
+    StandardAST --> |toCode|Code["{ code, types: { Type, Encoded, DecodingServices, EncodingServices }, imports }"]
+    Schema --> |toArbitrary|Arbitrary
+    Schema --> |toEquivalence|Equivalence
+    Schema --> |toFormatter|Formatter
+```
+
 **Example** (Encoding and decoding a schema)
 
 ```ts


gcanti:
Since too many changes are piling up, I will start by merging this PR and then continue with more granular PRs

PR #804, opened 2025-12-10
Port child process execution APIs

IMax153:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

```ts
import { NodeRuntime, NodeServices } from "@effect/platform-node"
import { Stream } from "effect/stream"
import { ChildProcess } from "effect/unstable/process"

  const handle = yield* ChildProcess.make`ls -lah`.pipe(
    ChildProcess.pipeTo(ChildProcess.make("wc", ["-l"]))
  )

  yield* Effect.log("PID: ", handle.pid)

  const exitCode = yield* handle.exitCode
  yield* Effect.log("Exit Code: ", exitCode)

  // Read stdout
  yield* handle.stdout.pipe(
    Stream.decodeText(),
    Stream.runForEach((output) => Effect.log(output))
  )
  // Read stderr
  yield* handle.stdout.pipe(
    Stream.decodeText(),
    Stream.runForEach((output) => Effect.log(output))
  )
  // Or read both, interleaved
  yield* handle.all.pipe(
    Stream.decodeText(),
    Stream.runForEach((output) => Effect.log(output))
  )
})

Effect.scoped(program).pipe(
  Effect.provide(NodeServices.layer),
  NodeRuntime.runMain
)
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -5,7 +5,7 @@ Releases are snapshot only
 - [x] Port Channel apis
 - [ ] Port Stream / Sink apis
 - [x] Port worker modules
-- [ ] Port command execution modules
+- [x] Port command execution modules
 - [ ] Port platform-browser
 - [x] Add CLI modules
 - [ ] Add opentelemetry package


packages/effect/src/stream/Sink.ts
```diff
@@ -454,6 +474,14 @@ export const fromEffect = <A, E, R>(
     (a) => Pull.halt<End<A>>([a])
   ))))
 
+/**
+ * Drains elements from the stream by ignoring all inputs.
+ *
+ * @since 2.0.0
+ * @category constructors
+ */
+export const drain: Sink<void, unknown> = fromChannel(Channel.drain(Channel.identity()) as any)
```
tim-smart: This doesn't look quite right. You will need to use Channel.mapDone to set an output of `void`.

IMax153: Got it - I was going off the v3 version, but I think our type signatures weren't quite as good there. Thanks for pushing a fix 👍

Issue #831, opened 2025-12-16
instanceof leftovers


gcanti:
There are still a couple of `instanceof`s in the codebase

- packages/effect/src/internal/hashMap.ts `remaining instanceof LeafNode`
- packages/effect/src/unstable/httpapi/HttpApiSchema.ts `u instanceof EmptyError`
- packages/effect/src/unstable/cli/Command.ts `error instanceof CliError.ShowHelp`

PR #848, opened 2025-12-17
Remove `instanceof` check from `HttpApiSchema` module

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Related

- Related Issue #831
- Closes #


IMax153:
@tim-smart - not sure if there's a better way to do the above check to avoid `instanceof` (https://github.com/Effect-TS/effect-smol/issues/831)

PR #866, opened 2026-01-01
chore: migrate to oxlint

fubhy:
Starting to experiment with oxlint and oxfmt


IMax153 (approved):
This LGTM - we should probably notify the rest of the team regarding these changes though so people can update their editor settings etc.

fubhy:
> This LGTM - we should probably notify the rest of the team regarding these changes though so people can update their editor settings etc.

I have bundled editor settings for .vscode/settings.json but for stuff like zed, nvim, etc. some manual changes are necessary.

PR #893, opened 2026-01-09
improve OPTIC.md docs, closes #885

### LINKED ISSUE #885: `packages/effect/OPTIC.md` mentions `ToOptic` which is not present in the codebase
### What is the type of issue?

Documentation is incorrect

### What is the issue?

`ToOptic.ts` doesn't exist and IIUC was replaced by `Schema.makeIso`

### Where did you find it?

https://github.com/Effect-TS/effect-smol/blob/d92328569a2c308d15ce674fd5c94447842a103a/packages/effect/OPTIC.md#L12-L13

gcanti:


nikelborm:
Oh, wait, my mistake. You already fixed mentions of `ToOptic.ts` in https://github.com/Effect-TS/effect-smol/pull/802; I just had an old local version. Anyway, thanks for making the docs even better 🫶

PR #937, opened 2026-01-15
EFF-177: Refactor Effect.filterOrElse

tim-smart:
## Summary
- shift `Effect.filterOrElse` overloads to predicate/refinement signatures
- update `filterOrFailCause` to predicate/refinement overloads
- remove unused `Filter` import in HttpClient


packages/effect/src/internal/effect.ts
tim-smart: filterOrFailCause should use predicates too

PR #1105, opened 2026-01-28
TIM-439: Stream jsdocs

tim-smart:
## Summary
- update Stream JSDoc examples to log finalizer output and remove Chunk-shaped output comments
- switch updateService example to ServiceMap.Service class syntax
- keep examples aligned with Stream Array outputs

### DIFF
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -559,7 +559,7 @@ export const transformPull = <A, E, R, B, E2, R2, EX, RX>(
  *   stream,
  *   (pull, _scope, forkedScope) =>
  *     Effect.gen(function*() {
- *       yield* Scope.addFinalizer(forkedScope, Effect.void)
+ *       yield* Scope.addFinalizer(forkedScope, Console.log("Releasing scope"))
  *       return pull
  *     })
  * )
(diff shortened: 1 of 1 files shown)


packages/effect/src/Stream.ts
```diff
@@ -6002,8 +8164,23 @@ export const encodeText = <E, R>(self: Stream<string, E, R>): Stream<Uint8Array,
...
+ * ```ts
+ * import { Console, Effect, Stream } from "effect"
+ *
+ * Effect.runPromise(Effect.gen(function* () {
+ *   const lines = yield* Stream.runCollect(
+ *     Stream.make("a\nb\r\n", "c\n").pipe(Stream.splitLines)
+ *   )
+ *   yield* Console.log(lines)
+ * }))
+ * // { _id: 'Chunk', values: [ 'a', 'b', 'c' ] }
```
tim-smart: Remove Chunk from output

packages/effect/src/Stream.ts
```diff
@@ -6410,6 +8777,38 @@ export const provideServices: {
...
+ *       Stream.provideService(Greeter, {
+ *         greet: (name) => `Hello, ${name}`
+ *       })
+ *     )
+ *   )
+ *   yield* Console.log(collected)
+ * })
+ *
+ * Effect.runPromise(program)
+ * //=> { _id: 'Chunk', values: [ 'Hello, Ada' ] }
```
tim-smart: Remove chunk from output

PR #1144, opened 2026-01-30
TIM-457: Add PriorityScheduler to Scheduler module

tim-smart:
## Summary

This PR adds a new `PriorityScheduler` to the Scheduler module, copying the behavior from Effect 3's `MixedScheduler`.

### Changes

- **Added `PriorityBuckets` class**: A utility class that maintains tasks sorted by priority in descending order (higher numbers = higher priority). Tasks with the same priority are grouped together and executed in FIFO order.

- **Added `PriorityScheduler` class**: A priority-based scheduler implementation that:
  - Uses microtask-based execution (`Promise.resolve()`) for low latency
  - Falls back to `setTimeout` after a configurable number of microtask iterations to prevent event loop starvation
  - Configurable `maxNextTickBeforeTimer` parameter (default: 2048)
  - Implements the standard `Scheduler` interface with `executionMode: "async"`
  - Higher priority tasks execute before lower priority tasks

- **Added comprehensive tests**: Full test coverage for both `PriorityBuckets` and `PriorityScheduler` including:
  - Priority ordering verification
  - FIFO ordering within same priority
  - Task scheduling and execution
  - `shouldYield` behavior
  - Nested task scheduling

### API

```typescript
import { PriorityScheduler } from "effect/Scheduler"

// Create with default settings (2048 max microtasks before setTimeout)
const scheduler = new PriorityScheduler()

// Or with custom threshold
const customScheduler = new PriorityScheduler(100)

// Schedule tasks with different priorities
scheduler.scheduleTask(() => console.log("Low priority"), -1)
scheduler.scheduleTask(() => console.log("High priority"), 10)
// Output: "High priority" (executes first), then "Low priority"
```

### Implementation Details

The scheduler maintains tasks in priority buckets and processes them using a "starvation" pattern:
1. Tasks are queued in `PriorityBuckets` sorted by priority
2. On the first schedule, the scheduler starts a microtask chain
3. After `maxNextTickBeforeTimer` microtask iterations, it falls back to `setTimeout` to allow the event loop to process other events
4. This prevents the scheduler from monopolizing the event loop while maintaining low latency for high-priority tasks

### Testing

All tests pass:
- `pnpm test Scheduler.test.ts` - 9 tests passing
- `pnpm lint-fix` - No issues
- `pnpm check` - Type checking passes
- `pnpm docgen` - Documentation generation passes

### DIFF
--- a/packages/effect/src/Scheduler.ts
+++ b/packages/effect/src/Scheduler.ts
@@ -195,14 +195,15 @@ export class MixedScheduler implements Scheduler {
 
 /**
  * A priority bucket queue that maintains tasks sorted by priority.
- * Higher priority tasks (higher numbers) are scheduled before lower priority tasks.
+ * Lower numbers mean higher priority.
  *
  * @since 2.0.0
  * @category utils
  */
 export class PriorityBuckets<in out T = () => void> {
   /**
-   * The buckets array containing [priority, tasks] tuples, sorted by priority in descending order.
+   * The buckets array containing [priority, tasks] tuples, sorted by priority in ascending order.
+   * Lower priority numbers are executed first.
    * @since 2.0.0
    */
   public buckets: Array<[number, Array<T>]> = []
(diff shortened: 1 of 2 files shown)


packages/effect/src/Scheduler.ts
```diff
@@ -193,6 +193,181 @@ export class MixedScheduler implements Scheduler {
   }
 }
 
+/**
+ * A priority bucket queue that maintains tasks sorted by priority.
+ * Higher priority tasks (higher numbers) are scheduled before lower priority tasks.
```
tim-smart: Lower numbers mean higher priority

PR #1159, opened 2026-02-03
EFF-393 Fix Prompt.text wrapping clear

tim-smart:
## Summary
- replace `*Plain` render helpers with `RenderOptions` and use plain mode for clear paths on input prompts (confirm/date/number/text/autoComplete)
- keep file prompt clears aligned with prompt + resolved path/choices output to stabilize line erasing
- use plain rendering for select/multi-select/toggle clears to avoid ANSI wrapping


tim-smart (commented):
Do any other prompt types have this issue? We need to fix those too.

tim-smart (commented):
I think we should only handle the cases where the user can type some input

PR #1173, opened 2026-02-05
EFF-413 Fix Atom Node listeners missing listeners on notify

tim-smart:
## Summary
- switch `NodeImpl` listener storage in `packages/effect/src/unstable/reactivity/AtomRegistry.ts` from an array to a `Set`, so unsubscribe mutations during notification do not corrupt listener iteration.
- align listener lifecycle checks with the new container (`size`, `add`, `delete`, `clear`) and use `Set.forEach` with a shared callback in `notify` to address review feedback.
- keep regression coverage in `packages/effect/test/reactivity/Atom.test.ts` for self-unsubscribe during notify and re-run validation: `pnpm lint-fix`, `pnpm test packages/effect/test/reactivity/Atom.test.ts`, `pnpm check`, `pnpm build`, `pnpm docgen`.

### DIFF
--- a/packages/effect/src/unstable/reactivity/AtomRegistry.ts
+++ b/packages/effect/src/unstable/reactivity/AtomRegistry.ts
@@ -221,6 +221,10 @@ export const mount: {
 
 const constImmediate = { immediate: true }
 
+const notifyListener = (listener: () => void): void => {
+  listener()
+}
+
 const SerializableTypeId: Atom.SerializableTypeId = "~effect-atom/atom/Atom/Serializable"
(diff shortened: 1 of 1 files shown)


packages/effect/src/unstable/reactivity/AtomRegistry.ts
```diff
@@ -622,8 +622,8 @@ class NodeImpl<A> {
   }
 
   notify(): void {
-    for (let i = 0; i < this.listeners.length; i++) {
-      this.listeners[i]()
+    for (const listener of this.listeners) {
```
tim-smart: Use this.listeners.forEach, with a constant callback defined elsewhere `function constListenerCallback` potentially

PR #1188, opened 2026-02-09
EFF-418 Port Effect.awaitAllChildren from Effect 3

tim-smart:
## Summary
- ported `Effect.awaitAllChildren` to the v4 runtime in `packages/effect/src/internal/effect.ts`
- implemented scoped child waiting by capturing a baseline `fiberIdStore.id` and awaiting only still-live children with `child.id > baseline` in the finalizer
- kept behavior scoped to children forked by the wrapped effect, without eagerly snapshotting `_children`
- added a regression test in `packages/effect/test/Effect.test.ts` that verifies preexisting same-fiber children are not awaited

## Validation
- `pnpm lint-fix`
- `pnpm test packages/effect/test/Effect.test.ts`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`

### DIFF
--- a/packages/effect/src/internal/effect.ts
+++ b/packages/effect/src/internal/effect.ts
@@ -13,6 +13,7 @@ import type { LazyArg } from "../Function.ts"
 import { constant, constFalse, constTrue, constUndefined, constVoid, dual, identity } from "../Function.ts"
 import * as Hash from "../Hash.ts"
 import { toJson, toStringUnknown } from "../Inspectable.ts"
+import * as Iterable from "../Iterable.ts"
 import type * as Logger from "../Logger.ts"
 import type * as LogLevel from "../LogLevel.ts"
 import type * as Metric from "../Metric.ts"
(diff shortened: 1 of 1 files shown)


packages/effect/src/internal/effect.ts
```diff
@@ -4362,6 +4362,33 @@ export const forkDetach: {
 ): Effect.Effect<Fiber.Fiber<A, E>, never, R> =>
   withFiber((fiber) => succeed(forkUnsafe(fiber, self, options?.startImmediately, true, options?.uninterruptible))))
 
+/** @internal */
+export const awaitAllChildren = <A, E, R>(
+  self: Effect.Effect<A, E, R>
+): Effect.Effect<A, E, R> =>
+  withFiber((fiber) => {
+    const initialChildren = fiber._children === undefined ? undefined : new Set(fiber._children)
```
tim-smart: Don't need to capture the initial children here.

PR #1195, opened 2026-02-10
Add catchTagOrDie and catchTagsOrDie APIs

kyrregjerstad:
Adapted from https://github.com/Effect-TS/effect/pull/6043

## Summary
- Add `catchTagOrDie` — catches specific tagged errors by `_tag`, converting unmatched errors into defects (die). Unlike `catchTag`, unhandled errors are removed from the error channel entirely.
- Add `catchTagsOrDie` — same semantics but accepts an object of handlers (like `catchTags`).
- Both support dual (data-first and data-last) calling conventions and variadic tag arguments (for `catchTagOrDie`).

## Motivation

A common pattern is catching one specific tagged error and re-mapping it, while treating all other errors as unexpected defects. The natural approach doesn't work:

```ts
declare const findUser: Effect<User, DbNotFound | DbUnknown>

findUser.pipe(
  Effect.catchTag("DbNotFound", () => new MyError({ message: "not found" })),
  Effect.orDie // also turns MyError into a defect!
)
```

`orDie` kills *all* errors — including the `MyError` that was just introduced by the handler. One workaround today is a manual `catchAll` branch:

```ts
findUser.pipe(
  Effect.catchAll((e) =>
    e._tag === "DbNotFound"
      ? new MyError({ message: "not found" })
      : Effect.die(e)
  )
)
```

`catchTagOrDie` makes this a one-liner:

```ts
findUser.pipe(
  Effect.catchTagOrDie("DbNotFound", () =>
     new MyError({ message: "not found" })
  )
)
// Effect<User, MyError>
```

### DIFF
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -2636,6 +2636,159 @@ export const catchTags: {
 } = internal.catchTags
 
 /**
+ * Catches specific errors by their `_tag` field and handles them, while
+ * converting all other (unmatched) errors into defects (die).
+ *
+ * **When to Use**
+ *
+ * Use `catchTagOrDie` when you want to handle one or more specific tagged
+ * errors and treat any remaining errors as unexpected defects. Unlike
+ * {@link catchTag}, unmatched errors are removed from the error channel and
+ * become fatal defects instead.
+ *
+ * The error type must have a readonly `_tag` field to use `catchTagOrDie`.
+ *
+ * @example
+ * ```ts
+ * import { Data, Effect } from "effect"
+ *
+ * class HttpError extends Data.TaggedError("HttpError") {}
+ *
+ * class ValidationError extends Data.TaggedError("ValidationError") {}
+ *
+ * declare const program: Effect.Effect<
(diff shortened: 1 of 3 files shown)


IMax153:
@kyrregjerstad - I think this API has a bit too much indirection / lack of explicitness to be accepted into the core library.

A better API, which we've already been considering and has a bit less indirection could be `orDieTag`, so you could `catch` / `catchTag` / `catchTags` the errors you want to handle first, and then use `orDieTag` to die on specific ones. You could support multiple tags in `orDieTag` too like we do with `catchTag` so you don't have to chain multiple together to die on multiple tags.

kyrregjerstad:
@IMax153 Thanks for the feedback! That makes sense, `orDieTag` is a cleaner separation of concerns. `catchTagOrDie` conflates "handle this" and "die on the rest" into one step, whereas `orDieTag` lets you be explicit about *which* errors become defects.

I've been using `catchTagOrDie` in my own codebase and found it useful, but I see how `orDieTag` makes a better API for the core library.

PR #1236, opened 2026-02-14
port Array tests from v3

gcanti:


packages/effect/src/index.ts
```diff
@@ -2525,9 +2606,10 @@ export * as TxRef from "./TxRef.ts"
 export * as TxSemaphore from "./TxSemaphore.ts"
 
 /**
- * A collection of types that are commonly used types.
+ * Repesents any type that can also be unioned with other types.
```
IMax153: I think the prior description was actually better than this one.

PR #1246, opened 2026-02-16
EFF-467 Port PartitionedSemaphore from effect v3

tim-smart:
## Summary
- Port `PartitionedSemaphore` into `packages/effect/src/PartitionedSemaphore.ts`, including v4 callback-based waiter registration and round-robin permit release across partitions.
- Add coverage in `packages/effect/test/PartitionedSemaphore.test.ts` for fairness, interruption safety, partial acquisition, and edge cases.
- Regenerate the effect barrel export (`packages/effect/src/index.ts`) via `pnpm codegen`.

## Validation
- `pnpm lint-fix`
- `pnpm test packages/effect/test/PartitionedSemaphore.test.ts`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`

### DIFF
--- a/packages/effect/src/PartitionedSemaphore.ts
+++ b/packages/effect/src/PartitionedSemaphore.ts
@@ -1,5 +1,5 @@
 /**
- * @since 4.0.0
+ * @since 3.19.4
  * @experimental
  */
 import * as Effect from "./Effect.ts"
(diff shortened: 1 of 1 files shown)


packages/effect/src/PartitionedSemaphore.ts
```diff
@@ -0,0 +1,182 @@
+/**
+ * @since 4.0.0
```
tim-smart: Use the same since tags as effect v3

PR #1280, opened 2026-02-18
EFF-491 Port Resource module from effect v3

tim-smart:
## Summary
- port `Resource` as a public v4 module in `packages/effect/src/Resource.ts` with `manual`, `auto`, `get`, and `refresh` built on `ScopedRef<Exit<A, E>>`
- capture construction-time services in `manual` with `Effect.updateServices` + `ServiceMap.merge`, and keep refresh semantics where failed refreshes preserve the last successful cached value
- address review feedback by removing `Effect.Yieldable` / variance from `Resource`, simplifying `auto` to scoped `Effect.repeat(refresh(self), policy)`, and updating tests accordingly

## Validation
- `pnpm lint-fix`
- `pnpm test packages/effect/test/Resource.test.ts`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`

### DIFF
--- a/packages/effect/src/Resource.ts
+++ b/packages/effect/src/Resource.ts
@@ -2,7 +2,7 @@
  * @since 2.0.0
  */
 import * as Effect from "./Effect.ts"
-import type * as Exit from "./Exit.ts"
+import * as Exit from "./Exit.ts"
 import { identity } from "./Function.ts"
 import { PipeInspectableProto } from "./internal/core.ts"
 import type { Pipeable } from "./Pipeable.ts"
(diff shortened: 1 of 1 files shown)


packages/effect/src/Resource.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * @since 2.0.0
+ */
+export declare namespace Resource {
+  /**
+   * @since 2.0.0
+   * @category models
+   */
+  export interface Variance<in out A, in out E> {
```
tim-smart: Don't need Variance, as all the type parameters are used in the interface

PR #1313, opened 2026-02-19
feat: add `Request.Parameters` to extract parameters from a `Request`

### LINKED ISSUE #1312: Add `Request.Parameters` to extract input from a `Request`
### What is the problem this feature would solve?

At the moment there is no utility type for extracting the parameters needed to create a request.

This creates duplication when the request is wrapped in a function (eg: a service method that executes a request).

Another improvement, imo, is not having to pass an empty object if there are no parameters like other modules (eg: `Data.TaggedError`).

### What is the feature you are proposing to solve the problem?

```ts
import type { Request } from "effect"

interface GetUser extends Request.Request<string, Error> {
  readonly _tag: "GetUser"
  readonly id: number
}

// Extract the parameters type from a Request using the utility
type GetUserParameters = Request.Parameters<GetUser> // { id: number }
```

Can be used, for example, in `Layer` constructors as the pattern is pretty common.

```ts
import { Effect, Request, RequestResolver, ServiceMap } from "effect"

class GetUser extends Request.TaggedClass("GetUser")<
  { id: number },
  { email: string }
> {}

class DeleteAllUsers extends Request.TaggedClass("DeleteAllUsers")<
  {},
  void
> {}

export class Queries extends ServiceMap.Service<Queries>()("Queries", {
  make: Effect.gen(function*() {
    const getUserResolver = RequestResolver.fromEffect<GetUser>(
      Effect.fnUntracedEager(function*() {
        return { email: "foo@bar.tld" }
      })
    )

    const deleteUserResolver = RequestResolver.fromEffect<DeleteAllUsers>(
      Effect.fnUntracedEager(function*() {
        return Effect.log(`All users deleted`)
      })
    )

    //                 ┌─── { id: number }
    //                 ▼
    const getUser = (params: Request.Parameters<GetUser>) => Effect.request(new GetUser(params), getUserResolver)

    //                                                        ┌─── void
    //                                                        ▼
    const deleteUser = () => Effect.request(new DeleteAllUsers(), deleteUserResolver)

    return { getUser, deleteUser } as const
  })
}) {}
```

### What alternatives have you considered?

Manually copying and duplicating request constructor's properties in function parameters.

f15u:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Closes #1312

### DIFF
--- a/.changeset/sour-sheep-glow.md
+++ b/.changeset/sour-sheep-glow.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `Request.Parameters` to extract parameters from a `Request`
--- a/packages/effect/dtslint/Request.tst.ts
+++ b/packages/effect/dtslint/Request.tst.ts
@@ -1,0 +1,32 @@
+import { Request } from "effect"
+import { describe, expect, it } from "tstyche"
+
+class TaggedNoInput extends Request.TaggedClass("TaggedNoInput")<
+  {},
+  void
+> {}
+
+class TaggedWithInput extends Request.TaggedClass("TaggedWithInput")<
+  {
+    readonly foo: string
+  },
+  void
+> {}
+
+class UntaggedNoInput extends Request.Class<{}, void> {}
+
+class UntaggedWithInput extends Request.Class<
+  {
+    readonly foo: string
+  },
+  void
+> {}
+
+describe("Request", () => {
+  it("Input", () => {
+    expect<Request.Parameters<TaggedNoInput>>().type.toBe<void>()
+    expect<Request.Parameters<UntaggedNoInput>>().type.toBe<void>()
+    expect<Request.Parameters<TaggedWithInput>>().type.toBe<{ readonly foo: string }>()
+    expect<Request.Parameters<UntaggedWithInput>>().type.toBe<{ readonly foo: string }>()
+  })
+})
--- a/packages/effect/src/Request.ts
+++ b/packages/effect/src/Request.ts
@@ -164,6 +164,29 @@ export type Services<T extends Request<any, any, any>> = [T] extends [Request<in
 export type Result<T extends Request<any, any, any>> = T extends Request<infer A, infer E, infer _R> ? Exit.Exit<A, E>
   : never
 
+/**
+ * A utility type to extract the parameters type from a `Request`.
+ *
+ * @example
+ * ```ts
+ * import type { Request } from "effect"
+ *
+ * interface GetUser extends Request.Request<string, Error> {
+ *   readonly _tag: "GetUser"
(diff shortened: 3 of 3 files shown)


IMax153:
Perhaps a better name for this would be `Request.Params` or `Request.Parameters` instead? Just a suggestion.

f15u:
> Perhaps a better name for this would be `Request.Params` or `Request.Parameters` instead? Just a suggestion.

I agree, `Request.Parameters` sounds better. Thanks for the suggestion!

I'm currently waiting for feedback in the issue. I'm not sure I got Tim's suggestion right. Otherwise I can remove the draft status from here and proceed

IMax153:
> I'm currently waiting for feedback in the issue

I pinged @tim-smart for review

IMax153:
@f15u - is this pull request still relevant given that you must pass the full `Request` instance? I'm not sure there's a use case for this any longer?

f15u:
> @f15u - is this pull request still relevant given that you must pass the full `Request` instance? I'm not sure there's a use case for this any longer?

@IMax153 imo is still relevant. Using a `Layer` to wrap `Effect.request` and requiring the request itself required as an argument I see it as kind of leak and introduce another layer of indirection; this proposal solve this and improve DX.

But if the team doesn't agree, I'm ok with that and we can close this.

IMax153:
>requiring the request itself required as an argument I see it as kind of leak and introduce another layer of indirection

As @tim-smart mentioned [here](https://github.com/Effect-TS/effect-smol/issues/1312#issuecomment-3931140608), the request constructor is explicitly _required_ for the request resolver subsystem to work properly in Effect. 

I can see the utility of this type helper for folks who want to create helper functions for the request, e.g.

```ts
function getUser(params: Request.Parameters<GetUser>) {
  return Effect.request(new GetUser(params), getUserResolver)
}
```

but do note that the `new GetUser(...)` is still required for `Effect.request`.

I'll let @tim-smart have the final say on this proposal.

f15u:
> [...] the request constructor is explicitly _required_ for the request resolver subsystem to work properly in Effect. [...]

> [...] but do note that the `new GetUser(...)` is still required for `Effect.request`.

Absolutely! To be clear, I don't want to remove the constructor passed to `Effect.request`.
I want to simplify and improve the DX for wrapping `Effect.request`.

> I can see the utility of this type helper for folks who want to create helper functions for the request, e.g.
> 
> ```ts
> function getUser(params: Request.Parameters<GetUser>) {
>   return Effect.request(new GetUser(params), getUserResolver)
> }
> ```
This use case is exactly what made me think of this proposal :100:

IMax153:
Hi @f15u,

PR #1314, opened 2026-02-19
Port ReactHydration

zeyuri:
## Summary

- Add `Hydration` module to `effect/unstable/reactivity` with `dehydrate`, `hydrate`, and `toValues` for SSR atom state serialization
- Add `HydrationBoundary` React component to `@effect/atom-react` with two-phase hydration (new atoms in render, existing atoms after commit)
- Support streaming via `resultPromise` for async atoms in Initial state
- **Fix `Atom.serializable` encode/decode for wire transfer** (see bug details below)
- 9 new hydration test cases, all passing

Ported from `@effect-atom/atom` Hydration module, adapted for v4 APIs (`AsyncResult`, `AtomRegistry`, etc.).

### Problem

`Atom.serializable` used `Schema.encodeSync(schema)` / `Schema.decodeSync(schema)` directly. For `AsyncResult.Schema`, this only reaches the intermediate `AsyncResult<Encoded>` type — instances with custom `ResultProto` prototype containing the `TypeId` symbol.

After wire transfer (JSON serialization, seroval, or any serialization that doesn't preserve prototypes), the custom prototype is lost. `Schema.decodeSync` then fails because the `declareConstructor` decoder calls `isAsyncResult(input)` which checks for the `TypeId` symbol on the prototype — returning `false` for plain objects.

**Error:**
```
Error: Expected AsyncResult, got {"_tag":"Success","value":{...},"waiting":false,"timestamp":...}
```

This breaks any SSR scenario where atom state is dehydrated on the server, serialized over the wire (HTML script tags, streaming), and hydrated on the client.

### Root cause

`Schema.encodeSync(AsyncResult.Schema({...}))` only goes through the `declareConstructor` level, which:
- **Encode**: validates the input is an `AsyncResult` instance and encodes inner values, producing another `AsyncResult` instance (not plain JSON)
- **Decode**: checks `isAsyncResult(input)` first, then decodes inner values

The `toCodec` defined on `AsyncResult.Schema` handles the full pipeline (Type ↔ plain JSON tagged structs), but `Schema.encodeSync` doesn't go through it.

### Fix

Changed `Atom.serializable` to use `Schema.toCodecJson(schema)` which produces a codec that goes through the full pipeline via `toCodec`:
- **Encode**: produces plain JSON tagged structs (`{ _tag: "Success", value: ..., waiting: false, timestamp: ... }`)
- **Decode**: accepts plain JSON tagged structs and reconstructs proper `AsyncResult` instances

## Test plan

- [x] `pnpm --filter @effect/atom-react test` — 18 pass, 4 pre-existing failures (DOM test isolation on main)
- [x] `serializable encode/decode survives JSON roundtrip (wire transfer)` — unit test for encode → JSON.parse(JSON.stringify()) → decode
- [x] `dehydrate + JSON roundtrip + hydrate works (SSR simulation)` — integration test with HydrationBoundary
- [x] Rebased on latest upstream/main

### DIFF
--- a/.changeset/breezy-meals-see.md
+++ b/.changeset/breezy-meals-see.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+support tag unions in Effect.catchTag/Reason
--- a/packages/effect/src/FileSystem.ts
+++ b/packages/effect/src/FileSystem.ts
@@ -733,8 +733,7 @@ export const make = (
         Effect.as(true),
         Effect.catchTag(
           "PlatformError",
-          (e) =>
-            e.reason._tag === "SystemError" && e.reason.kind === "NotFound" ? Effect.succeed(false) : Effect.fail(e)
+          (e) => e.reason._tag === "NotFound" ? Effect.succeed(false) : Effect.fail(e)
         )
       ),
     readFileString: (path, encoding) =>
@@ -802,7 +801,7 @@ const notFound = (method: string, path: string) =>
   systemError({
     module: "FileSystem",
     method,
-    kind: "NotFound",
+    _tag: "NotFound",
     description: "No such file or directory",
     pathOrDescriptor: path
   })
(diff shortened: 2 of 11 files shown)


.changeset/port-react-hydration.md
```diff
@@ -0,0 +1,8 @@
+---
+"effect": minor
+"@effect/atom-react": minor
```
tim-smart:
```suggestion
"effect": patch
"@effect/atom-react": patch
```

zeyuri:
@tim-smart I changed to draft because I have found a little bug, will fix soon

PR #1340, opened 2026-02-19
fix: Stream.fromQueue accepts PubSub.Subscription

cevr:
## Summary

In v3, `PubSub.Subscription` extended `Queue.Dequeue`, so `Stream.fromQueue(subscription)` worked. In v4, they're unrelated types — `Stream.fromQueue` only accepts `Dequeue<A, E>`, and `Stream.fromSubscription` exists as the alternative but isn't discoverable from the type error.

This widens `Stream.fromQueue` to accept `Dequeue<A, E> | PubSub.Subscription<A>`, with runtime dispatch via the subscription type ID. Type inference works correctly: `Subscription<A>` has no error channel, so `E` infers as `never` and the return collapses to `Stream<A>`.

`Stream.fromSubscription` remains unchanged — this just makes the common v3 pattern work again.

**Why not overloads?** Overloads break point-free usage (`Effect.map(Stream.fromQueue)`) which is used in Reactivity, SqlClient, and Geolocation internally. The union approach preserves inference in all positions.

## Test plan

- [x] New test: `fromQueue - accepts PubSub.Subscription` — creates PubSub, subscribes, publishes values, reads via `Stream.fromQueue(subscription)`
- [x] Existing `Stream.fromQueue` tests pass unchanged
- [x] Typecheck clean

### DIFF
--- a/.changeset/stream-fromqueue-accepts-subscription.md
+++ b/.changeset/stream-fromqueue-accepts-subscription.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+`Stream.fromQueue` now accepts `PubSub.Subscription` in addition to `Queue.Dequeue`, restoring the v3 pattern where subscriptions could be passed directly.


cevr:
Closing — `Stream.fromSubscription` already exists for this use case. Documented the migration path in #1344 instead.

PR #1342, opened 2026-02-19
fix: allow omitting props in TaggedErrorClass with no fields

cevr:
## Summary

`TaggedErrorClass` with no extra fields requires `new MyError({})` — passing an empty object to a constructor that carries no information. In v3, `new MyError()` worked.

The issue is in the `Class` interface constructor signature:

```ts
new(props: S["~type.make.in"], options?: MakeOptions): S["Type"] & Inherited
```

For `TaggedErrorClass("NotFoundError", {})`, the `_tag` field has a constructor default, so `~type.make.in` resolves to `{ _tag?: "NotFoundError" }` — all properties optional. But the `props` parameter itself is required, so you still must pass `{}`.

**Fix:** Use a conditional type to make `props` optional when all fields are defaulted:

```ts
new(...args: {} extends S["~type.make.in"]
  ? [props?: S["~type.make.in"], options?: MakeOptions]
  : [props: S["~type.make.in"], options?: MakeOptions]
): S["Type"] & Inherited
```

When `{} extends { _tag?: "NotFoundError" }` is true (all optional), `props` becomes optional. The runtime constructor defaults `undefined` input to `{}`.

This also applies to `Schema.Class` and `Schema.ErrorClass` — any class where all fields have constructor defaults can now omit the props argument.

## Test plan

- [x] New test: `zero-field TaggedErrorClass allows omitting props argument` — verifies both `new NotFoundError()` and `new NotFoundError({})` work
- [x] Existing `TaggedErrorClass` tests pass (4/4)
- [x] Typecheck clean — no regressions from the conditional constructor type

### DIFF
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -7665,8 +7665,10 @@ export interface Class<Self, S extends Top & { readonly fields: Struct.Fields },
   >
 {
   // intentionally left without `readonly "~rebuild.out": this`
-  new(...args: {} extends S["~type.make.in"] ? [props?: S["~type.make.in"], options?: MakeOptions]
-    : [props: S["~type.make.in"], options?: MakeOptions]): S["Type"] & Inherited
+  new(
+    ...args: {} extends S["~type.make.in"] ? [props?: S["~type.make.in"], options?: MakeOptions]
+      : [props: S["~type.make.in"], options?: MakeOptions]
+  ): S["Type"] & Inherited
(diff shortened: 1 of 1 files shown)


IMax153:
@cevr - you have linting issues (can fix with `pnpm lint-fix`). You'll also need to rebase against `main`. Then this should be able to be auto-merged.

cevr:
@IMax153 ah dang i think i messed up the auto-merge - my bad 😔

PR #1343, opened 2026-02-19
feat: add ServiceMap.unsafeEmpty() for contravariance workaround

cevr:
## Summary

`ServiceMap<in Services>` is contravariant. This means `ServiceMap.empty()` (which returns `ServiceMap<never>`) is **not** assignable to `ServiceMap<any>` — variance goes the wrong direction. `never` is the bottom type, so `ServiceMap<never>` is the most specific (accepts no services), while `ServiceMap<any>` is the most general.

This makes it painful to use empty service maps as default values in generic containers:

```ts
// Fails: ServiceMap<never> not assignable to ServiceMap<any>
const services: ServiceMap<any> = ServiceMap.empty()

// Workaround: cast at every call site
const services: ServiceMap<any> = ServiceMap.empty() as ServiceMap<any>
```

**Fix:** Add `ServiceMap.unsafeEmpty(): ServiceMap<any>` — same runtime value as `empty()`, typed as `ServiceMap<any>`. The `unsafe` prefix signals that you're opting out of the type-level service tracking.

## Test plan

- [x] Typecheck clean
- [x] `unsafeEmpty()` returns the same singleton as `empty()` (no runtime cost)

### DIFF
--- a/.changeset/servicemap-unsafe-empty.md
+++ b/.changeset/servicemap-unsafe-empty.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `ServiceMap.unsafeEmpty()` which returns `ServiceMap<any>`, working around the contravariance issue where `ServiceMap.empty()` (`ServiceMap<never>`) is not assignable to `ServiceMap<any>`.


tim-smart:
I think you can just use the type cast if you are doing something unsafe :)

PR #1375, opened 2026-02-20
feat(rpc): add supportsSpanPropagation option to RpcServer.toHttpEffect

marbemac:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add optional `supportsSpanPropagation` option to `RpcServer.toHttpEffect`.

The HTTP protocol hardcodes `supportsSpanPropagation: false`, which makes sense as a default for typical HTTP RPC servers behind reverse proxies. But for same-origin setups like Cloudflare Durable Object RPC, the client and server share a trust boundary and propagating spans is expected.

This adds a single optional field to `toHttpEffect`'s options. When set to `true`, it overrides the protocol's `supportsSpanPropagation` before providing it to `make()`. No changes to `makeProtocolWithHttpEffect` — no breaking changes.

```ts
const httpApp = RpcServer.toHttpEffect(myGroup, {
  supportsSpanPropagation: true
})
```

## Related

- #6072, https://github.com/Effect-TS/effect/pull/6073

### DIFF
--- a/.changeset/rpc-toHttpEffect-span-propagation.md
+++ b/.changeset/rpc-toHttpEffect-span-propagation.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Add optional `supportsSpanPropagation` option to `RpcServer.toHttpEffect` for enabling distributed tracing across HTTP RPC boundaries.
--- a/packages/effect/src/unstable/rpc/RpcServer.ts
+++ b/packages/effect/src/unstable/rpc/RpcServer.ts
@@ -1078,6 +1078,7 @@ export const toHttpEffect: <Rpcs extends Rpc.Any>(
     readonly disableTracing?: boolean | undefined
     readonly spanPrefix?: string | undefined
     readonly spanAttributes?: Record<string, unknown> | undefined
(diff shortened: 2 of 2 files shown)


tim-smart:
The reason it is set to false, is so the http spans are included in the trace.

Rpc request -> Http client request -> Http server -> Rpc server

If span propagation is turned on, the http client and server spans will get orphaned.

marbemac:
Thanks for the explanation Tim, that makes sense — I hadn't considered the orphaned HTTP spans issue.

My use case is Cloudflare Durable Objects where the transport isn't actually HTTP (`stub.__rpc()` uses structured clone, `stub.fetch()` goes through CF internal routing) so there are no Effect HTTP client/server spans to orphan. `toHttpApp` was just a convenient handler format on the server side.

But I think you're right that adding an opt-in to `toHttpApp` is the wrong solution — it muddies the API for a non-HTTP use case. We ended up building a custom `RpcServer.Protocol` instead, which gives us full control over `supportsSpanPropagation` without needing any upstream changes. Closing this.

PR #1381, opened 2026-02-21
EFF-508 Fix UrlParams.Input to allow interfaces

tim-smart:
## Summary
- keep the `HttpClient` / `HttpClientRequest` / `HttpServerResponse` public signatures non-generic by removing the branch-wide generic plumbing added in the first attempt
- update `UrlParams.Input` to use a mapped `CoercibleRecord` shape plus non-generic `Input`, which accepts interface-typed params in HTTP helpers without threading call-site generics
- exclude iterable values from the `CoercibleRecord` branch so tuple inputs are validated by the iterable variant instead of being accepted as records
- expand dtslint coverage for `HttpClient` `urlParams` to include interface records, coercible nested records/arrays, iterable tuple entries, `URLSearchParams`, and negative type assertions for invalid inputs
- include a patch changeset for `effect`

## Validation
- `pnpm lint-fix`
- `pnpm test packages/effect/test/unstable/http/HttpClientRequest.test.ts`
- `pnpm test packages/effect/test/unstable/http/UrlParams.test.ts`
- `pnpm test-types packages/effect/dtslint/unstable/http/HttpClient.tst.ts`
- `pnpm check`
- `pnpm build`
- `pnpm docgen`

### DIFF
--- a/packages/effect/dtslint/unstable/http/HttpClient.tst.ts
+++ b/packages/effect/dtslint/unstable/http/HttpClient.tst.ts
@@ -78,5 +78,17 @@ describe("HttpClient", () => {
         Effect.Effect<HttpClientResponse.HttpClientResponse, HttpClientError.HttpClientError, HttpClient.HttpClient>
       >()
     })
+
+    it("should reject non-urlParams input", () => {
+      // @ts-expect-error!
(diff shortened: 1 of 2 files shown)


tim-smart (commented):
Need to find a way of doing it without introducing genetics

PR #1400, opened 2026-02-21
fix(opentelemetry): use monotonic clock for log timestamps

sellooh:
## Summary

- Logger used `options.date` (`Date.now()`, wall clock, ms precision) for log `timestamp`, while Tracer used `clock.currentTimeNanosUnsafe()` (monotonic clock, ns precision) for span `startTime`
- This caused logs to appear ~5-10ms before their parent span in observability UIs due to drift between `Date.now()` and `performance.now()` in Node.js
- Both now use the same monotonic clock source via `nanosToHrTime(clock.currentTimeNanosUnsafe())`
- Extracted `nanosToHrTime` from `Tracer.ts` to shared `internal/attributes.ts` to avoid duplication

## Test plan

- [x] Existing `Logger.test.ts` passes (2 tests)
- [x] Existing `Tracer.test.ts` passes (8 tests)
- [x] `pnpm check` passes
- [x] `pnpm lint-fix` passes
- [x] Verified in production: log timestamps now align with span timestamps in HyperDX/ClickStack UI

Before, the log was out of position:
Screenshot 2026-02-21 at 18 10 18

After, log appears under the service, after the repository finished saving a result as expected:
Screenshot 2026-02-21 at 18 09 49

```typescript
const createKey = Effect.fn("ApiKeyService.createKey")(
  function* (accountId: AccountIdType, name: string) {
    yield* Effect.annotateCurrentSpan("accountId", accountId);
    yield* Effect.annotateCurrentSpan("name", name);

    const token = generateToken();
    const tokenHash = hashToken(token);
    const tokenPrefix = token.slice(0, 12);

    const apiKey = yield* repo.create(accountId, name, tokenHash, tokenPrefix);

    yield* Effect.annotateCurrentSpan("tokenPrefix", tokenPrefix);

    yield* Effect.log(
      `API key created: name="${name}" prefix="${tokenPrefix}" accountId="${accountId}"`
    );

   // cont.
  },
  (effect) =>
    effect.pipe(
      Effect.tapError((e) =>
        Effect.logError("[ApiKeyService] createKey failed", { error: String(e) })
      )
    )
);
```

### DIFF
--- a/packages/opentelemetry/src/Logger.ts
+++ b/packages/opentelemetry/src/Logger.ts
@@ -60,13 +60,13 @@ export const make: Effect.Effect<
     // }
 
     const message = Arr.ensure(options.message).map(unknownToAttributeValue)
-    const timestamp = clock.currentTimeNanosUnsafe()
+    const hrTime = nanosToHrTime(clock.currentTimeNanosUnsafe())
     otelLogger.emit({
       body: message.length === 1 ? message[0] : message,
       severityText: options.logLevel,
       severityNumber: LogLevel.getOrdinal(options.logLevel),
-      timestamp: nanosToHrTime(timestamp),
-      observedTimestamp: nanosToHrTime(timestamp),
+      timestamp: hrTime,
+      observedTimestamp: hrTime,
       attributes
     })
   })


packages/opentelemetry/src/Logger.ts
```diff
@@ -60,12 +60,13 @@ export const make: Effect.Effect<
     // }
 
     const message = Arr.ensure(options.message).map(unknownToAttributeValue)
+    const timestamp = clock.currentTimeNanosUnsafe()
```
tim-smart:
```suggestion
    const hrTime = nanosToHrTime(clock.currentTimeNanosUnsafe())
```

PR #1419, opened 2026-02-22
Add @effect/atom-react-devtools package

candrewlee14:
## Summary

- New `@effect/atom-react-devtools` package: floating devtools panel for inspecting Effect atoms
- Drop `<AtomDevtools />` inside a `<RegistryProvider>` to get a searchable atom list with value inspector
- Inline editable JSON tree: double-click any leaf value to edit in-place with type-aware parsing
- Array/object manipulation: add/remove items and keys via inline controls
- Clickable dependency/dependent links to navigate between atoms
- Slide-in/out panel animations with CSS transitions
- Effect logo toggle button (black bubble, hover scale)
- Auto tree-shaken in production (`process.env.NODE_ENV === "production"` guard)
- Stacked on PR 1 (`47c8531f`) which added `Node.parents/children/state/listenerCount` to AtomRegistry

## Test plan

- [x] `pnpm lint-fix` passes
- [x] `pnpm check` passes
- [x] `pnpm build` passes
- [x] 17 tests pass covering: toggle, open/close, localStorage persistence, atom display, search filtering, detail panel, inline editing (enter/escape/nested), array add/remove, object add key/remove key, refresh derived atoms
- [x] Manual: verify panel slide animations in browser
- [x] Manual: verify inline editing UX on various atom types


tim-smart:
I think the design will need re-thinking. 

Atom isn't just used on the web, so I think some shared service will need to be introduced, so if you are using atom with opentui etc. you can still use the dev tools.

It should also be possible to use the dev tools with vue or solid.

candrewlee14:
Gotcha, I can rework this. Tanstack devtools does a similar thing with having some core agnostic pkgs, then per-framework devtools pkgs.
https://github.com/TanStack/devtools/tree/main/packages

candrewlee14:
Updated that, is this more what you're looking for? @tim-smart

tim-smart:
Hmm still doesn't feel quite right yet. I'll set some time aside after the beta stabilises to look at this in more detail.

IMax153:
Hi @candrewlee14,

PR #1425, opened 2026-02-22
Fix approval artifact stripping across multi-round conversations

candrewlee14:
## Summary
- Strip resolved approval artifacts unconditionally, not only when the current round has pending approvals. Previously, stale artifacts from earlier rounds leaked to the provider causing errors.
- In streaming mode, emit pre-resolved tool results as stream parts so `Chat.streamText` persists them to history, preventing re-resolution on subsequent rounds.

## Test plan
- [x] Existing `LanguageModel.test.ts` tests pass (11/11)

### DIFF
--- a/.changeset/fix-strip-approval-artifacts-multi-round.md
+++ b/.changeset/fix-strip-approval-artifacts-multi-round.md
@@ -1,0 +1,13 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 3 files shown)


IMax153:
I think this needs some tests to ensure we don't regress.

candrewlee14:
Thanks, added some tests

PR #1433, opened 2026-02-23
EFF-524 Add Effect.annotateLogsScoped

tim-smart:
## Summary
- add `Effect.annotateLogsScoped` to install log annotations for the lifetime of a `Scope` and automatically restore previous annotations on scope close
- keep `Effect.annotateLogs` on `updateService` semantics and tighten logger regression coverage to assert scoped annotations compose with temporary `annotateLogs` values before close
- include runtime coverage in `packages/effect/test/Logger.test.ts`, dtslint coverage for both `annotateLogsScoped` overloads, and a patch changeset for `effect`

## Validation
- pnpm lint-fix
- pnpm test packages/effect/test/Logger.test.ts
- pnpm test-types packages/effect/dtslint/Effect.tst.ts
- pnpm check
- pnpm build
- pnpm docgen

### DIFF
--- a/packages/effect/test/Logger.test.ts
+++ b/packages/effect/test/Logger.test.ts
@@ -29,10 +29,12 @@ describe("Logger", () => {
   it.effect("annotateLogsScoped", () =>
     Effect.gen(function*() {
       const scope = yield* Scope.make()
-      assert.deepStrictEqual(yield* References.CurrentLogAnnotations, {})
-
-      yield* Effect.annotateLogsScoped("requestId", "req-1").pipe(Scope.provide(scope))
-      assert.deepStrictEqual(yield* References.CurrentLogAnnotations, { requestId: "req-1" })
+      const annotations = yield* Effect.gen(function*() {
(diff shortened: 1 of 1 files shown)


packages/effect/src/Effect.ts
```diff
@@ -13031,6 +13031,52 @@ export const annotateLogs = dual<
...
+    : arguments[0]
+
+  return internal.uninterruptible(
+    core.withFiber((fiber) => {
+      const scope = ServiceMap.getUnsafe(fiber.services, internal.scopeTag)
+      const current = fiber.getRef(CurrentLogAnnotations)
+      fiber.setServices(ServiceMap.add(fiber.services, CurrentLogAnnotations, { ...current, ...values }))
+      return internal.scopeAddFinalizerExit(
+        scope,
+        () => internal.sync(() => fiber.setServices(ServiceMap.add(fiber.services, CurrentLogAnnotations, current)))
```
tim-smart: I think this will lead to issues when the Scope closes after another Effect.annotateLogs call.

PR #1468, opened 2026-02-24
fix Rpc.ExtractProvides to use middleware service ID

lucas-barake:


packages/effect/dtslint/unstable/rpc/Rpc.tst.ts
```diff
@@ -0,0 +1,20 @@
...
+type Mixed = typeof GetUser | typeof StreamEvents
+
+describe("Rpc", () => {
+  describe("IsStream", () => {
+    it("returns true for stream RPCs in a mixed group", () => {
+      expect<Rpc.IsStream<Mixed, "StreamEvents">>().type.toBe<true>()
+    })
+
+    it("returns never for non-stream RPCs in a mixed group", () => {
+      expect<Rpc.IsStream<Mixed, "GetUser">>().type.toBe<never>()
```
tim-smart: It returns never intentionally to prevent typescript merging into a `boolean` type.

PR #1487, opened 2026-02-25
refactor: make Statement a Yieldable instead of an Effect subtype

mattiamanzati:
## Summary

`Statement` is no longer an `Effect` subtype. Instead, it implements `Effect.Yieldable`, which means:

- **Yielding in generators still works** — `Statement` remains yieldable in `Effect.gen` blocks, so `yield* sql\`...\`` continues to work as before.
- **Piping into Effect combinators requires `.asEffect()`** — since `Statement` no longer extends `Effect`, callers must explicitly convert via `.asEffect()` before using `Effect.flatMap`, `Effect.andThen`, `Effect.pipe`, `sql.withTransaction`, etc.

### Why

Decoupling `Statement` from the `Effect` type hierarchy makes the boundary between the SQL DSL and the Effect runtime explicit. Statements are descriptions of queries; converting them into executable effects is now an intentional step rather than an implicit subtype relationship. This simplifies the internal prototype chain (removing `core.EffectProto` / `core.evaluate` from `StatementProto`) and aligns with the direction of making yieldables a first-class, opt-in concept.

### Changes

- **`packages/effect/src/unstable/sql/Statement.ts`** — `Statement<A>` now extends `Fragment, Effect.Yieldable<...>, Pipeable, Inspectable` instead of `Fragment, Effect.Effect<...>, ...`. The `evaluate` symbol on the prototype is replaced by an `asEffect()` method that wraps the query execution in a span.
- **All SQL consumer modules** (`SqlMessageStorage`, `SqlRunnerStorage`, `SqlEventLogJournal`, `SqlEventLogServer`, `PersistedQueue`, `Persistence`, `Migrator`, `SqlModel`) — added `.asEffect()` calls wherever a `Statement` was previously used directly as an `Effect`.
- **Test files** (`d1`, `libsql`, `mysql2`, `sqlite-node`) — updated to call `.asEffect()` before piping into Effect operators.

### DIFF
--- a/.changeset/jolly-drinks-create.md
+++ b/.changeset/jolly-drinks-create.md
@@ -1,0 +1,9 @@
+---
+"@effect/sql-sqlite-node": patch
+"@effect/sql-libsql": patch
+"@effect/sql-mysql2": patch
+"effect": patch
+"@effect/sql-d1": patch
+---
+
+Remove subtyping of Statement in favour of Effect's Yieldability


tim-smart:
I had kept sql statements as effects as the DX of calling .asEffect() everywhere was a bit weird. Also sql statements are closer to "effects with extra properties" rather than another data type.

mattiamanzati:
Yeah that's a point 💪

PR #1491, opened 2026-02-25
pass full error to catchReason/catchReasons callbacks across Effect, …

mattiamanzati:
…Channel, and Stream

## Type

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Another option would be to instead of passing the plain reason to the handler, change it to pass in the error & {reason: PickedReason }, but I kinda see good use cases for both, wdyt?

### DIFF
--- a/packages/effect/dtslint/Effect.tst.ts
+++ b/packages/effect/dtslint/Effect.tst.ts
@@ -11,8 +11,12 @@ class QuotaExceededError extends Data.TaggedError("QuotaExceededError")<{
   readonly limit: number
 }> {}
 
+class UnknownAiModelError extends Data.TaggedError("UnknownAiModelError")<{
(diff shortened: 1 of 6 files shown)


mikearnaldi:
We could also pass `error & { reason: x }` to narrow it

mattiamanzati:
Yeah, added type utils as well for it

alajos00:
## 🛑 Missing Changeset Alert: @mattiamanzati 

It looks like this PR is missing a **Changeset**. Since this includes logic refinements for the AI `LanguageModel` utility types, we need a changeset to ensure these improvements are captured in the next release.

### 🛠️ How to fix:
You can generate one locally by running:
```bash
npx changeset

PR #1534, opened 2026-02-27
Separate Service from ServiceImpl to allow custom ServiceImpl.

patroza:
I understand the motivation to unify all the Tag implementations of v3 into a single Service implementation in v4.
However the current implementation also prevents any customisations that users might want to do in app/user land, like making an Opaque version, where `of` will take Shape, but returns the Opaque Identifier. 
Trying to override these with custom versions will lead to inheritance and variance conflicts.

By separating the core service interface from the implementation specifics (`of`/`use` etc), we allow the user to customize the experience to his needs.

Notes about the PR:
- current state is illustrative only to gauge interest and opinions
- naming; perhaps instead of `Service` vs `ServiceImpl`, maybe inverting it may be nicer? `ServiceKey` and `Service` for instance

### DIFF
--- a/packages/effect/test/Effect.test.ts
+++ b/packages/effect/test/Effect.test.ts
@@ -160,6 +160,13 @@ describe("Effect", () => {
       Effect.provideService(ATag, ATag.of({ test: Effect.succeed("A" as const) })),
       Effect.runPromise
     )
+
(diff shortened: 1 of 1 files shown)


tim-smart:
This might be better https://github.com/Effect-TS/effect-smol/pull/1536

IMax153:
Closing in favor of the work @tim-smart did in https://github.com/Effect-TS/effect-smol/pull/1536.

PR #1557, opened 2026-02-28
fix(mcp): check isParam on original schema in compileUriTemplate

### LINKED ISSUE #1540: `isParam` check fails on `toCodecStringTree` result, causing parameter name resolution to break in MCP resource templates

4.0.0-beta.19

1. Define an MCP resource template with a parameter and completion:
   ```typescript
   const idParam = McpSchema.param("id", Schema.NumberFromString);
   const ReadmeTemplate = McpServer.resource`file://readme/${idParam}`({
     name: "README Template",
     completion: {
       id: (_) => Effect.succeed([1, 2, 3, 4, 5])
     },
     content: Effect.fn(function* (_uri, id) {
       return `# MCP Server Demo - ID: ${id}`;
     })
   });
   ```
2. Start the MCP server using Effect v4 beta.
3. The server crashes with `TypeError: undefined is not an object (evaluating 'schema.ast')`.

The parameter name should be correctly resolved as `"id"`, and completions should work without errors.

### What do you see instead?

The server fails to start due to the undefined schema AST error.

### Additional information

**Description:**  
When defining MCP resource templates with parameters using `McpSchema.param`, the completion functionality fails with a `TypeError: undefined is not an object (evaluating 'schema.ast')` because the parameter name lookup returns `undefined`, leading to `Schema.Array(undefined)` being called.

**Root Cause:**  
In the `compileUriTemplate` function, `Schema.toCodecStringTree(schemas[i])` creates a new schema that does not preserve the custom `Param` properties (including the `name` field). As a result, `isParam(toCodecStringTree)` always returns `false`, causing the parameter name to default to `"param${key}"` (e.g., `"param0"`) instead of the intended name (e.g., `"id"`). When completions are processed, `params["id"]` is `undefined`, and `Schema.Array(undefined)` fails when trying to access `undefined.ast`.

**Environment:**  
- Effect version: 4.0.0-beta.19
- Platform: Node.js/Bun
- OS: Linux

**File/Function/Line:**  
- **File:** `packages/effect/src/unstable/ai/McpServer.ts`
- **Function:** `compileUriTemplate`
- **Line:** 1032

**Suggested Fix:**  
Change line 1032 from:
```typescript
const paramName = isParam(toCodecStringTree) ? toCodecStringTree.name : `param${key}`;
```
To:
```typescript
const paramName = isParam(schemas[i]) ? schemas[i].name : `param${key}`;
```
This ensures the `isParam` check is performed on the original schema before transformation, preserving the parameter name for correct lookup in completions.

A386official:
Hi! I found the bug reported in #1540 and this is a minimal one-line fix for it.

## What
MCP resource template parameter names are always resolved as `param0`, `param1`, etc. instead of the actual names passed to `McpSchema.param()`. This causes `TypeError: undefined is not an object` when completions try to look up parameters by their intended name.

## Why
In `compileUriTemplate`, the `isParam` check is performed on the result of `Schema.toCodecStringTree(schemas[i])`. The `toCodecStringTree` transformation creates a new schema object that does not preserve the custom `Param` properties (`ParamSchemaTypeId`, `name`, `schema`). As a result, `isParam(toCodecStringTree)` always returns `false`, and the parameter name falls back to `param${key}`.

When completions are later processed, they look up `params["id"]` (the actual param name), but the key stored is `"param0"`, so `params["id"]` is `undefined`, causing `Schema.Array(undefined)` to crash.

## How
Check `isParam(schemas[i])` (the original schema) instead of `isParam(toCodecStringTree)` (the transformed schema). The original schema retains the `Param` marker and `name` property, so the parameter name is correctly resolved.

Closes #1540

### DIFF
--- a/packages/effect/test/unstable/ai/McpServer.test.ts
+++ b/packages/effect/test/unstable/ai/McpServer.test.ts
@@ -75,11 +75,13 @@ describe("McpServer", () => {
         return handler(request)
       }
 
+      const clientLayer = RpcClient.layerProtocolHttp({ url: "http://localhost/mcp" }).pipe(
+        Layer.provideMerge(RpcSerialization.layerJsonRpc()),
+        Layer.provideMerge(FetchHttpClient.layer),
+        Layer.provideMerge(Layer.succeed(FetchHttpClient.Fetch, customFetch))
+      )
       const client = yield* RpcClient.make(McpSchema.ClientRpcs).pipe(
-        Effect.provide(RpcClient.layerProtocolHttp({ url: "http://localhost/mcp" })),
-        Effect.provide(RpcSerialization.layerJsonRpc()),
-        Effect.provide(FetchHttpClient.layer),
-        Effect.provideService(FetchHttpClient.Fetch, customFetch)
+        Effect.provide(clientLayer)
       )
 
       const result = yield* client["tools/list"]({})


A386official:
Added two regression tests in `McpServer.test.ts`:

1. **param name in URI template** — registers a resource template with `McpSchema.param("id", Schema.Number)` and asserts the listed `uriTemplate` contains `{id}` (not `{param0}`)
2. **completion resolves named param** — registers the same template with a completion handler keyed on `"id"`, calls `completion/complete` with `argument.name: "id"`, and asserts the expected values are returned

Both tests would fail without the one-line fix (the transformed schema loses the Param marker, so `isParam` returns false and the name falls back to `param0`).

packages/effect/test/unstable/ai/McpServer.test.ts
```diff
@@ -9,6 +9,35 @@ import * as HttpRouter from "effect/unstable/http/HttpRouter"
...
+    const customFetch: typeof fetch = (input, init) => {
+      const request = input instanceof Request ? input : new Request(input, init)
+      return handler(request)
+    }
+
+    const client = yield* RpcClient.make(McpSchema.ClientRpcs).pipe(
+      Effect.provide(RpcClient.layerProtocolHttp({ url: "http://localhost/mcp" })),
+      Effect.provide(RpcSerialization.layerJsonRpc()),
+      Effect.provide(FetchHttpClient.layer),
+      Effect.provideService(FetchHttpClient.Fetch, customFetch)
```
IMax153: Although multiple `Effect.provide`s share a `MemoMap` in Effect v4, this still does not reflect best practices. 

The layers should be composed separately, and then provided to the program via a single `Effect.provide`.

A386official:
Nicee

PR #1563, opened 2026-03-01
fix: RateLimiter.layerStoreMemory shares state across layer provisions

### LINKED ISSUE #1562: RateLimiter.layerStoreMemory creates independent state when layer is provided multiple times — no shared rate limiting across consumers

4.0.0-beta.20

```
import { Effect, Layer, ServiceMap, Stream } from "effect";
import { RateLimiter } from "effect/unstable/persistence";

const LIMIT = 5;
const WINDOW = "10 seconds";
const CONCURRENCY = 5;

// --- Service with rate-limited method ---

interface ApiShape {
  readonly request: Effect.Effect<void, RateLimiter.RateLimiterError>;
}

class Api extends ServiceMap.Service<Api, ApiShape>()("repro/Api") {}

let count = 0;
const start = Date.now();

const ApiLayer = Layer.effect(
  Api,
  Effect.gen(function* () {
    const withLimiter = yield* RateLimiter.makeWithRateLimiter;

    const request = Effect.sync(() => {
      count++;
      const elapsed = ((Date.now() - start) / 1000).toFixed(2);
      console.log(`request #${count} at +${elapsed}s`);
    }).pipe(
      withLimiter({
        key: "test",
        limit: LIMIT,
        window: WINDOW,
        algorithm: "fixed-window",
        onExceeded: "delay",
      }),
    );

    return Api.of({ request });
  }),
).pipe(
  Layer.provide(RateLimiter.layer),
  Layer.provide(RateLimiter.layerStoreMemory),
);

// --- Two consumers that each provide ApiLayer independently ---

const phaseA = Effect.gen(function* () {
  const api = yield* Api;
  console.log("=== phase A (3 requests) ===");
  for (let i = 0; i < 3; i++) {
    yield* api.request;
  }
});

const phaseB = Effect.gen(function* () {
  const api = yield* Api;
  console.log("=== phase B (7 requests) ===");
  yield* Stream.fromIterable(Array.from({ length: 7 })).pipe(
    Stream.mapEffect(() => api.request, { concurrency: CONCURRENCY }),
    Stream.runDrain,
  );
});

// BUG: providing ApiLayer separately to each consumer
// creates two independent rate limiter stores
  yield* phaseA.pipe(Effect.provide(ApiLayer));
  yield* phaseB.pipe(Effect.provide(ApiLayer));

  console.log(`\nTotal: ${count} requests (limit was ${LIMIT} per ${WINDOW})`);
  console.log(`Expected: requests 6-10 delayed to next window`);
});

Effect.runPromise(program).catch(console.error);
```

Requests 1–5 execute immediately, requests 6–10 are delayed to the next window (+10s). The rate limiter should enforce a global limit of 5 requests per 10 seconds regardless of how many times the layer is provided.

### What do you see instead?

```
=== phase A (3 requests) ===
request #1 at +0.00s
request #2 at +0.00s
request #3 at +0.00s
=== phase B (7 requests) ===
request #4 at +0.01s
request #5 at +0.01s
request #6 at +0.01s
request #7 at +0.01s
request #8 at +0.01s
request #9 at +10.01s
request #10 at +10.01s
```

8 requests execute in the first window instead of 5. Each Effect.provide(ApiLayer) builds a fresh layerStoreMemory, creating independent Map instances that don't share state. Phase B gets its own budget of 5, unaware of phase A's 3 consumed requests.

### Additional information

- layerStoreMemory uses Layer.sync which creates new in-memory Map instances per layer build. When the layer is provided to multiple consumers independently, each gets its own counter maps.
- In v3, RateLimiter.make was a self-contained effect with inline state — no separate store layer to accidentally duplicate. The v4 architecture splits the store into RateLimiterStore (via layerStoreMemory) which introduces this footgun.

In real-world usage, this manifests when a service wraps an API client with a rate limiter and is then consumed in multiple places — each independently yielding the layer. Because each provision creates its own rate limiter state, the combined throughput exceeds the API's server-side rate limit, resulting in HTTP 429 errors from the upstream service. The client-side rate limiter gives no indication that anything is wrong, since each instance stays within its own budget.

lindit:
## Summary

- Hoist `fixedCounters` and `tokenBuckets` Maps to module scope so `layerStoreMemory` shares rate limiter state across separate `Effect.provide` calls
- Add regression tests for both fixed-window and token-bucket state sharing
- Use unique keys per existing test to prevent cross-test state pollution

**Similar issues may still occur in:** 
`packages/effect/src/unstable/persistence/KeyValueStore.ts`
`packages/effect/src/unstable/persistence/PersistedQueue.ts`
`packages/effect/src/unstable/persistence/Persistence.ts`

Closes #1562

### DIFF
--- a/.changeset/fix-ratelimiter-shared-memory-store.md
+++ b/.changeset/fix-ratelimiter-shared-memory-store.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+fix RateLimiter.layerStoreMemory creating independent state per layer build
--- a/packages/effect/src/unstable/persistence/RateLimiter.ts
+++ b/packages/effect/src/unstable/persistence/RateLimiter.ts
@@ -459,56 +459,57 @@ export class RateLimiterStore extends ServiceMap.Service<
  */
 export const layerStoreMemory: Layer.Layer<
   RateLimiterStore
-> = Layer.sync(RateLimiterStore, () => {
+> = (() => {
   const fixedCounters = new Map<string, { count: number; expiresAt: number }>()
   const tokenBuckets = new Map<string, { tokens: number; lastRefill: number }>()
 
-  return RateLimiterStore.of({
-    fixedWindow: (options) =>
-      Effect.clockWith((clock) =>
-        Effect.sync(() => {
-          const refillRateMillis = Duration.toMillis(options.refillRate)
-          const now = clock.currentTimeMillisUnsafe()
-          let counter = fixedCounters.get(options.key)
-          if (!counter || counter.expiresAt <= now) {
-            counter = { count: 0, expiresAt: now }
-            fixedCounters.set(options.key, counter)
-          }
-          if (options.limit && counter.count + options.tokens > options.limit) {
-            return [counter.count + options.tokens, counter.expiresAt - now] as const
-          }
-          counter.count += options.tokens
-          counter.expiresAt += refillRateMillis * options.tokens
-          return [counter.count, counter.expiresAt - now] as const
-        })
-      ),
-    tokenBucket: (options) =>
-      Effect.clockWith((clock) =>
-        Effect.sync(() => {
-          const refillRateMillis = Duration.toMillis(options.refillRate)
-          const now = clock.currentTimeMillisUnsafe()
-          let bucket = tokenBuckets.get(options.key)
-          if (!bucket) {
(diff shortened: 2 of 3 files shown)


tim-smart:
The original behaviour is intended and something we would like to preserve. You will need to share the Layer using something like a ManagedRuntime.

PR #1568, opened 2026-03-01
fix(cli): non-zero exit on parse errors, Command.exit, withErrorHandler

cevr:
Noticed that `Command.runWith` exits 0 on parse errors (unknown flags, unknown subcommands, missing required values) because `showHelp` always succeeds after displaying the error. This means downstream tooling (CI, scripts, shells) can't distinguish between a successful run and a malformed invocation.

Totally optional - feel free to close if this isn't the direction you want to go. Just figured I'd put it up since the fix is small.

### Non-zero exit on parse errors
`showHelp` now fails with all parse errors via `Effect.failCause(Cause.fromReasons(...))` after printing them to stderr, so runtimes (`BunRuntime.runMain` / `NodeRuntime.runMain`) exit non-zero. When there are no errors (clean `--help`, no-handler commands), it still succeeds and exits 0.

### `Command.exit(code)`
New API to terminate the CLI with a specific exit code from handlers. `Command.run` intercepts the `CliExit` signal and calls `process.exit(code)`. `Command.runWith` propagates it as a `CliExit` error for testability.

### `Command.withErrorHandler`
Post-hook that runs after default error display (help + stderr) on parse errors. The default behavior is preserved; the hook runs afterward for custom exit codes or additional error handling.

### Error handler resolution from command tree
`withErrorHandler` now resolves leaf-to-root — a subcommand's handler takes precedence over the root's. Root handler acts as fallback.

### `CliExit` separated from `CliError` union
`CliExit` is a control flow signal, not a parse error. Removed from the `CliError` union. Added `CliExit.is` static guard via `Schema.is`.

> [!NOTE]
> AI (Claude) was used under human supervision to develop this PR.

### DIFF
--- a/packages/effect/src/unstable/cli/CliError.ts
+++ b/packages/effect/src/unstable/cli/CliError.ts
@@ -499,19 +499,18 @@ export class UserError extends Schema.ErrorClass(`${TypeId}/UserError`)({
  *
  * const deploy = Command.make("deploy", {
  *   env: Flag.string("env")
- * }, (config) =>
- *   Effect.gen(function*() {
- *     if (config.env === "production") {
- *       yield* Command.exit(1)
- *     }
- *   }))
+ * }).pipe(Command.withHandler(Effect.fnUntraced(function*(config) {
+ *   if (config.env === "production") {
+ *     yield* Command.exit(1)
+ *   }
+ * })))
  * ```
  *
  * @since 4.0.0
  * @category models
  */
-export class CliExit extends Schema.ErrorClass(`${TypeId}/CliExit`)({
-  _tag: Schema.tag("CliExit"),
+export class ExitCode extends Schema.ErrorClass(`${TypeId}/ExitCode`)({
+  _tag: Schema.tag("ExitCode"),
   code: Schema.Number
 }) {
   /**
@@ -523,11 +522,15 @@ export class CliExit extends Schema.ErrorClass(`${TypeId}/CliExit`)({
    * @since 4.0.0
    */
   override get message() {
-    return `CLI exiting with code ${this.code}`
+    return `ExitCode(${this.code})`
   }
-
-  /**
-   * @since 4.0.0
-   */
-  static is = Schema.is(CliExit)
 }
+
+/**
+ * Type guard to check if a value is an `ExitCode`.
+ *
+ * @since 4.0.0
+ * @category guards
+ */
+export const isExitCode = (u: unknown): u is ExitCode =>
+  Predicate.hasProperty(u, TypeId) && Predicate.isTagged(u, "ExitCode")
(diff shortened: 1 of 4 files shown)


IMax153:
@cevr - I was thinking we might want to add an external API for folks that would force the CLI to exit with a non-zero status code for situations like this. At the moment, the only way to do this is to actually fail the program, but it would be nice if there was a programmatic way to do this from the CLI.

Do you think we could re-work this PR in that direction instead?

cevr:
@IMax153 yeah that makes sense, will explore that area

cevr:
@IMax153 reworked the PR again. instead of making parse errors fail by default, this adds three composable primitives:

- `CliError.CliExit` - typed error carrying an exit code
- `Command.exit(code)` - fails with `CliExit`, usable from handlers or error hooks
- `Command.withErrorHandler(fn)` - post-hook that runs after the default error display (help + stderr) on parse errors. receives `(errors, commandPath)`, returns an `Effect`

`Command.run` intercepts `CliExit` and calls `process.exit(code)`. `Command.runWith` propagates it as a typed error for testability. default behavior is unchanged (exit 0).

to opt into non-zero exit on parse errors:

```ts
const app = Command.make("app", { ... }, handler).pipe(
  Command.withErrorHandler(() => Command.exit(1))
)
```

what do you think of this approach?

cevr:
@IMax153 updated!

i had thought about withErrorHandler being more true to its name. Thoughts on that route? 

Export a Command.showParseErrors helper that does the "print help + stderr errors" dance, so consumers can call it themselves and whatever else they want

IMax153:
> i had thought about withErrorHandler being more true to its name. Thoughts on that route?

That might be, but it is not semantic from a user perspective. Looking at that code it is very difficult to understand that usage of `Command.withErrorHandler` semantically means that branch would only run on parse errors.

I think that it would be good for the end user if a combinator is exposed that makes it abundantly clear that the function provided to the combinator is only executing when the command line fails to be parsed.

Whether we also add controls for actually displaying the help text in addition is up for debate. For example, we could think about:

```ts
Command.onParseError(() => /* ... logic to run on parse errors */, {
  showHelp: true // user indicates that if a parse error occurs the help should be shown
})
```

I don't know the correct approach to be honest, I'm just suggesting APIs. But what I can say is `withErrorHandler` does not feel intuitive.

cevr:
@IMax153 yeah that's fair.

what about something like this

```ts
Command.onParseError((errors, commandPath, showHelp) =>
  Effect.gen(function*() {
    yield* showHelp // print help + errors to stderr, exactly like default
    yield* Command.exit(1)
  })
)

// Or skip help entirely
Command.onParseError((errors, _commandPath) =>
  Effect.gen(function*() {
    yield* Console.error(errors[0].message)
    yield* Command.exit(2)
  })
)

```

something like that, allows consumers to control when exactly they want to show help etc

IMax153:
@cevr - I think @tim-smart just did some work to allow customizing exit codes via a trait in https://github.com/Effect-TS/effect-smol/pull/1599. 

I won't have the opportunity to take a look today, but does this satisfy your use case and invalidate this work?

If not, perhaps we can hook into this new functionality to add what you need?

cevr:
@IMax153 thanks for the update - yeah that solves this use case!

PR #1570, opened 2026-03-01
feat(schema): add Schema.opaqueBrand for fully opaque primitive types

kitlangton:
NOTICE: this PR description was written by robots under the supervision of Kit Langton

## Type
- [x] Feature

## Description
This PR adds `Schema.opaqueBrand` for fully opaque primitive types.

- Runtime behavior matches `Schema.brand` (same validation / AST branding).
- Type output is opaque: `Brand.Brand<B>` instead of `T & Brand.Brand<B>`.
- Type-level behavior is rebranding: previous brand composition is replaced by the new opaque brand.

## Why
In incremental migrations, we want strict boundaries between new typed IDs and old stringly APIs.

Example: `WorkspaceId` should not be passable to legacy functions like `directoryId: string`.
`Schema.opaqueBrand` enforces that boundary.

## Example
```ts
type WorkspaceIdTransparent = string & Brand.Brand<"WorkspaceId">
type WorkspaceIdOpaque = Brand.Brand<"WorkspaceId">

declare const legacyDirectoryApi: (directoryId: string) => void
declare const workspaceIdOpaque: WorkspaceIdOpaque

legacyDirectoryApi(workspaceIdOpaque) // type error
```

## Tests
- Added runtime coverage for `Schema.opaqueBrand` branding metadata.
- Added dtslint coverage for opaque output type, rebranding behavior, and non-assignability to stringly APIs.

### DIFF
--- a/.changeset/tame-wombats-juggle.md
+++ b/.changeset/tame-wombats-juggle.md
@@ -2,4 +2,4 @@
 "effect": minor
 ---
 
-Add `Schema.opaqueBrand` for fully opaque primitive types. It keeps `Schema.brand` runtime behavior while returning an opaque `Brand.Brand<B>` type instead of exposing the primitive base type.
+Add `Schema.newtype` for fully opaque types with `NewtypeBrand<K, From>` carrier. Unlike `Schema.brand`, which exposes the underlying type, `newtype` produces a fully opaque type while carrying the original type as a phantom for recovery via `NewtypeFrom<T>`. Uses a separate `newtypes` annotation independent of `brands`.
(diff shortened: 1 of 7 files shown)


gcanti:
If I understand correctly, this is a request for a newtype implementation, right?
If so, I agree with the Copilot observation. It should use a different runtime implementation, similar to `Schema.brand`, but with a different annotation to accumulate identifiers. I would also explicitly name it "newtype" since that is what it is.

mikearnaldi:
> If I understand correctly, this is a request for a newtype implementation, right? If so, I agree with the Copilot observation. It should use a different runtime implementation, similar to `Schema.brand`, but with a different annotation to accumulate identifiers. I would also explicitly name it "newtype" since that is what it is.

Also it should probably carry the original type as a reference, like the old newtype implementation did

gcanti:
I have added a Newtype module that can be used for this https://github.com/Effect-TS/effect-smol/pull/1672

kitlangton:
Thanks @gcanti!

PR #1574, opened 2026-03-01
EFF-611 Fix LanguageModel.generateText toolkit error channel

tim-smart:
## Summary
- remove `AiErrorReason` from `Tool.HandlerError` so toolkit-backed `LanguageModel.generateText` / `streamText` expose wrapped `AiError` values instead of raw reason classes
- align `LanguageModel` internal stream queue error typing with normalized tool-handler errors
- add a dtslint regression at `packages/effect/dtslint/unstable/ai/LanguageModel.tst.ts` and include an `effect` patch changeset

## Validation
- pnpm lint-fix
- pnpm test packages/effect/test/unstable/ai/Tool.test.ts
- pnpm test packages/effect/test/unstable/ai/LanguageModel.test.ts
- pnpm test-types packages/effect/dtslint/unstable/ai/LanguageModel.tst.ts
- pnpm check:tsgo
- pnpm docgen

### DIFF
--- a/.changeset/sharp-peas-march.md
+++ b/.changeset/sharp-peas-march.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix AI tool handler error typing so `LanguageModel.generateText` with a toolkit exposes wrapped `AiError` values rather than leaking raw `AiErrorReason` in the error channel.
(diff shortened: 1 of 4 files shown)


tim-smart:
@IMax153 was there a reason the AiErrorReason was exposed here?

IMax153:
> @IMax153 was there a reason the AiErrorReason was exposed here?

To allow the end user to return a reason directly instead of an `AiError`, but I think this is generally cleaner so I'm good with it 👍

PR #1579, opened 2026-03-02
EFF-613 remove fiber keepAlive mechanism

tim-smart:
## Summary
- remove the fiber-level keep-alive interval from `Effect.async` callback execution
- add keep-alive interval management to `Runtime.makeRunMain`, with safe fallback when `setInterval` is unavailable
- update `EffectKeepAlive` regression coverage to assert `makeRunMain` starts/stops keep-alive and still works when timers are blocked

## Validation
- pnpm lint-fix
- pnpm test packages/effect/test/EffectKeepAlive.test.ts
- pnpm check:tsgo
- pnpm docgen

### DIFF
--- a/packages/effect/src/Runtime.ts
+++ b/packages/effect/src/Runtime.ts
@@ -212,18 +212,12 @@ export const makeRunMain = (
           return Effect.logError(cause)
         })
       )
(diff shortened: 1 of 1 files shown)


IMax153:
@tim-smart - if we're removing keep-alive, we should remove the associated migration guide no?

PR #1583, opened 2026-03-02
feat: Support Reference classes

patroza:
As Reference is just a Service with defaultValue, it would be nice to have feature parity; in this case the ability to extend a Reference class (creating a namespace for static properties etc)


.changeset/huge-moons-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": major
```
tim-smart:
```suggestion
"effect": patch
```

PR #1593, opened 2026-03-02
fix(unstable/ai): correct jsonSchema for tools with empty parameters

lloydrichards:
## Type

- [x] Bug Fix

## Description

Since `beta.18`, when using `Schema.Struct({})` as parameter schema for an MCP tool (`Tool.make`), the MCP client can no longer read the tools because the JSON schema returned by `Tool.getJsonSchema` is no longer a plain object schema.
Example (`tools/list` payload for no‑param tools):
```txt
{ inputSchema: { anyOf: [ { type: object }, { type: array } ] } }
```
I traced the issue to #1520 where it looks like a guard clause was removed from `getJsonSchemaFromSchema`.  I've added a unit test to show that removing this guard will result in:

```bash
 FAIL   effect  test/unstable/ai/McpServer.test.ts > McpServer > includes Tool.Meta in tools/list response
AssertionError: expected { anyOf: [ { type: 'object' }, …(1) ] } to deeply equal { type: 'object' }

- Expected
+ Received

  {
+   "anyOf": [
+     {
        "type": "object",
+     },
+     {
+       "type": "array",
+     },
+   ],
  }
```

Adding back the guard will cause this to return with a valid schema type of:
```json
{
  "type": "object",
  "properties": {},
  "required": [],
  "additionalProperties": false
}
```

## Related

- Discord [link](https://discord.com/channels/795981131316985866/1477775219363152013)

### DIFF
--- a/.changeset/silent-onions-end.md
+++ b/.changeset/silent-onions-end.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+correct jsonSchema for mcp tools with empty parameters


packages/effect/src/unstable/ai/Tool.ts
```diff
@@ -1586,6 +1586,15 @@ export const getJsonSchema = <Tool extends Any>(tool: Tool, options?: {
 export const getJsonSchemaFromSchema = <S extends Schema.Top>(schema: S, options?: {
   readonly transformer?: CodecTransformer
 }): JsonSchema.JsonSchema => {
+  const props = AST.isObjects(schema.ast) ? schema.ast.propertySignatures : []
```
tim-smart: I think the fix needs to be lifted into the json schema transformer itself

lloydrichards: Have updated the `Tool.getJsonSchema()` instead of `getJsonSchemaFromSchema`

IMax153:
@lloydrichards @tim-smart  I'm not sure that this "fix" makes sense in general. 

A `Schema` is meant to represent a TypeScript type. `Schema.Struct({})` does not represent an empty object,  it technically represents the `{}` TypeScript type, which is _almost_ the same as `any`.

An empty object type is better represented by `Schema.Record(Schema.String, Schema.Never)`, which when converted to a JSON schema is exactly what you probably want:

```ts
import { Schema } from "effect"

const schema = Schema.Record(Schema.String, Schema.Never)

const doc = Schema.toJsonSchemaDocument(schema)

console.log(JSON.stringify(doc, null, 2))
/*
{
  "dialect": "draft-2020-12",
  "schema": {
    "type": "object",
    "additionalProperties":  false
  },
  "definitions": {}
}
*/
```

lloydrichards:
> @lloydrichards @tim-smart I'm not sure that this "fix" makes sense in general.

So you're suggestion would be to pass the `Schema.Record` in instead of `Schema.Struct({})`:
```ts
 const CallTool = Tool.make("CallTool", {
        description: "A test tool without parameters",
        parameters: Schema.Record(Schema.String, Schema.Never),
        success: Schema.Struct({ ok: Schema.Boolean })
      })
```
My main argument against this would be that its hard to read that the tool call doesn't expect any parameters, especially in this example where the success schema is shorter than the param one 🤔 

Doesn't it make sense to add the guard though to prevent `anyOf: []` ending up in the tool/list call since this is invalid on the McpClient and McpServer gives no indication of it being an invalid schema to use?

- https://modelcontextprotocol.io/specification/2025-11-25/server/tools#data-types
- https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-with-no-parameters

IMax153:
> Doesn't it make sense to add the guard though to prevent `anyOf: []` ending up in the tool/list call since this is invalid on the McpClient and McpServer gives no indication of it being an invalid schema to use?

Having these types of "special cases" is an anti-pattern in and of itself IMHO. I personally think it is better to keep the behavior of `Schema` is consistent everywhere and avoid adding custom checks everywhere.

Having said that, we can probably just expose a helper from the `Tool` module for cases like this:

```ts
// Tool.ts
interface EmptyParams extends Schema.$Record<Schema.String, Schema.Never> {}

const EmptyParams: EmptyParams = Schema.Record(Schema.String, Schema.Never)

// user-land.ts
import { Schema } from "effect"
import { Tool } from "effect/unstable/ai"

const Calculator = Tool.make("Calculator", {
  description: "Performs basic arithmetic operations",
  parameters: EmptyParams,
  success: Schema.Number
})

// EmptyParams
Calculator.parametersSchema
```

credit: @gcanti

lloydrichards:
fair enough 👍 i'll close this then

PR #1617, opened 2026-03-04
EFF-623 Add runForkWith runtime migration example

tim-smart:
## Summary
- expand `migration/runtime.md` with a migration section that maps v3 `Runtime.runFork(runtime)` to v4 `Effect.runForkWith(services)`
- include side-by-side v3/v4 examples and clarify that `Effect.runFork` remains the default for effects without service requirements
- add a changeset for the effect package documenting the migration guide update

## Validation
- pnpm lint-fix
- pnpm test packages/effect/test/ManagedRuntime.test.ts
- pnpm check:tsgo
- pnpm docgen

### DIFF
--- a/migration/runtime.md
+++ b/migration/runtime.md
@@ -18,8 +18,9 @@ process lifecycle utilities.
 
 ## `Runtime.runFork(runtime)` -> `Effect.runForkWith(services)`
 
-In v3, running an effect with dependencies usually meant building a
-`Runtime.Runtime<R>` value and calling `Runtime.runFork(runtime)`.
+In v3, running an effect with dependencies usually meant pulling the current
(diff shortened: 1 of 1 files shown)


migration/runtime.md
```diff
@@ -16,6 +16,59 @@ In v4, this type no longer exists and you can use `ServiceMap<R>` instead.
...
+**v3**
+
+```ts
+import { Context, Effect, Runtime } from "effect"
+
+class Logger extends Context.Tag("Logger")<Logger, {
+  readonly log: (message: string) => void
+}>() {}
+
+const runtime: Runtime.Runtime<Logger> = Runtime.defaultRuntime.pipe(
```
tim-smart: Instead of building a runtime, use `Effect.runtime<Logger>()` and run the effect inside of the main effect.

Also do the equivalent for the v4 example.

PR #1627, opened 2026-03-04
Set TaggedErrorClass error name to tag

tim-smart:
## Summary
- update `Schema.TaggedErrorClass` to assign `prototype.name` from the provided tag so instances consistently report the tagged error name
- add schema runtime coverage asserting `.name` follows the tag even when the class name differs

## Validation
- pnpm lint-fix
- pnpm test packages/effect/test/schema/Schema.test.ts
- pnpm check:tsgo
- pnpm docgen

### DIFF
--- a/.changeset/short-cows-relate.md
+++ b/.changeset/short-cows-relate.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Ensure `Schema.TaggedErrorClass` instances set `error.name` to the configured tag, even when the class name differs.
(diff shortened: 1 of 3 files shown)


gcanti:
Is there a reason we don't use the `identifier` as `name`?
if so, we could do this in the `Schema.ErrorClass` base schema

```diff
-return makeClass(core.Error, identifier, struct, annotations)
+const cls = makeClass(core.Error, identifier, struct, annotations)
+cls.prototype.name = identifier
+return cls
```

tim-smart:
> Is there a reason we don't use the identifier as name?

Usually the identifier isn't very concise as ideally it contains the package name (`@app/domain/UserError`) to prevent conflicts.

gcanti:
> Usually the identifier isn't very concise ... to **prevent conflicts**

Exactly, is that not a plus for the name?

tim-smart:
JS error names usually match the name of the constructor, which the tag usually aligns with.

I guess most people won't be setting the identifier with TaggedErrorClass regardless.

PR #1633, opened 2026-03-04
Add Schema Result decode and encode helpers

kitlangton:
## Summary
- add `SchemaParser.decodeUnknownResult` / `decodeResult` and `encodeUnknownResult` / `encodeResult` for synchronous `Result`-based parsing
- re-export the new Result-based helpers from `Schema`
- ensure `asResult` maps typed parse errors to `Result.fail(issue)` and throws for defect-only causes
- add schema tests for success / typed failure behavior and throw behavior for async decoders and missing dependencies

## Testing
- `pnpm lint-fix`
- `pnpm test packages/effect/test/schema/Schema.test.ts`
- `pnpm check:tsgo`
- `pnpm docgen`

### DIFF
--- a/.changeset/schema-result-combinators.md
+++ b/.changeset/schema-result-combinators.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Schema: add `decodeUnknownResult` / `decodeResult` and `encodeUnknownResult` / `encodeResult` helpers for synchronous `Result`-based parsing.
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -767,6 +767,18 @@ export const decodeOption = Parser.decodeOption
  * @category Decoding
  * @since 4.0.0
  */
+export const decodeUnknownResult = Parser.decodeUnknownResult
+
+/**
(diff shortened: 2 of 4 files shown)


gcanti:
@IMax153 looks good to me, but it's still marked as draft, not sure whether it's just an oversight

@kitlangton ?

kitlangton:
> @IMax153 looks good to me, but it's still marked as draft. I'm not sure whether it's just an oversight.
>
> @kitlangton?

Ah, sorry for not getting back to this. Yes, I kept it as a draft because I was somewhat dissatisfied with my own implementation. If I had not happened to implement this myself, I don't think I would have expected `decodeResult` to throw. 

I don't know if there's any other sensible implementation (returning an... `Option<Result>` seems equally awful), but it crossed my mind that perhaps this is why this component was missing. 😄 But if it feels reasonable to you, I am more than happy!

gcanti:
@kitlangton yes, it was missing because 

- I didn't need it for development and 
- I didn't like that it throws an error
 
however it already behaves this way in v3, so it makes sense to reintroduce it

PR #1644, opened 2026-03-05
fix: update Service interface to use 'this: void' in 'of' method signatures

patroza:
Aligns it with the implementation.


mikearnaldi:
Why?

patroza:
> Why?

`Effect.map(MyService.of)`

PR #1651, opened 2026-03-05
[RFC] feat(unstable/ai): add client specific resources based on capabilities

lloydrichards:
## Type

- [x] Request For Comment
- [x] Feature

## Objective

Part of the mcp-app [spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#clientserver-capability-negotiation) and extensions [spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2133-extensions.md) is the expansion of the client and server capabilities to be able to negotiate what tools and resources should be available based on each's ability.  I've already updated the `McpSchema` in #1440 but there is currently no way to access the client's capabilities.  In this PR I would like to enable `McpServer` to 
1. access client init context
2. advertise server extensions
3. transform tool/resource/prompt lists based on client capabilities. 

While at the same time not introducing any specific extension into the /unstable/ai module so that extension application is still left to the implementation layer.

### API additions
For client context there nees to be a way to update a new `Service` which holds the client initialization params (separate from `McpServerClient`)
- `McpServerClientContext: ServiceMap.Service`
-  `McpService.getClientInitialize: (clientId: number) => Option.Option<typeof Initialize.payloadSchema.Type>`
-  `McpService.setClientInitialize: (clientId: number, init: typeof Initialize.payloadSchema.Type) => Effect.Effect<void>`
-  `McpService.clearClientInitialize: (clientId: number) => Effect.Effect<void>`

For server extensions there needs to be a way to register the extension similar to a toolkit so that it can be applies during initialization.
- `registerServerExtensions(extensions: Record<string, Schema.Json>)`
- `serverExtensions(extensions: Record<string, Schema.Json>): Layer`

For list transformations there needs to be something that can be registered to transform (filter, map, replace etc) depending on the `McpServerClientContext`
- `registerToolsTransform(transform: ToolsTransformer)`
- `toolsTransform(transform: ToolsTransformer): Layer`
- `registerResourcesTransform(transform: ResourcesTransformer)`
- `resourcesTransform(transform: ResourcesTransformer): Layer`
- `registerPromptsTransform(transform: PromptsTransformer)`
- `promptsTransform(transform: PromptsTransformer): Layer`

### Example
```ts
// Advertise server extensions:
const McpLayer = McpServer.layerHttp({
  name: "Example",
  version: "1.0.0",
  path: "/mcp"
}).pipe(
  Layer.provideMerge(RpcSerialization.layerJsonRpc()),
  Layer.provideMerge(
    McpServer.serverExtensions({
      "io.modelcontextprotocol/ui": { host: true }
    })
  )
)

// Filter tools list based on client capability:
const UiTool = Tool.make("UiTool", {
  description: "ui tool",
  parameters: Schema.Record(Schema.String, Schema.Never),
  success: Schema.Struct({ ok: Schema.Boolean })
}).annotate(Tool.Meta, { ui: { resourceUri: "ui://example/ui" } })
const UiToolkit = Toolkit.make(UiTool)
const ToolsLayer = McpServer.toolkit(UiToolkit).pipe(
  Layer.provide(UiToolkit.toLayer({
    UiTool: () => Effect.succeed({ ok: true })
  }))
)

const TransformerLayer = McpServer.toolsTransform((items, ctx) => {
  const hasUISupport = ctx.pipe(
    Option.flatMap((value) => Option.fromNullishOr(value.capabilities.extensions?.["io.modelcontextprotocol/ui"])),
    Option.map((value) =>
      typeof value == "object" && "mimeTypes" in value &&
      Array.isArray(value.mimeTypes) &&
      value.mimeTypes?.includes("text/html;profile=mcp-app")
    ),
    Option.getOrNull
  )
  return Effect.succeed(hasUISupport ? items : [])
})

const McpLayerLive = Layer.mergeAll(McpLayer, ToolsLayer, TransformerLayer)
```

In this way I'm able to define support and fallbacks for the mcp-app and follow the specs given for the client<>server negotiation without needing to have /unstable/ai know about anything more than the `ClientCapabilities`

## How to Test

I've gone and made a [patch](https://github.com/lloydrichards/edu_mcp-apps/blob/main/patches/effect%404.0.0-beta.25.patch) for my own repo in order to test these changes out already and things seem to work as expected, but still need a client setup with the capabilities to test the filtering (see [issue](https://github.com/MCPJam/inspector/issues/1538))

## Related

- Discord [thread](https://discord.com/channels/795981131316985866/1475976849476812872)

### DIFF
--- a/packages/effect/src/unstable/ai/McpServer.ts
+++ b/packages/effect/src/unstable/ai/McpServer.ts
@@ -132,6 +132,12 @@ export class McpServer extends ServiceMap.Service<McpServer, {
 
   readonly serverExtensions: Record<string, Schema.Json>
   readonly addServerExtensions: (extensions: Record<string, Schema.Json>) => Effect.Effect<void>
+  readonly toolsTransforms: ReadonlyArray<ToolsTransformer>
+  readonly addToolsTransform: (transform: ToolsTransformer) => Effect.Effect<void>
+  readonly resourcesTransforms: ReadonlyArray<ResourcesTransformer>
+  readonly addResourcesTransform: (transform: ResourcesTransformer) => Effect.Effect<void>
+  readonly promptsTransforms: ReadonlyArray<PromptsTransformer>
+  readonly addPromptsTransform: (transform: PromptsTransformer) => Effect.Effect<void>
 }>()("effect/ai/McpServer") {
   /**
    * @since 4.0.0
@@ -164,6 +170,9 @@ export class McpServer extends ServiceMap.Service<McpServer, {
     >()
     const clientInitialize = new Map<number, typeof Initialize.payloadSchema.Type>()
     const serverExtensions: Record<string, Schema.Json> = {}
+    const toolsTransforms: Array<ToolsTransformer> = []
+    const resourcesTransforms: Array<ResourcesTransformer> = []
+    const promptsTransforms: Array<PromptsTransformer> = []
     const notificationsQueue = yield* Queue.make<RpcMessage.Request<any>>()
     const listChangedHandles = new Map<string, any>()
     const notifications = yield* RpcClient.makeNoSerialization(ServerNotificationRpcs, {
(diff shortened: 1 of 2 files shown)


lloydrichards:
Looks like Tim had a better version (#1653) in mind 😄 will be happy to test out his version in the next beta

PR #1658, opened 2026-03-05
Tied stdout and stderr sinks lifecycle to layer's

nikelborm:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Repro made by Valentine in Telegram group https://t.me/fp_ts/10318

```ts
import * as NodeRuntime from "@effect/platform-node/NodeRuntime";
import * as NodeServices from "@effect/platform-node/NodeServices";
import * as Effect from "effect/Effect";
import * as Stdio from "effect/Stdio";
import * as Stream from "effect/Stream";

const main = Effect.gen(function* () {
  const stdio = yield* Stdio.Stdio;

  const cmd = Effect.fnUntraced(function* () {
    yield* Stream.fromIterable(
      "wefwefwefwefwefwefwefwefwefwefwefwefwefwefwefwefwef",
    ).pipe(Stream.run(stdio.stdout));
    console.log("❌");
  });

  yield* cmd();
  yield* cmd();
  yield* cmd();
}).pipe(Effect.provide([NodeServices.layer]));

NodeRuntime.runMain(main);
```

The code above crashes when it attempts to pipe another stream into stdout. The PR fixes that

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/vast-deserts-travel.md
+++ b/.changeset/vast-deserts-travel.md
@@ -2,4 +2,4 @@
 "@effect/platform-node-shared": patch
 ---
 
-Made stdout and stderr sinks from node.js's Stdio be able to take elements from more than one stream, effectively tying them to layer lifecycle. Previously both of them called .end() on underlying process.stdout and process.stdin the moment the first stream finished
+Add `endOnDone` option to Stdio stdout / stderr
--- a/packages/effect/src/Stdio.ts
+++ b/packages/effect/src/Stdio.ts
@@ -27,8 +27,12 @@ export const TypeId: TypeId = "~effect/Stdio"
 export interface Stdio {
   readonly [TypeId]: TypeId
   readonly args: Effect.Effect<ReadonlyArray<string>>
-  readonly stdout: Sink.Sink<void, string | Uint8Array, never, PlatformError>
-  readonly stderr: Sink.Sink<void, string | Uint8Array, never, PlatformError>
+  stdout(options?: {
+    readonly endOnDone?: boolean | undefined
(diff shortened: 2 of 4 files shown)


tim-smart (dismissed):

nikelborm:
I really think the default setting should be reconsidered.

Node.js here is a lucky case, because at least it says _something_ when things go wrong. Although it's a completely useless error from debugging standpoint because it doesn't even have a message, it's something.
image

On the other hand, bun doesn't give any feedback at all, it just silently ignores the second `.end()` call, and doesn't emit `"finish"`, making the second `Effect.callback` just hang forever. Or maybe timeout should be added to `onDone` `Effect.callback`, so that it doesn't hang forever if it was cleaned twice.

I would not expect people to just infer and understand by default, why their `Stream.run` operation just hangs forever. It would involve repeated debugging.

Whatever you decide, I'm open to make a new PR with the new logic.

nikelborm:
Could you please explain the reason, why tying cleanup to the lifecycle of the layer, instead of the lifecycle of the first stream is a bad idea?

nikelborm:
Also as a side-note Evgenii from telegram pointed-out that `onError` is not even used in the body of the function and it's not clear why was it defined there. 
image

tim-smart:
> Could you please explain the reason, why tying cleanup to the lifecycle of the layer, instead of the lifecycle of the first stream is a bad idea?

The platform services need to have no side effects. Also if the layer is used multiple times in a program that would lead to issues.

I think setting endOnDone to false by default is fine: https://github.com/Effect-TS/effect-smol/pull/1671

PR #1659, opened 2026-03-06
EFF-673 Send MCP-Protocol-Version header on HTTP MCP requests after initialize

tim-smart:
## Summary
- add `McpServer.layerClientProtocolHttp` to persist MCP HTTP session state from the initialize response and replay `Mcp-Session-Id` plus negotiated `MCP-Protocol-Version` on later HTTP MCP requests
- keep generic `RpcClient.makeProtocolHttp` unchanged and continue looking up replayed session ids in `McpServer` via `Headers.get(...)`
- cover the HTTP flow with an end-to-end regression test that asserts initialize stays header-free and the second request replays both MCP headers

## Validation
- pnpm lint-fix
- pnpm test packages/effect/test/unstable/ai/McpServer.test.ts
- pnpm check:tsgo
- pnpm docgen

### DIFF
--- a/packages/effect/src/unstable/ai/McpServer.ts
+++ b/packages/effect/src/unstable/ai/McpServer.ts
@@ -5,6 +5,7 @@ import * as Arr from "../../Array.ts"
 import * as Cause from "../../Cause.ts"
 import * as Effect from "../../Effect.ts"
 import * as Exit from "../../Exit.ts"
+import * as Fiber from "../../Fiber.ts"
 import * as Layer from "../../Layer.ts"
 import * as Option from "../../Option.ts"
 import * as Queue from "../../Queue.ts"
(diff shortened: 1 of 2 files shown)


packages/effect/src/unstable/rpc/RpcClient.ts
```diff
@@ -816,6 +816,41 @@ export const makeProtocolHttp = (client: HttpClient.HttpClient): Effect.Effect<
   Protocol.make(Effect.fnUntraced(function*(writeResponse) {
     const serialization = yield* RpcSerialization.RpcSerialization
     const isFramed = serialization.includesFraming
+    let mcpTransportHeaders = Headers.empty
```
tim-smart: RpcClient should not be modified in any way.

PR #1660, opened 2026-03-06
EFF-674 Add HttpServerResponse.toClientResponse

tim-smart:
## Summary
- implement `HttpServerResponse.toClientResponse` as a dedicated `ServerHttpClientResponse` wrapper that reads status, headers, cookies, and request metadata directly from the server response
- decode body accessors without routing through `HttpClientResponse.fromWeb` or `toWeb`, while still honoring `withoutBody` and provided `services` for stream-backed responses
- extend regression coverage for JSON, stream, form-data, and `withoutBody` conversion behavior

## Validation
- pnpm lint-fix
- pnpm test packages/effect/test/unstable/http/HttpServerResponse.test.ts
- pnpm check:tsgo
- pnpm docgen

### DIFF
--- a/packages/effect/src/unstable/http/HttpServerResponse.ts
+++ b/packages/effect/src/unstable/http/HttpServerResponse.ts
@@ -823,14 +823,11 @@ export const toClientResponse = (
   response: HttpServerResponse,
   options?: {
     readonly request?: HttpClientRequest.HttpClientRequest | undefined
-    readonly withoutBody?: boolean | undefined
-    readonly services?: ServiceMap.ServiceMap<never> | undefined
   }
 ): HttpClientResponse.HttpClientResponse =>
   new ServerHttpClientResponse(
     options?.request ?? HttpClientRequest.empty,
-    response,
-    options
+    response
   )
 
 class ServerHttpClientResponse extends Inspectable.Class implements HttpClientResponse.HttpClientResponse {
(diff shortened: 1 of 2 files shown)


packages/effect/src/unstable/http/HttpServerResponse.ts
```diff
@@ -811,6 +813,23 @@ export const toWeb = (
...
+ */
+export const toClientResponse = (
+  response: HttpServerResponse,
+  options?: {
+    readonly request?: HttpClientRequest.HttpClientRequest | undefined
+    readonly withoutBody?: boolean | undefined
+    readonly services?: ServiceMap.ServiceMap<never> | undefined
+  }
+): HttpClientResponse.HttpClientResponse =>
+  HttpClientResponse.fromWeb(
```
tim-smart: Do not use .fromWeb. Instead implement a new wrapper class.

packages/effect/src/unstable/http/HttpServerResponse.ts
```diff
@@ -811,6 +815,184 @@ export const toWeb = (
...
+    return HttpIncomingMessage.inspect(this, {
+      _id: "HttpClientResponse",
+      request: this.request.toJSON(),
+      status: this.status
+    })
+  }
+
+  private cachedResponse?: globalThis.Response
+  private get source(): globalThis.Response {
+    return this.cachedResponse ??= toWeb(this.response, {
```
tim-smart: Do not use `toWeb`. Instead set `source` to the `HttpClientRequest`.

The other method need to use the `HttpClientRequest` for their implementation.

PR #1661, opened 2026-03-06
EFF-675 Add HttpServerRequest.fromClientRequest

tim-smart:
## Summary
- implement `HttpServerRequest.fromClientRequest` with a dedicated `HttpClientRequest`-backed request implementation instead of converting through a web `Request`
- preserve request `source`, relative and absolute URLs, GET request bodies, and multipart / stream body handling while using a web `Response` only for body serialization where needed
- add regression coverage for source identity, request metadata, GET bodies, and encoded form-data / stream bodies

### DIFF
--- a/packages/effect/src/unstable/http/HttpServerRequest.ts
+++ b/packages/effect/src/unstable/http/HttpServerRequest.ts
@@ -302,14 +302,11 @@ export const fromClientRequest = (
   if (request.hash !== undefined) {
     url.hash = request.hash
   }
-  return new ServerRequestImpl(
-    new ClientRequestSource(
-      request,
-      url,
-      isAbsoluteClientUrl(request.url) ? url.toString() : removeHost(url.toString()),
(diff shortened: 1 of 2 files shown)


packages/effect/src/unstable/http/HttpServerRequest.ts
```diff
@@ -283,6 +284,38 @@ export const schemaBodyFormJson = <A, I, RD, RE>(
...
+  const url = new URL(request.url, clientRequestBaseUrl)
+  for (let i = 0; i < request.urlParams.params.length; i++) {
+    const [key, value] = request.urlParams.params[i]
+    url.searchParams.append(key, value)
+  }
+  if (request.hash !== undefined) {
+    url.hash = request.hash
+  }
+  const webRequest = new Request(url, toWebRequestInit(request, options))
+  return new ServerRequestImpl(
```
tim-smart: Create a new wrapper class. Do not convert to a web request first.

PR #1667, opened 2026-03-06
EFF-679 Fix Atom.withReactivity registry initial values

tim-smart:
## Summary
- redirect `Atom.initialValue(...)` for `Atom.withReactivity(...)` wrappers to the underlying source atom so the wrapper still installs its reactivity handler on first read
- add a regression test covering registry-seeded values refreshing after a keyed mutation
- add a patch changeset for the fix

### DIFF
--- a/packages/effect/src/unstable/reactivity/Atom.ts
+++ b/packages/effect/src/unstable/reactivity/Atom.ts
@@ -1471,11 +1471,19 @@ export const transform: {
         )
     )
     if (options?.initialValueTarget) {
-      ;(atom as Mutable<Atom<B>>).initialValueTarget = options.initialValueTarget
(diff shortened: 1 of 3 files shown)


packages/effect/src/unstable/reactivity/Atom.ts
```diff
@@ -311,6 +313,9 @@ const WritableProto = {
  */
 export const isWritable = <R, W>(atom: Atom<R>): atom is Writable<R, W> => WritableTypeId in atom
 
+const getInitialValueTarget = <A>(self: Atom<A>): Atom<A> =>
```
tim-smart: Move this to a property alongside .keepAlive etc.

PR #1709, opened 2026-03-11
Extract PartitionedSemaphore and add module helpers

mikearnaldi:
## Summary
- addressed PR #1709 review feedback by changing module-level withPermit-family helpers so the non-partial overload takes semaphore/latch first, matching Layer-style dual APIs
- preserved pipeline ergonomics via curried forms:
  - `effect.pipe(Semaphore.withPermit(self))`, `effect.pipe(Semaphore.withPermits(self, n))`, `effect.pipe(Semaphore.withPermitsIfAvailable(self, n))`
  - same pattern for `PartitionedSemaphore` and `Latch.whenOpen`
- aligned `TxSemaphore.withPermit` and `TxSemaphore.withPermits` to the same overload order; behavior unchanged
- updated tests to assert the new self-first direct-call forms and existing pipe usage

## Tests
- `pnpm lint-fix`
- `pnpm test packages/effect/test/Semaphore.test.ts`
- `pnpm test packages/effect/test/PartitionedSemaphore.test.ts`
- `pnpm test packages/effect/test/Latch.test.ts`
- `pnpm test packages/effect/test/TxSemaphore.test.ts`
- `pnpm check:tsgo`
- `pnpm docgen`

### DIFF
--- a/packages/effect/src/PartitionedSemaphore.ts
+++ b/packages/effect/src/PartitionedSemaphore.ts
@@ -304,16 +304,16 @@ export const withPermits: {
   ): <A, E, R>(effect: Effect.Effect<A, E, R>) => Effect.Effect<A, E, R>
   <K, A, E, R>(
     self: PartitionedSemaphore<K>,
-    effect: Effect.Effect<A, E, R>,
     key: K,
-    permits: number
+    permits: number,
+    effect: Effect.Effect<A, E, R>
   ): Effect.Effect<A, E, R>
 } = ((...args: Array<any>) => {
   if (args.length === 3) {
     const [self, key, permits] = args
     return (effect: Effect.Effect<any, any, any>) => self.withPermits(key, permits)(effect)
   }
-  const [self, effect, key, permits] = args
+  const [self, key, permits, effect] = args
   return self.withPermits(key, permits)(effect)
 }) as any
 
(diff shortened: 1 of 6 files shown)


packages/effect/src/PartitionedSemaphore.ts
```diff
@@ -0,0 +1,359 @@
...
+/**
+ * Runs an effect only if the permits are immediately available.
+ *
+ * @since 4.0.0
+ * @category combinators
+ */
+export const withPermitsIfAvailable: {
+  <A, E, R>(
+    permits: number,
+    effect: Effect.Effect<A, E, R>
```
tim-smart: I think this should be effect last, so you can use it in a pipe.

packages/effect/src/Latch.ts
```diff
@@ -110,3 +111,66 @@ export const makeUnsafe: (open?: boolean | undefined) => Latch = internal.makeLa
...
+export const closeUnsafe = (self: Latch): boolean => self.closeUnsafe()
+
+/**
+ * Runs the given effect only when the latch is open.
+ *
+ * @category combinators
+ * @since 4.0.0
+ */
+export const whenOpen: {
+  <A, E, R>(effect: Effect.Effect<A, E, R>): (self: Latch) => Effect.Effect<A, E, R>
```
tim-smart: The effect and latch should be swapped to use in an effect pipe.

tim-smart (dismissed):

PR #1720, opened 2026-03-12
fix: json RpcSerialization decode double-wraps array responses

### LINKED ISSUE #1719: JSON RpcSerialization decode double-wraps array responses, silently discarding all RPC results
## Summary

When using `RpcSerialization.layerJson` with `RpcClient.layerProtocolHttp`, all RPC responses are silently discarded. The HTTP request succeeds (200 OK with valid response data), but the client-side Effect never resolves — it hangs until timeout.

## Environment

- `effect@4.0.0-beta.31`
- Client: React app using `AtomRpc.Service` + `ManagedRuntime` (also reproduced with `Effect.provide` + `Effect.runPromise`)
- Server: Cloudflare Worker using `RpcServer.layerProtocolHttp` + `RpcSerialization.layerJson`

## Root Cause

The `json` serializer's `decode` method in `RpcSerialization.ts` unconditionally wraps the `JSON.parse()` output in an array:

```typescript
decode: (bytes) => [JSON.parse(typeof bytes === "string" ? bytes : decoder.decode(bytes))]
```

The server's HTTP protocol (`makeProtocolWithHttpEffect` in `RpcServer.ts`) collects all response messages from a queue and encodes them as a JSON array:

```typescript
const responses = yield* Queue.collect(queue)
return HttpServerResponse.text(parser.encode(responses), { contentType: ... })
```

So the wire format is: `[{"_tag":"Exit","requestId":"0","exit":{"_tag":"Success","value":{...}}}]`

When the client decodes this:
1. `JSON.parse(text)` → `[{_tag: "Exit", requestId: "0", ...}]` (correct array of messages)
2. `decode` wraps it → `[[{_tag: "Exit", requestId: "0", ...}]]` (double-nested)

The client's `makeProtocolHttp` non-framed path iterates the decoded result expecting each element to be a message object:

```typescript
const u = parser.decode(text)
// u = [[msg1, msg2, ...]]
// u[0] = [msg1, msg2, ...] — an array, not a message object!
writeResponse(u[i++])  // handler receives an array
```

The handler's `switch(message._tag)` sees `undefined` (arrays don't have `_tag`), falls to `default: return Effect.void`, and the response is **silently discarded**. The waiting `Effect.callback` is never resumed.

## How We Discovered This

We instrumented the compiled `RpcClient.js` with `console.log` at key points in the response processing chain:

```
[RPC-DEBUG] HTTP response text: [{"_tag":"Exit","requestId":"0","exit":{"_tag":"Success","value":{"pong":"v1-test","ts":...}}}]
[RPC-DEBUG] decoded: [[{"_tag":"Exit","requestId":"0",...}]] isArray: true
[RPC-DEBUG] calling writeResponse with: [{"_tag":"Exit","requestId":"0",...}]   ← array, not object
[RPC-DEBUG] run handler received:  requestId:  entries keys: Array(2)           ← _tag is undefined
```

## Our Workaround

We created a custom JSON serialization layer that returns the parsed array directly instead of wrapping:

```typescript
const layerJsonFixed = Layer.succeed(
  RpcSerialization.RpcSerialization,
  RpcSerialization.RpcSerialization.of({
    contentType: "application/json",
    includesFraming: false,
    makeUnsafe: () => {
      const decoder = new TextDecoder()
      return {
        decode: (bytes: string | Uint8Array) => {
          const parsed = JSON.parse(typeof bytes === "string" ? bytes : decoder.decode(bytes))
          return Array.isArray(parsed) ? parsed : [parsed]
        },
        encode: (response: unknown) => JSON.stringify(response),
      }
    },
  }),
)
```

After applying this, both `ManagedRuntime` and `Effect.provide` patterns work correctly — responses flow through the full chain and `entry.resume(exit)` is called as expected.

## Note

We're not sure if this is a misuse of the RPC server on our end (perhaps the server shouldn't be encoding responses as a JSON array?), or if it is a bug in the JSON serialization layer. Either way, the `json.decode` / `json.encode` roundtrip is broken — `encode([msg1, msg2])` produces `"[msg1,msg2]"`, but `decode("[msg1,msg2]")` produces `[[msg1, msg2]]` instead of `[msg1, msg2]`.

We have a PR with a proposed fix and tests: https://github.com/matthew-sessions/effect-smol/tree/fix/json-serialization-double-wrap

matthew-sessions:
## Summary

Fixes #1719

- `RpcSerialization.json.decode` unconditionally wraps `JSON.parse()` output in `[...]`, causing double-nesting when the server sends a JSON array of RPC response messages via the HTTP protocol
- The `makeProtocolHttp` non-framed path iterates the decoded result expecting message objects, but receives arrays — `message._tag` is `undefined`, responses are silently discarded, and the RPC call hangs until timeout
- Fix: return the parsed value directly when it is already an array, only wrapping non-array values

## What changed

**`packages/effect/src/unstable/rpc/RpcSerialization.ts`**

```typescript
// Before:
decode: (bytes) => [JSON.parse(typeof bytes === "string" ? bytes : decoder.decode(bytes))]

// After:
decode: (bytes) => {
  const parsed = JSON.parse(typeof bytes === "string" ? bytes : decoder.decode(bytes))
  return Array.isArray(parsed) ? parsed : [parsed]
}
```

## Test plan

- [x] Added test: `json decode returns flat array when server sends array of messages` — verifies decode returns `[msg1, msg2]` not `[[msg1, msg2]]`
- [x] Added test: `json decode wraps single non-array value in array` — verifies single objects are still wrapped
- [x] Added test: `json encode/decode roundtrip preserves messages` — verifies `decode(encode(messages))` equals `messages`
- [x] Existing `Rpc.test.ts` passes (3/3)
- [x] Existing `AtomRpc.test.ts` passes (1/1)
- [x] Existing `RpcSerialization.test.ts` jsonRpc tests pass (2/2)
- [x] Validated end-to-end in a real app (React + Cloudflare Worker) — RPC responses now flow through correctly

### DIFF
--- a/packages/effect/src/unstable/rpc/RpcSerialization.ts
+++ b/packages/effect/src/unstable/rpc/RpcSerialization.ts
@@ -37,7 +37,10 @@ export const json: RpcSerialization["Service"] = RpcSerialization.of({
   makeUnsafe: () => {
     const decoder = new TextDecoder()
     return {
-      decode: (bytes) => [JSON.parse(typeof bytes === "string" ? bytes : decoder.decode(bytes))],
+      decode: (bytes) => {
+        const parsed = JSON.parse(typeof bytes === "string" ? bytes : decoder.decode(bytes))
+        return Array.isArray(parsed) ? parsed : [parsed]
+      },
       encode: (response) => JSON.stringify(response)
     }
   }
--- a/packages/effect/test/rpc/RpcSerialization.test.ts
+++ b/packages/effect/test/rpc/RpcSerialization.test.ts
@@ -11,6 +11,47 @@ const responseExitSuccess = (requestId: string, value: unknown) => ({
 })
 
 describe("RpcSerialization", () => {
+  it("json decode returns flat array when server sends array of messages", () => {
+    const parser = RpcSerialization.json.makeUnsafe()
+
+    const msg1 = responseExitSuccess("0", { name: "Alice" })
+    const msg2 = responseExitSuccess("1", { name: "Bob" })
+    const serverResponse = JSON.stringify([msg1, msg2])
+
+    const decoded = parser.decode(serverResponse)
+
+    // decoded should be [msg1, msg2], NOT [[msg1, msg2]]
+    assert.strictEqual(decoded.length, 2)
+    assert.deepStrictEqual(decoded[0], msg1)
+    assert.deepStrictEqual(decoded[1], msg2)
+  })
+
+  it("json decode wraps single non-array value in array", () => {
+    const parser = RpcSerialization.json.makeUnsafe()
+
+    const msg = responseExitSuccess("0", { name: "Alice" })
+    const serverResponse = JSON.stringify(msg)
+
+    const decoded = parser.decode(serverResponse)
+
+    assert.strictEqual(decoded.length, 1)
+    assert.deepStrictEqual(decoded[0], msg)
+  })
+
+  it("json encode/decode roundtrip preserves messages", () => {
+    const parser = RpcSerialization.json.makeUnsafe()
+
+    const messages = [
+      responseExitSuccess("0", { name: "Alice" }),
+      responseExitSuccess("1", { name: "Bob" })
+    ]
+
+    const encoded = parser.encode(messages)
+    const decoded = parser.decode(encoded as string)
+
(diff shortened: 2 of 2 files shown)


tim-smart:
Ah sorry didn't see this and have already fixed :)

PR #1723, opened 2026-03-12
Schema.Class plain objects rejected by HttpApi encode at runtime

kitlangton:
## Problem

When using `Schema.Class` as a payload or success type in HttpApi, plain objects that **typecheck** fail at **runtime**.

```ts
class MyBody extends Schema.Class<MyBody>("MyBody")({
  name: Schema.String,
  age: Schema.Number
}) {}

const Api = HttpApi.make("Api").add(
  HttpApiGroup.make("users").add(
    HttpApiEndpoint.post("createUser", "/users", {
      payload: MyBody,
      success: MyResponse
    })
  )
)

const client = yield* HttpApiClient.make(Api)

// Typechecks — MyBody is structurally { name: string, age: number }
// Runtime: SchemaError: Expected MyBody, got {"name":"Kit","age":30}
yield* client.users.createUser({ payload: { name: "Kit", age: 30 } })
```

Same issue on the server — a handler returning `Effect.succeed({ id: 1, ... })` for a `Schema.Class` success type typechecks but fails during response encoding.

The root cause: `Schema.Class` creates a Declaration AST whose guard requires `instanceof`. TypeScript's structural typing makes plain objects assignable to the class type (no nominal brand), so the types accept them while the runtime rejects them.

## Proposed fix

This PR adds `schemaForEncoding` in `HttpApiSchema` — when building encode-only schema pipelines (payload → HttpBody, success → HttpServerResponse), class Declaration guards are replaced with a lenient `isObject` check. All annotations, checks, encoding config, and context are preserved.

**This does not change `Schema.Class` behavior outside of HttpApi.** `Schema.encodeUnknownSync(MyBody)(plainObj)` still fails — only the HttpApi encode pipeline is made lenient.

Not sure if this is the right implementation — the type/runtime mismatch in `Schema.Class` is arguably a deeper issue (maybe classes should be nominally branded, or maybe `encodeUnknown` should be lenient for classes globally). This is more of a reproducer and one possible fix to consider.

## Test plan

- [ ] Client payload: plain object accepted for Schema.Class endpoint
- [ ] Client payload: class instance still works (no regression)
- [ ] Server encode: handler can return plain object for Schema.Class success
- [ ] Server decode: plain object payload decoded into proper class instance
- [ ] Client decode: response is a proper class instance (`instanceof` check)
- [ ] Existing Schema.Class tests unchanged
- [ ] Existing HttpApi tests pass (including asNoContent, error encoding, etc.)

### DIFF
--- a/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
@@ -716,7 +716,7 @@ function toResponseSchema(getStatus: (ast: AST.AST) => number) {
       return cached as any
     }
     const responseSchema = $HttpServerResponse.pipe(
-      Schema.decodeTo(schema, getResponseTransformation(getStatus, schema))
+      Schema.decodeTo(HttpApiSchema.schemaForEncoding(schema), getResponseTransformation(getStatus, schema))
     )
     cache.set(responseSchema.ast, responseSchema)
     return responseSchema
--- a/packages/effect/src/unstable/httpapi/HttpApiClient.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiClient.ts
@@ -651,7 +651,7 @@ function getEncodePayloadSchemaFromBody(
   }
   const encoding = HttpApiSchema.getPayloadEncoding(ast)
   const out = $HttpBody.pipe(Schema.decodeTo(
-    schema,
+    HttpApiSchema.schemaForEncoding(schema),
     Transformation.transformOrFail({
       decode(httpBody) {
         return Effect.fail(new Issue.Forbidden(Option.some(httpBody), { message: "Encode only schema" }))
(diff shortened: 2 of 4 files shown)


mikearnaldi:
I don't think it's a good idea, classes are not plain objects and making special cases will inevitably lead to confusion, the advise is to brand the classes like:

```ts
interface MyBody {
  readonly _brand: unique symbol
}
class MyBody extends Schema.Class<MyBody>("MyBody")({
  name: Schema.String,
  age: Schema.Number
}) {}
```

cc @gcanti this keeps coming up, I am positive we should add an easy way to brand classes and similar constructors like:

```ts
class MyBody extends Schema.Class<MyBody, { readonly _brand: unique symbol }>("MyBody")({
  name: Schema.String,
  age: Schema.Number
}) {}
```

gcanti:
@mikearnaldi I already answered in the linked issue. The current way to brand classes is exactly as you described. I think we can close both this PR and the issue.

PR #1734, opened 2026-03-13
fix: middleware union errors return 500 instead of declared httpApiStatus

kitlangton:
## Summary

- When a middleware declares `error: Schema.Union([ErrorA, ErrorB])`, each error's `httpApiStatus` annotation was lost — the server always returned HTTP 500
- `getErrorSchemas` now flattens union middleware errors into individual members so `toResponseErrorSchema` correctly resolves each member's status code

Not sure if this is the right approach — consider this a proposal. Happy to adjust if there's a preferred pattern.

## Test plan

- [x] Added tests for middleware union errors (Unauthorized → 401, Forbidden → 403)
- [x] All 5674 existing tests pass
- [x] Types clean (`pnpm check:tsgo`)

### DIFF
--- a/.changeset/fix-middleware-union-error-status.md
+++ b/.changeset/fix-middleware-union-error-status.md
@@ -1,0 +1,9 @@
+---
+"effect": patch
+---
+
+HttpApi: fix middleware union errors returning 500 instead of declared httpApiStatus.
+
+When a middleware declares `error: Schema.Union([ErrorA, ErrorB])`, each member's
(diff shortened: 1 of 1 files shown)


gcanti:
use `error: [ErrorA, ErrorB]`, see https://github.com/Effect-TS/effect-smol/blob/6d9393a0770a18722d23340e77f15455de341245/packages/effect/HTTPAPI.md#adding-custom-error-responses

PR #1799, opened 2026-03-19
fix(RcMap): invalidate should close scope regardless of refCount

kitlangton:
Howdy all! I ran into some unexpected semantics when using `LayerMap` in the opencode repo. As it says at the bottom, I'm not sure if this is intended or not, but here's a reproducer and potential fix if it isn't indeed unintended.

# Human Approved Robot Summary 🤖🤝👵

## Problem

`RcMap.invalidate` does not close the scope or interrupt consumers when `refCount > 0`. It removes the entry from the map and silently returns:

```ts
MutableHashMap.remove(self.state.map, key)
if (entry.refCount > 0) return              // ← silently abandons
if (entry.fiber) yield* Fiber.interrupt(entry.fiber)
yield* Scope.close(entry.scope, Exit.void)  // never reached
```

This is inconsistent with `ScopedCache.invalidate` which always closes the scope:

```ts
MutableHashMap.remove(self.state.map, key)
return Scope.close(oentry.value.scope, effect.exitVoid)  // no refCount check
```

Every other Effect lifecycle abstraction (FiberMap, FiberSet, ScopedCache) interrupts and tears down on invalidation/close. RcMap is the only one that silently abandons active consumers without interrupting them.

## Impact

This makes `LayerMap` (which wraps `RcMap`) unable to cleanly tear down services with long-running consumers. When a layer has a PubSub that streams subscribe to, `invalidate` can't trigger the layer's finalizers to shut down the PubSub, so streams never end and `ensuring` callbacks never fire.

## Fix

Remove the `if (entry.refCount > 0) return` line. `invalidate` now always closes the scope and runs finalizers, matching `ScopedCache` semantics.

## Test

Added: "invalidate with active consumer closes scope" — acquires a resource, holds the scope open, invalidates, and asserts the finalizer runs. This test fails before the fix and passes after.

All 8 existing RcMap tests continue to pass (they only test invalidate with refCount=0).

## Question for maintainers

Is the silent abandon behavior (bail on refCount > 0) intentional? If so, should there be a separate `forceInvalidate` API? The current behavior means "invalidate" doesn't actually invalidate when consumers are active — it just removes the entry from the map and hopes consumers eventually release on their own.

### DIFF
--- a/.changeset/fix-rcmap-invalidate.md
+++ b/.changeset/fix-rcmap-invalidate.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+fix(RcMap): invalidate now closes the scope regardless of refCount, matching ScopedCache.invalidate semantics


tim-smart:
This is the intended behaviour. RcMap should only closed the resource when the ref count reaches 0.

You can use ScopedCache if you don't need ref counting.

PR #1801, opened 2026-03-19
Use computation end time for cachedInvalidateWithTTL

mattiamanzati:
Use computation end time for cachedInvalidateWithTTL


tim-smart:
Already fixed, sorry didn't see this earlier :)

PR #1804, opened 2026-03-19
Stream.fromPubSub propagates interrupts on PubSub shutdown

kitlangton:
Stacked on #1800. The diff against main includes #1800's commits. The only new commit here is the top one (`Stream.fromPubSub propagates interrupts on PubSub shutdown`).

## Problem

`Stream.fromQueue` propagates interrupts when the Queue shuts down, but `Stream.fromPubSub` completed gracefully with `[]`.

`Channel.fromSubscriptionArray` wrapped the pull with:
```ts
Effect.onInterrupt(PubSub.takeAll(subscription), () => Cause.done())
```

This converted ALL interrupts into a `Done` signal (graceful stream end). It exists because subscription scope closure also interrupts deferreds, and the stream should end cleanly when its scope closes. But PubSub shutdown also interrupts deferreds, so both cases were treated the same.

## How onInterrupt -> Cause.done() works

`onInterrupt` is built on `onExit`. When the inner effect is interrupted, the handler runs. If the handler **succeeds** (`Effect.void`), the original interrupt exit propagates via `flatMap(eff, (_) => exit)`. If the handler **fails** (like `Cause.done()` which is `Effect.fail(Done)`), the failure replaces the original interrupt. The channel treats `Done` as end-of-stream.

## Fix

Make the handler conditional using `pubsubShutdownFlag` from #1800:

```ts
// Before
() => Cause.done()

// After
() => subscription.pubsubShutdownFlag.current ? Effect.void : Cause.done()
```

- PubSub shutdown (`pubsubShutdownFlag` is true) -> handler returns `Effect.void` -> original interrupt propagates -> stream exits with interrupts
- Subscription scope closure (`pubsubShutdownFlag` is false) -> handler returns `Cause.done()` -> interrupt replaced with Done -> stream ends gracefully

Applied to both `fromSubscription` and `fromSubscriptionArray`.

## Test plan

- [x] `Stream.fromPubSub interrupts on shutdown` test passes
- [x] SubscriptionRef test updated to expect interrupt on shutdown
- [x] All PubSub, Queue, SubscriptionRef, Stream tests pass (334 total)
- [x] Types check clean

### DIFF
--- a/.changeset/pubsub-stream-interrupt.md
+++ b/.changeset/pubsub-stream-interrupt.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Stream.fromPubSub now propagates interrupts on PubSub shutdown, matching Stream.fromQueue behavior.


tim-smart:
The Stream ending cleanly is the intended behaviour.

PR #1818, opened 2026-03-20
docs(sql): add JSDoc example for `Statement.CurrentTransformer`

krist7599555:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description                                                 

I am so new in effect world so I not sure that this the right pattern for using `Statement.CurrentTransformer`.

```ts
import { Statement } from "effect/unstable/sql";

const removePasswordTransformer: Statement.Transformer = (self, _sql, _fiber, _span) => {
  (self as any).transformRows = (rows: readonly any[]) => {
    return rows.map((row) => {
      delete row["PASSWORD"]
      return row
    })
  }
  return Effect.succeed(self)
}

const RemovePasswordLayer = Layer.succeed(Statement.CurrentTransformer, removePasswordTransformer)
const AppLayer = Layer.mergeAll(RemovePasswordLayer, DatabaseLayer)
```

If this is not best practice, could you please help suggest me the right one.

> My real use case is that i convert column name in dot notation (eg `project.id` `project.name`) to `{ project: { id, name } }` and it will apply to all rows so it suit for `transformRows`

### DIFF
--- a/packages/effect/src/unstable/sql/Statement.ts
+++ b/packages/effect/src/unstable/sql/Statement.ts
@@ -73,8 +73,36 @@ export type Transformer = (
 ) => Effect.Effect<Statement<unknown>>
 
 /**
- * @category transformer
+ * A service reference holding an optional `Transformer` that is applied to every
+ * SQL statement before it executes. Set it via `Layer.succeed` to post-process
+ * result rows globally without modifying individual queries.
+ *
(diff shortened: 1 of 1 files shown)


IMax153:
Hi @krist7599555,

PR #1825, opened 2026-03-21
fix: flush devtools queued messages on teardown

skoshx:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently `DevToolsClient` Layer simply kills the Queue when being torn down, but it doesn't wait for all items to be flushed through the `Socket` connection, this means that Span end events are discarded, leading to spans being left in a state where they never end, minimal repro below.

This PR fixes it by awaiting the request queue processing fiber in the finalizer.

```ts
// scratchpad/repro.ts
import { NodeRuntime, NodeSocketServer } from "@effect/platform-node";
import { Effect, Stream } from "effect";
import { DevTools, DevToolsServer } from "effect/unstable/devtools";

const received: Array<{ _tag: string; name?: string; status?: string }> = [];

// Start a DevTools server that records all incoming events
const ServerLive = DevToolsServer.run((client) =>
  Stream.fromQueue(client.queue).pipe(
    Stream.runForEach((request) =>
      Effect.sync(() => {
        if (request._tag === "Span") {
          received.push({
            _tag: "Span",
            name: request.name,
            status: request.status._tag,
          });
        } else if (request._tag === "SpanEvent") {
          received.push({ _tag: "SpanEvent", name: request.name });
        }
      }),
    ),
  ),
).pipe(
  Effect.provide(NodeSocketServer.layerWebSocket({ port: 34437 })),
  Effect.forkScoped,
);

// A simple program with a span
  yield* Effect.log("working...");
  yield* Effect.sleep("200 millis");
  yield* Effect.log("done");
}).pipe(Effect.withSpan("my-span"));

// Run both: server + client program
Effect.gen(function* () {
  yield* ServerLive;
  yield* Effect.sleep("500 millis"); // let server start

  yield* program.pipe(Effect.provide(DevTools.layer()));

  yield* Effect.sleep("500 millis"); // give time for final events to flush

  console.log("\n--- Events received by server ---");
  for (const e of received) {
    console.log(JSON.stringify(e));
  }

  const endedSpan = received.find(
    (e) => e._tag === "Span" && e.status === "Ended",
  );
  if (endedSpan) {
    console.log("\n✅ Span Ended event was received (bug is fixed)");
  } else {
    console.log("\n❌ BUG: Span Ended event was NEVER received");
    console.log(
      "   The DevToolsClient finalizer runs after the forkScoped stream",
    );
    console.log(
      "   consumer is already interrupted, so final events are lost.",
    );
  }
}).pipe(Effect.scoped, NodeRuntime.runMain);
```

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/unstable/devtools/DevToolsClient.ts
+++ b/packages/effect/src/unstable/devtools/DevToolsClient.ts
@@ -65,7 +65,7 @@ const makeEffect = Effect.gen(function*() {
     ),
     Stream.onFirst(() => Deferred.completeWith(connected, Effect.void)),
     Stream.runForEach(handleResponse),
-    Effect.forkChild
+    Effect.forkDetach
   )
 
   yield* Effect.addFinalizer(() =>
@@ -73,8 +73,7 @@ const makeEffect = Effect.gen(function*() {
       Effect.andThen(
         Effect.flatMap(Effect.fiberId, (id) => Queue.failCause(requests, Cause.interrupt(id)))
       ),
-      Effect.andThen(Fiber.await(fiber)),
-      Effect.asVoid
+      Effect.andThen(Fiber.await(fiber))
     )
   )
 
--- a/packages/effect/src/unstable/devtools/DevToolsSchema.ts
+++ b/packages/effect/src/unstable/devtools/DevToolsSchema.ts
@@ -1,8 +1,11 @@
 /**
  * @since 4.0.0
  */
+import * as Exit from "../../Exit.ts"
+import { identity } from "../../Function.ts"
 import type * as Option from "../../Option.ts"
 import * as Schema from "../../Schema.ts"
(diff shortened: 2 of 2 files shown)


packages/effect/src/unstable/devtools/DevToolsClient.ts
```diff
@@ -57,12 +65,16 @@ const makeEffect = Effect.gen(function*() {
     ),
     Stream.onFirst(() => Deferred.completeWith(connected, Effect.void)),
     Stream.runForEach(handleResponse),
-    Effect.forkScoped
+    Effect.forkChild
```
tim-smart: This should be forkDetach, so it doesn't get killed if the main fiber finishes for some reason.

PR #1835, opened 2026-03-22
EFF-750 Use pg_notify instead of NOTIFY in PgClient

tim-smart:
## Summary
- switch `PgClient.notify` to execute `SELECT pg_notify($1, $2)` with parameterized channel and payload
- add an integration test that verifies `sql.notify` delivers payloads to an active `sql.listen` stream
- add a changeset for `@effect/sql-pg`

## Validation
- pnpm lint-fix
- pnpm test packages/sql/pg/test/Client.test.ts
- pnpm check:tsgo
- pnpm docgen


packages/sql/pg/test/Client.test.ts
tim-smart: This should use the built in notify api

PR #1858, opened 2026-03-24
fix: enable Tool.EmptyParams with OpenAiLanguageModel

lloydrichards:
## Type

- [x] Bug Fix

## Description

Added a test first to make sure the pattern is correct in `OpenAiStructuredOutput`.  I struggled a bit to find the exact reference in the OpenAI spec but I think it has to do with [strict mode](https://developers.openai.com/api/docs/guides/function-calling#strict-mode) and the `All fields in properties must be marked as required.` which I take to mean that properties is required even if empty.

This should fulfill the error:

```
System error: effect/ai/AiError/AiError: OpenAiClient.createResponseStream: Invalid request. Invalid schema for function 'getCurrentTime': In context=(), object schema missing properties. (POST https://api.openai.com/v1/responses) [code: invalid_function_parameters]  
```

## Related

- Closes #1857

PR #1859, opened 2026-03-24
fix: add dynamic tooling for OpenAILanguageModel

lloydrichards:
## Type

- [x] Bug Fix

## Description

I noticed a while back with #1775 that this only added the support for dynamic tooling for Anthropic and not for OpenAI and OpenAI-compat. While i had the repo open with #1858, I figured I would mirror over the tests and add the same implementation strategy as was in 954e6d655cd32d329b0bfeb872bb654f88b48a13

## Related

- Related Issue #1751

### DIFF
--- a/.changeset/petite-months-allow.md
+++ b/.changeset/petite-months-allow.md
@@ -1,0 +1,6 @@
+---
+"@effect/ai-openai": patch
+"@effect/ai-openai-compat": patch
+---
+
+add dynamic tooling for openai and openai-compact language models


IMax153:
@lloydrichards this just needs a changeset to be able to merge

lloydrichards:
> @lloydrichards this just needs a changeset to be able to merge

Is that still with `pnpm changeset-version` or through the comment from changeset in this PR?

IMax153:
> Is that still with `pnpm changeset-version` or through the comment from changeset in this PR?

You can just run `pnpm changeset` and then make a patch changeset with a short description and add it to the PR.

PR #1862, opened 2026-03-24
fix(schema): preserve jump-to-definition and typedoc for fields in Schema.Struct

sam-goodwin:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Schema.Struct and Schema.Class do not preserve symbol references for fields. The language server can't jump to the definition of a property or see its typedoc:
```ts
class Foo extends Schema.Class<Foo>("Foo")({
  /** Some typedoc */
  someProp: S.String
}) {}

new Foo().someProp; // <- connection lost, no typedoc, not jump-to-def
Foo.fields.someProp; // <- connect preserved
```

### DIFF
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -2288,10 +2288,10 @@ export declare namespace Struct {
     O extends keyof F = TypeOptionalKeys<F>,
     M extends keyof F = TypeMutableKeys<F>
   > =
-    & { readonly [K in Exclude<keyof F, M | O>]: F[K]["Type"] }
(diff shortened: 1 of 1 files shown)


tim-smart:
If you could add a `pnpm changeset` :)

sam-goodwin:
> If you could add a `pnpm changeset` :)

Do I select all packages (since you pin everything) or just effect?

tim-smart:
Just effect at patch level. Also it looks like CI is failing

gcanti:
Resolved in https://github.com/Effect-TS/effect-smol/pull/2158

PR #1875, opened 2026-03-25
Fix AI `Toolkit` type inference

### LINKED ISSUE #1874: Generic Toolkit can't be passed as ToolkitOption

v4.0.0-beta.39

This technically comes from Discord, but there is an issue with the issuify'er

The questions is about creating a agentic loop that can accept a generic Toolkit as a params to pass to Chat.streamText.  This was type-happy in v3 but now in v4 there are some complaints.

[AgenticLoop v3](https://github.com/lloydrichards/base_bevr-stack/blob/main/packages/ai/src/workflow/AgenticLoop.ts) <- Works and typescript is happy 
[AgenticLoop v4](https://github.com/lloydrichards/base_bevr-stack/blob/chore/v4-migrate/packages/ai/src/workflow/AgenticLoop.ts) <- Works, but returns Effect.Effect<string, AiError.AiError, unknown>

Max, Mattia and Michael tried to debug this in the Office Hour 22 ([timestamp](https://www.youtube.com/live/tOsOpvQjglM?si=W9ZR-PMwEYrrTUl9&t=2506)) and the conclusion seems to be that its a type bug with `ToolkitOption<Tool>` and the union between `Option & LanguageModel.GenerateTextOptions<LanguageModel.ExtractTools<Options>>`.

the minified example was:
```ts
const loop = <Tools extends Record<string, Tool.Any>>({
  chat, toolkit,
}: {
  chat: Chat.Service;
  toolkit: Toolkit.WithHandler<Tools>;
}) =>
  Effect.gen(function* () {

    const _ = yield* chat
      .streamText({
        prompt: [],
        toolkit,
      })

    return Stream.runDrain(_)
  });
```

I would expect this to not throw a type error.

### What do you see instead?

Instead we see that `WithHandler<Tools> is not assignable to type ToolkitOption<Tools, never, any>`

### Additional information

- [Discord](https://discord.com/channels/795981131316985866/1485695176566640670)

IMax153:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #1874
- Closes #1874

### DIFF
--- a/packages/effect/src/unstable/ai/LanguageModel.ts
+++ b/packages/effect/src/unstable/ai/LanguageModel.ts
@@ -121,6 +121,7 @@ export interface Service {
    * Generate text using the language model.
    */
   readonly generateText: {
+    // No toolkit: force `{}` instead of falling back to `Record<string, Tool.Any>`.
     <Options extends NoExcessProperties<GenerateTextOptionsWithoutToolkit, Options>>(
       options: Options & GenerateTextOptionsWithoutToolkit
     ): Effect.Effect<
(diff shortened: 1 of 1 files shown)


lloydrichards:
added an issue for you #1874

PR #1896, opened 2026-03-27
chore: update TSTyche to v7

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR updates TSTyche to v7. Release notes: https://tstyche.org/releases/tstyche-7

Mostly configuration refinements and clean up.

To draw your attention, a new matcher `.toBeInstantiableWith()` to test generics got added. Covers some corner cases. At this moment not useful is `effect`, but who knows (;

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "rulesync": "^6.2.0",
     "terser": "^5.46.0",
     "ts-patch": "^3.3.0",
-    "tstyche": "^6.2.0",
+    "tstyche": "7.0.0-rc.0",
     "typescript": "^5.9.3",
     "vite": "^7.3.1",
     "vite-tsconfig-paths": "^6.0.5",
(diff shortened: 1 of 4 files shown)


mrazauskas:
@gcanti It’s worth adding a note that TypeScript 7 is not yet supported. TSTyche needs programmatic APIs, but those are not yet implemented. The TypeScript team is planning to work on the APIs only after TypeScript 7 is shipped.

- I made sure `tstyche --target '>=5.8‘` only includes `6.x`, but not `7.x`
- Similarly, running `tstyche` does fallback to version 6 even when only version 7 is installed
- If I get it right, the TypeScript team plans to keep the behaviour between 6 and 7 as similar as possible

Let’s see. This is a challenging transition, but I also find it exciting. Please ping me if something gets broken. (Testing against nightly releases will not work, but that is acceptable.)

gcanti:
@mrazauskas ok, thank you!

PR #1904, opened 2026-03-28
Refactor id checks for null and empty string

juliusmarminge:
From Codex:

> Yes. There is an upstream/internal parser bug here.
> 
> Two separate things are happening:
> 
> session/update without an id becoming _tag: "Request", id: "" is expected in Effect’s internal RPC representation. That part is not the bug. The transport uses empty-string id as its internal “notification” sentinel, and our code already depends on that in protocol.ts.
> 
> The real bug is that the Effect serializer treats 0 as falsy. In RpcSerialization.ts, decodeJsonRpcMessage does this:
> 
> ```
> if (!decoded.id && decoded.method.startsWith("@effect/rpc/")) { ... }
> 
> id: decoded.id ? String(decoded.id) : ""
> ```
> 
> and encode does this:
> 
> ```
> id: response.id && Number(response.id)
> id: response.requestId ? Number(response.requestId) : undefined
> ```
> 
> Those are wrong for valid JSON-RPC id 0.
> 
> So:
> 
> notification with no id -> id: "" is intentional
> request with id: 0 -> also id: "" is the bug

## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/fix-rpc-json-id-edges.md
+++ b/.changeset/fix-rpc-json-id-edges.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix JSON-RPC serialization for `id` values that are falsey but valid, including `0` and `""`, while still mapping `null` to Effect's internal notification sentinel.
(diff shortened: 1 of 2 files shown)


IMax153:
@juliusmarminge thanks for the PR! Fixed a few issues from CI and added some additional regression tests.

Also pinged @tim-smart for a final review.

PR #1912, opened 2026-03-28
fix(openapi-generator): detect binary and JSON-compatible content types in httpclient format

### LINKED ISSUE #1911: `@effect/openapi-generator`: `httpclient` format only detects `application/octet-stream` for binary and `application/json` for JSON schemas
## What version of Effect is running?

`@effect/openapi-generator@4.0.0-beta.42` / `effect@4.0.0-beta.42`

## What steps can reproduce the bug?

Given an OpenAPI 3.1 spec with:
1. A binary response using a content type other than `application/octet-stream` (e.g., `application/zip`)
2. An error response using `application/problem+json` (RFC 9457)

```json
{
  "/archive": {
    "get": {
      "operationId": "downloadArchive",
      "parameters": [],
      "responses": {
        "200": {
          "description": "Success",
          "content": {
            "application/zip": {
              "schema": { "type": "string", "format": "binary" }
            }
          }
        },
        "404": {
          "description": "Not found",
          "content": {
            "application/problem+json": {
              "schema": {
                "type": "object",
                "properties": {
                  "type": { "type": "string" },
                  "title": { "type": "string" },
                  "status": { "type": "integer" }
                },
                "required": ["type", "title", "status"],
                "additionalProperties": false
              }
            }
          }
        }
      }
    }
  }
}
```

Run:
```sh
npx openapigen --spec spec.json --name MyClient
```

## What is the expected behavior?

The generator should:
- Recognize `application/zip` (with `format: "binary"`) as a binary response and produce a `*Stream` variant
- Extract the `404` error schema from `application/problem+json`

## What do you see instead?

Both the `200` and `404` responses are silently ignored. The generated client produces only the `orElse` fallback:

```typescript
"downloadArchive": (options) => HttpClientRequest.get(`/archive`).pipe(
  withResponse(options?.config)(HttpClientResponse.matchStatus({
    orElse: unexpectedStatus,
  })),
),
```

Every HTTP response — including `200` — triggers the `unexpectedStatus` error path. No `*Stream` variant is generated.

### Root cause

Two hard-coded content type checks in `OpenApiGenerator.ts` (`httpclient` format):

1. **Binary detection** (line 516) checks only `application/octet-stream`:
   ```typescript
   if (Predicate.isNotUndefined(content?.["application/octet-stream"]))
   ```

2. **JSON schema extraction** (line 437) checks only `application/json`:
   ```typescript
   const jsonResponseSchema = content?.["application/json"]?.schema
   ```

The `httpapi` format already has `isJsonMediaType` and `isBinaryMediaType` helpers that handle broader content types, but they are not wired into the `httpclient` format.

### Context

- **OpenAPI 3.1** states "any binary media type, including `application/octet-stream`, is sufficient to indicate binary content" (§4.4.2). The schema's `format: "binary"` is the canonical signal, not a specific MIME type.
- **RFC 9457** defines `application/problem+json` as the standard media type for HTTP problem details. It's widely used for error responses.
- The existing `isBinaryMediaType` helper is also too narrow — it misses common binary types like `application/zip`, `application/pdf`, `image/*`, `audio/*`, `video/*`.
- The runtime `HttpApiClient` already handles binary responses regardless of content type (it decodes based on schema encoding annotations, not HTTP headers).

craigsmitham:
## Summary

The `httpclient` format hard-codes content type checks to only `application/octet-stream` for binary detection and `application/json` for JSON schema extraction. This causes binary responses with types like `application/zip` and error responses with `application/problem+json` (RFC 9457) to be silently dropped — the generated client falls through to `orElse: unexpectedStatus` for every response.

- Wire the existing `isJsonMediaType` helper into `httpclient` JSON response schema extraction, so `application/problem+json` and other `+json` types are recognized
- Replace the hard-coded `application/octet-stream` binary check with iteration over all content types using `isBinaryMediaType` plus schema `format: "binary"` inspection
- Broaden `isBinaryMediaType` to cover `application/zip`, `application/gzip`, `application/pdf`, `image/*`, `audio/*`, `video/*`
- Add `findContentByMediaType` utility to search response content entries by media type predicate

Closes #1911

## Test plan

- [x] `application/zip` response generates `*Stream` variant with `binaryRequest`
- [x] Unknown MIME type with `format: "binary"` schema generates `*Stream` variant
- [x] `application/problem+json` error response generates `decodeError` call
- [x] Combined binary success + problem+json error in same operation works
- [x] All 66 existing tests pass (zero regressions)


craigsmitham:
Follow-up response-side work is now consolidated in Effect-TS/effect-smol#1979, with the broader problem statement in Effect-TS/effect#6368. This PR's scope is effectively superseded by the combined response-representations change so the response-model fixes can land together. The form-urlencoded request-body work remains separate.

craigsmitham:
Closing in favor of Effect-TS/effect-smol#1979. This PR covers a narrower slice of the same response-side problem space, and the active implementation has been consolidated into the broader response-representations change tracked in Effect-TS/effect#6368 and implemented in Effect-TS/effect-smol#1979 so the mixed-JSON and binary-success fixes can land together.

PR #1918, opened 2026-03-29
feat(Queue): add non-blocking drain API

cevr:
Adds `Queue.drain` — takes all available messages without blocking, returning `[]` on empty. Thought this could be useful for cleanup/shutdown paths where you want to grab whatever's there without waiting.

`takeAll` blocks until ≥1 item which seems intentional (returns `NonEmptyArray`, `collect` depends on it), so this is a separate non-blocking alternative following the same Done/error semantics as `clear` and `poll`.

Purely additive, totally optional — sorry if inconvenient. Made with the help of AI.

---

Should `TxQueue` and `PubSub` get a matching `drain` for parity?

### DIFF
--- a/.changeset/add-queue-drain.md
+++ b/.changeset/add-queue-drain.md
@@ -1,0 +1,10 @@
+---
+"effect": minor
+---
+
+Add `Queue.drain` for non-blocking take of all available messages.
+
+Unlike `Queue.takeAll` which blocks until at least one item is available,
+`Queue.drain` returns whatever is currently in the queue, including an
+empty array if no messages are present. Returns `[]` on graceful
+completion (`Done`) and propagates non-Done errors.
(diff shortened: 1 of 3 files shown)


mikearnaldi:
not sure about the name, maybe as an option to `takeAll` like `takeAll({ blocking: false})`

cevr:
The return type changes between modes (`NonEmptyArray<A>` → `Array<A>`, `E` → `ExcludeDone<E>`), which makes overloads awkward. thoughts on `takeAvailable`? 

but if you remain firm on the option to `takeAll` happy to go that route

tim-smart:
What is the issue with the current Queue.clear?

cevr:
@tim-smart 😳, yeah there's no need for this. I think the naming must have tripped up the agent since it might imply different semantics.

sorry about that - will close this

PR #1920, opened 2026-03-29
fix(platform-bun): settle response Promise on client abort

cevr:
Settles the `Promise<Response>` when a client aborts an HTTP request to a Bun server. Without this, Bun holds the request context alive and server-side scope finalizers never fire — streaming RPC fibers leak.

Same class of bug as [Firebun](https://trigger.dev/blog/firebun): Bun's `request.signal` abort fires, but if the handler's Promise never resolves, Bun pins the connection. One-line fix — resolve with 499 on abort.

Made with the help of AI.

### DIFF
--- a/.changeset/fix-bun-http-abort-promise.md
+++ b/.changeset/fix-bun-http-abort-promise.md
@@ -1,0 +1,10 @@
+---
+"@effect/platform-bun": patch
+---
+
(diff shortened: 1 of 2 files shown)


cevr:
Closing — tested in another project, this is a Bun runtime issue, not an Effect one.

Issue #1921, opened 2026-03-29
Build Toolchain Modernization: OXC / Rolldown / tsdown Migration


benpsnyder:
# Build Toolchain Modernization: tsdown / Rolldown / OXC / tsgo

## Summary

Migrate the Effect monorepo build pipeline from **tsc + Babel + Rollup** to **tsdown (unbundle) + Rolldown + OXC + tsgo**. tsdown 0.21.7 with `unbundle: true` produces the same 1:1 file output Effect ships today, replacing both tsc emit and Babel annotation in one step.

## Architecture

```
BEFORE                                    AFTER
------                                    -----
tspc -b tsconfig.packages.json            tsgo -b tsconfig.packages.json (emit .d.ts only)
  -> tsc per package (.js + .d.ts)           |
  -> babel annotate-pure-calls              tsdown per package (unbundle mode)
  -> dist/*.js + .d.ts + .js.map              -> rolldown (.js + .js.map)
                                              -> pureAnnotations() plugin (@__PURE__)
Type check: tspc -b (67s)                     -> native sourcemaps
Bundle analysis: Rollup + 4 plugins
                                          Type check: tsgo -b (7.3s, 9.2x faster)
                                          Bundle analysis: Rolldown (native Rust)
```

## Implementation Status

| # | Workstream | Status | Result |
|---|---|---|---|
| 1 | tsdown Unbundle POC | **Done** | `@effect/platform-browser` builds in ~800ms, 1:1 output, tests pass |
| 2 | Pure Annotation Plugin | **Done** | `@effect/oxc/rolldown-plugin-pure-annotations` -- 17 tests, MagicString sourcemaps, runs in tsdown transform pipeline |
| 3 | ~~isolatedDeclarations~~ | **Dropped** | Per maintainer: tsgo handles declarations, no source code churn needed |
| 4 | tsdown Full Rollout | **Done** | All 31 packages migrated, `pnpm build` succeeds, effect builds in 681ms |
| 5 | Rolldown Bundle Tool | **Done** | `Rolldown.ts` replaces `Rollup.ts` in `@effect/bundle`, 4 plugins removed |
| 6 | tsgo Default Checker | **Done** | `pnpm check` = tsgo (7.3s vs 67.5s) |
| 7 | Dependency Cleanup | **Done** | Babel (5 packages), Rollup plugins (4), terser removed |
[...]

PR #1923, opened 2026-03-29
feat: replace tsc + Babel + Rollup with tsdown + Rolldown + tsgo

benpsnyder:
> Note: please disregard the unfortunate branch name. Otherwise, this PR should hold weight.

## Why

| Metric | Before (tsc + Babel) | After (tsdown + tsgo) | Change |
|:--|--:|--:|:--|
| Full build | 67.4s | 10.1s | **6.7x faster** |
| Type check | 64.5s | 3.2s (warm) | **20x faster** |
| `effect` package build | 19.6s | 1.6s | **12x faster** |
| `@__PURE__` annotations | 3,472 | 3,749 | **+8% more** |
| Build dependencies | 47 | 38 | **-9 packages** |
| Babel in tree | Yes | No | **eliminated** |
| Tests (comprehensive) | 2,231 pass | 2,231 pass | identical |
| `.d.ts` files | tsc output | tsc output | byte-identical |

Measured on darwin-arm64, 5 build runs averaged, 3 type-check runs averaged.

---

## What

Replaces **tsc + Babel + Rollup** with **tsdown (unbundle) + Rolldown + tsgo** across all 31 packages.

### Build pipeline

```
BEFORE                                  AFTER
tspc -b tsconfig.packages.json          tsgo -b tsconfig.packages.json
  -> tsc per package (.js + .d.ts)        -> tsgo per package (.d.ts only)
  -> babel annotate-pure-calls            |
  -> 67.4s                              tsdown per package (unbundle mode)
                                          -> rolldown (.js + .js.map)
Type check: tspc (64.5s)                  -> pureAnnotations() plugin
                                          -> 10.1s total
Bundle analysis:
  Rollup + 4 plugins                  Type check: tsgo (3.2s warm)
                                      Bundle analysis: Rolldown (native Rust)
```

### Per-package config

Each of the 31 packages gets a `tsdown.config.ts`:

```ts
import { pureAnnotations } from "@effect/oxc/rolldown-plugin-pure-annotations"
import { defineConfig } from "tsdown"

export default defineConfig({
  entry: ["src/**/*.ts"],
  unbundle: true,
  format: "esm",
  dts: false,
  fixedExtension: false,
  sourcemap: true,
  clean: false,
  plugins: [pureAnnotations()]
})
```

Build script becomes `"build": "tsdown"` (was `tsc -b && babel`).

### Pure annotation plugin

`packages/tools/oxc/src/rolldown-plugin-pure-annotations.ts` -- a rolldown transform plugin that:
- Parses source with `oxc-parser`, walks the AST to find call/new expressions during module initialization
- Inserts `/* @__PURE__ */` via `MagicString` with correct sourcemaps
- Handles OXC's `ParenthesizedExpression`, IIFE detection, idempotency
- 17 dedicated tests

Produces **3,749 annotations** vs Babel's 3,472 (+8%). The plugin is more thorough in class extends and nested initialization patterns.

### `@effect/bundle`

`Rolldown.ts` replaces `Rollup.ts`:
- Built-in module resolution (no `@rollup/plugin-node-resolve`)
- Built-in OXC transform (no `rollup-plugin-esbuild`)
- Built-in minification (no `@rollup/plugin-terser`)
- Built-in define (no `@rollup/plugin-replace`)

### Dependencies

**Removed (11):** `@babel/cli`, `@babel/core`, `@babel/plugin-transform-export-namespace-from`, `@babel/plugin-transform-modules-commonjs`, `babel-plugin-annotate-pure-calls`, `rollup`, `@rollup/plugin-node-resolve`, `@rollup/plugin-replace`, `@rollup/plugin-terser`, `rollup-plugin-esbuild`, `terser`

**Added (2):** `tsdown` (^0.21.7), `rolldown` (1.0.0-rc.12)

---

## Honest tradeoffs

| Tradeoff | Detail |
|:--|:--|
| Dist code style differs | Rolldown uses tabs, `export {}` at end, `//#region` comments. Functionally equivalent. |
| Rolldown runtime helper | ~350 byte `_virtual/_rolldown/runtime.js` added per package for `export *` handling |
| Dist sourcemaps larger | 30 MB -> 41 MB for effect (richer rolldown maps). Only affects dev, not published size. |
| rolldown is pre-1.0 | v1.0.0-rc.12, pinned. API is stable (Rollup-compatible). |
| ~153 fewer annotations than theoretical max | 3,749 of 3,902 possible (96%). Gap is in barrel re-export patterns that rolldown restructures. Still 8% more than Babel. |

---

## Test plan

- [x] `pnpm build` succeeds across all 31 packages
- [x] `effect` package: 1.6s via tsdown (was 19.6s with tsc + babel)
- [x] `@__PURE__` annotations: 3,749 in effect dist (8% more than Babel's 3,472)
- [x] Chunk tests: 97/97 pass
- [x] Option tests: 58/58 pass
- [x] Array tests: 144/144 pass
- [x] Schema tests: 1,420/1,420 pass
- [x] Stream tests: 269/269 pass
- [x] OXC plugin tests: 57/57 pass
- [x] `pnpm check` (tsgo): 3.2s warm, zero errors
- [x] `.d.ts` files byte-identical to tsc baseline
- [x] All exports match (verified on Effect, Schema, Function, Stream, index)
- [x] Source maps valid and point to `.ts` sources
- [ ] Full CI test suite
- [ ] Bundle fixture size comparison vs baseline

Tracking issue: #1921


benpsnyder:
@mikearnaldi -- ready for feedback on this effort. Let me know what works and what doesn't.

Full benchmark report is in `Report.md` in the branch. The PR description has the summary, but a few details worth calling out:

### Build time breakdown

| Step | Old | New (tsgo path) |
|:--|--:|--:|
| Declaration emit | 56.8s (tspc) | 6.0s (tsgo) |
| JS transform | ~10.6s (babel) | 4.3s (tsdown) |
| **Total** | **67.4s** | **10.1s** |

tsgo eliminates 50s of serial tsc work. tsdown replaces babel at 2.5x throughput.

### Per-package scaling

| Package | Old | New | Speedup |
|:--|--:|--:|:--|
| `effect` (403 files) | 19.6s | 1.6s | 12x |
| `platform-browser` (16 files) | 2.27s | 73ms | 31x |
| `vitest` (8 files) | 2.04s | 30ms | 68x |

Smaller packages see bigger gains because babel's ~600ms startup is eliminated.

### What's honestly different in dist

1. Code style: tabs, `export {}` at end, `//#region` comments (functionally equivalent)
2. Small `_virtual/_rolldown/runtime.js` (~350 bytes) per package for `export *` handling
3. Sourcemaps are larger (+11 MB for effect) -- richer rolldown maps
4. rolldown is v1.0.0-rc.12 (pinned, API stable, but pre-1.0)

benpsnyder:
# `@effect/oxc/rolldown-plugin-pure-annotations`

A rolldown transform plugin that adds `/* @__PURE__ */` comments to compiled output so bundlers can tree-shake unused Effect imports.

## What it does

When a downstream consumer writes:

```ts
import { pipe } from "effect/Function"
```

Their bundler needs to know that every other top-level call in Effect's compiled output is side-effect-free and can be dropped. The `@__PURE__` comment before a call expression is the universal signal for this:

```js
const schedule = /* @__PURE__ */ dual(2, (self, sched) => ...)
```

Without it, the bundler must assume `dual(2, ...)` has side effects and keep it in the bundle. With it, the call is eliminated if `schedule` is never imported.

The plugin inserts these comments at **3,749 call sites** across the `effect` package -- 8% more than the Babel plugin it replaces.

## How it works

The plugin hooks into rolldown's `transform` phase. For each source file:

```
source.ts
  -> oxc-parser parseSync()        [~0.1ms per file, native Rust binary]
  -> setParents()                   [attach _parent refs for upward traversal]
  -> walk() + collectPositions()    [single DFS pass, ~0.2ms]
  -> MagicString insertions         [O(n) string splicing with sourcemap tracking]
  -> return { code, map }           [rolldown composes map into final output]
```

### The five-filter pipeline

Every `CallExpression` and `NewExpression` in the AST is tested against five filters. A call must pass ALL of them to get annotated:

```
CallExpression found
  │
  ├─ Already annotated?          skip (idempotent)
  │   isPureAnnotated() checks 30 chars before the call for /*@__PURE__*/
  │
  ├─ Used as a callee?           skip (it's the inner call in foo()())
  │   isUsedAsCallee() walks up through ParenthesizedExpression wrappers
  │
  ├─ Inside a callee position?   skip (nested in a callee chain)
  │   isInCallee() walks ancestors, stops at function/statement boundaries
  │
  ├─ Executed during init?       must be true (top-level or inside IIFE)
  │   isExecutedDuringInitialization() checks every enclosing function
  │   is itself used as a callee (the IIFE pattern)
  │
  └─ In assignment context?      must be true
      isInAssignmentContext() checks for VariableDeclaration,
      AssignmentExpression, ClassDeclaration, ClassExpression
      -- OR isInExportDefault()
```

If a call passes all five, its `start` position is collected. Positions are deduplicated and sorted descending so MagicString inserts right-to-left without shifting earlier offsets.

### OXC-specific fixes

Three bugs were found and fixed during development, all caused by OXC's AST differing from estree:

**1. `ParenthesizedExpression` unwrapping**

OXC wraps grouped expressions in `ParenthesizedExpression` nodes that estree doesn't have. Without unwrapping, `(() => foo())()` fails IIFE detection because the arrow function's parent is `ParenthesizedExpression`, not the outer `CallExpression`.

```ts
// isUsedAsCallee walks up through ParenthesizedExpression
function isUsedAsCallee(node: AstNode): boolean {
  let current = node
  let parent = current._parent
  while (parent?.type === "ParenthesizedExpression") {
    current = parent
    parent = current._parent
  }
  return isCallOrNew(parent) && parent.callee === current
}
```

**2. `isInCallee` boundary ordering**

Function boundaries must be checked BEFORE callee detection. Otherwise, calls inside IIFE bodies are incorrectly skipped (the function is detected as a callee before being recognized as a boundary):

```ts
function isInCallee(node: AstNode): boolean {
  let current = node._parent
  while (current) {
    if (isStatement(current) || isFunction(current)) break  // boundary first
    if (isUsedAsCallee(current)) return true                 // then callee check
    current = current._parent
  }
  return false
}
```

**3. `isPureAnnotated` regex**

The plugin inserts `/* @__PURE__ */` (with spaces). The detection regex must tolerate spaces to avoid double-annotating on re-runs:

```ts
/\/\*\s*[#@]__PURE__\s*\*\/\s*$/
```

This matches both `/*@__PURE__*/` (no spaces) and `/* @__PURE__ */` (with spaces), as well as the `#` variant.

## Why it's fast

**Parse: native Rust.** `oxc-parser`'s `parseSync` is a Rust binary called via N-API. It parses a 434 KB TypeScript file (Effect.ts) in ~0.5ms. Babel's parser takes ~15ms for the same file. The parser is the bottleneck in the old pipeline; eliminating it is the single largest speedup.

**Single pass.** The AST is walked once (depth-first). Each node is tested against the five filters using only parent-pointer lookups -- no second pass, no scope analysis, no type resolution.

**No serialization.** The AST lives in JavaScript memory as plain objects. There's no JSON serialization step between the Rust parser and the JS visitor (oxc-parser returns the AST directly via N-API structured clone).

**String, not AST, output.** MagicString does character-level insertions on the original source string. There's no AST-to-code printing step (which is where Babel spends most of its time). The output is the original source with 16-char comments spliced in.

**Runs inside rolldown.** The plugin's `transform` hook runs in rolldown's module graph traversal. There's no separate post-build pass, no second parse of the output, no sourcemap composition. Rolldown handles map merging internally.

### Measured overhead

The plugin adds approximately **0.1--0.3ms per file** to the transform phase. For the `effect` package (403 files), total plugin overhead is ~80ms -- invisible within tsdown's 1.6s build time. Babel's equivalent work takes 2.6s.

## Usage

```ts
// tsdown.config.ts
import { pureAnnotations } from "@effect/oxc/rolldown-plugin-pure-annotations"
import { defineConfig } from "tsdown"

export default defineConfig({
  entry: ["src/**/*.ts"],
  unbundle: true,
  format: "esm",
  dts: false,
  sourcemap: true,
  clean: false,
  plugins: [pureAnnotations()]
})
```

## Testing

17 dedicated tests in `packages/tools/oxc/test/rolldown-plugin-pure-annotations.test.ts`:

| Test | What it verifies |
|:--|:--|
| Basic variable assignment | `const x = foo()` -> annotated |
| NewExpression | `const x = new Foo()` -> annotated |
| Class extends | `class Foo extends bar()` -> annotated |
| Inside functions | `function f() { bar() }` -> NOT annotated |
| IIFEs | `(() => foo())()` -> both outer and inner annotated |
| Idempotency (`@__PURE__`) | Already-annotated code unchanged |
| Idempotency (`#__PURE__`) | Babel-style annotations recognized |
| Callee skipping | `factory()()` -> only outer call |
| Export default | `export default foo()` -> annotated |
| Assignment expression | `x = foo()` -> annotated |
| Bare calls | `foo()` without assignment -> NOT annotated |
| Multiple declarations | Multiple calls in one statement -> all annotated |
| New in class extends | `class Foo extends new Bar()` -> annotated |
| node_modules filtering | Files in node_modules -> skipped |
| Non-JS filtering | `.css`, `.json` files -> skipped |
| .ts file processing | TypeScript files -> processed |
| .jsx file processing | JSX files -> processed |

tim-smart:
At this stage I would prefer to keep the toolchain simple :) We will migrate to plain tsgo at some point.

benpsnyder:
> At this stage I would prefer to keep the toolchain simple :) We will migrate to plain tsgo at some point.

I respect your choice. For me, there is no way I am going to contribute to v4 with PRs if agentic development and performant tooling isn't a priority.

PR #1925, opened 2026-03-30
Add tagged overload to Schema.Opaque for scalar newtypes

kitlangton:
Adds a tagged overload to `Schema.Opaque` for scalar newtypes:

```ts
class QuestionID extends Schema.Opaque<QuestionID>()("QuestionID", Schema.String) {
  static make(id: string): QuestionID {
    return this.makeUnsafe(id)
  }
}
```

Previously `Opaque` couldn't be extended with scalar carriers (TS errors on primitives in constructor return types). The tag string also provides nominal distinctness between newtypes wrapping the same carrier. Existing tagless API is unchanged.

Motivated by the custom [`Newtype` wrapper](https://github.com/sst/opencode/blob/dev/packages/opencode/src/util/schema.ts#L19-L53) in opencode, used for domain IDs like:

[QuestionID](https://github.com/sst/opencode/blob/dev/packages/opencode/src/question/schema.ts):
```ts
class QuestionID extends Newtype<QuestionID>()("QuestionID", Schema.String) {
  static make(id: string): QuestionID {
    return this.makeUnsafe(id)
  }

  static ascending(id?: string): QuestionID {
    return this.makeUnsafe(Identifier.ascending("question", id))
  }
}
```

[PermissionID](https://github.com/sst/opencode/blob/dev/packages/opencode/src/permission/schema.ts):
```ts
class PermissionID extends Newtype<PermissionID>()("PermissionID", Schema.String) {
  static make(id: string): PermissionID {
    return this.makeUnsafe(id)
  }

  static ascending(id?: string): PermissionID {
    return this.makeUnsafe(Identifier.ascending("permission", id))
  }
}
```

This change makes the custom `Newtype` function unnecessary. `Schema.Opaque` handles it directly.

### DIFF
--- a/.changeset/thirty-pans-love.md
+++ b/.changeset/thirty-pans-love.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-browser": patch
+---
+
+allow Model.Class for indexeddb schemas
--- a/packages/platform-browser/src/IndexedDbTable.ts
+++ b/packages/platform-browser/src/IndexedDbTable.ts
@@ -42,14 +42,6 @@ export interface IndexedDbTable<
  */
 export type AnySchemaStruct = Schema.Top & {
   readonly fields: Schema.Struct.Fields
-  mapFields<To extends Schema.Struct.Fields>(
-    f: (fields: Schema.Struct.Fields) => To,
-    options?:
-      | {
-        readonly unsafePreserveChecks?: boolean | undefined
-      }
-      | undefined
-  ): Schema.Struct<To>
 }
 
 /**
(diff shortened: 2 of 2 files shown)


mikearnaldi:
It feels way too ad-hoc, Opaque should not change the schema it should just make the types opaque (not nominal) and Newtype is not only about primitives (you can newtype any type)

kitlangton:
> It feels way too ad-hoc, Opaque should not change the schema it should just make the types opaque (not nominal) and Newtype is not only about primitives (you can newtype any type)

Gotcha! I can keep this as my own local ad hoc helper.

PR #1945, opened 2026-03-31
add unstable machine prototype

mikearnaldi:
## Summary
- add an unstable `effect/unstable/machine/Machine` prototype with schema-backed tagged events and states, typed handlers, and a scoped actor runtime
- include focused runtime and dtslint coverage for the current API shape
- ignore `.repos` during vitest tsconfig-path discovery to avoid subtree noise

## Caveat
- this branch is fully vibe coded and should not yet be considered for merge
- this draft PR is for experimentation and API feedback only

### DIFF
--- a/.changeset/silent-cameras-wave.md
+++ b/.changeset/silent-cameras-wave.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add an unstable machine prototype with schema-backed tagged events and states, typed handlers, and a scoped actor runtime under `effect/unstable/machine/Machine`.
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ scratchpad/**/*
 # lalph
 .lalph/
 .repos/
+.specs/
 
 # ralph auto loop (runtime output)
 .ralph-auto/
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ scratchpad/**/*
 
 # Claude Code
 .claude/
+
+# OpenCode
+.opencode/
--- a/packages/effect/dtslint/unstable/machine/Machine.tst.ts
+++ b/packages/effect/dtslint/unstable/machine/Machine.tst.ts
@@ -1,4 +1,6 @@
+import * as Effect from "effect/Effect"
 import * as Schema from "effect/Schema"
+import * as ServiceMap from "effect/ServiceMap"
 import * as Machine from "effect/unstable/machine/Machine"
 import { describe, expect, it } from "tstyche"
 
@@ -57,6 +59,9 @@ describe("Machine", () => {
       | Created
       | Deleted
     >()
+    expect<Machine.DeferredServicesOf<typeof UserMachine>>().type.toBe<never>()
+    const next = Machine.next(UserMachine, Machine.initial(UserMachine), new Create({ email: "a@example.com" }))
+    expect<typeof next>().type.toBe<Effect.Effect<Uncreated | Created | Deleted, Machine.UnhandledEventError>>()
   })
 
   it("contextually types handlers", () => {
@@ -86,6 +91,34 @@ describe("Machine", () => {
         return new Created({ email: event.email })
       }
     })
+  })
+
+  it("does not infer deferred requirements from plain effectful handlers", () => {
+    class Create extends Schema.TaggedClass<Create, { readonly _: unique symbol }>()(
+      "Create",
+      { email: Schema.String }
+    ) {}
+
+    class Uncreated extends Schema.TaggedClass<Uncreated, { readonly _: unique symbol }>()(
+      "Uncreated",
+      {}
+    ) {}
+
+    class Created extends Schema.TaggedClass<Created, { readonly _: unique symbol }>()(
+      "Created",
+      { email: Schema.String }
+    ) {}
+
+    const machine = Machine.make({
+      events: [Create],
+      initial: () => new Uncreated({}),
+      states: [Uncreated, Created]
+    }).handlers("Uncreated")({
+      Create: ({ event }) => Effect.succeed(new Created({ email: event.email }))
+    })
+
+    expect<Machine.DeferredErrorOf<typeof machine>>().type.toBe<never>()
+    expect<Machine.DeferredServicesOf<typeof machine>>().type.toBe<never>()
   })
 
   it("accepts events as a tuple of tagged schemas", () => {
@@ -195,5 +228,50 @@ describe("Machine", () => {
         return new Unauthenticated({})
       }
     })
+  })
+
+  it("tracks deferred effect errors and services separately from plan requirements", () => {
+    class DeferredError extends Schema.TaggedErrorClass<DeferredError, { readonly _: unique symbol }>()(
+      "DeferredError",
+      {}
+    ) {}
+
+    class DeferredDependency extends ServiceMap.Service<DeferredDependency, {
+      readonly emit: Effect.Effect<void, DeferredError>
+    }>()("test/Machine/DeferredDependency") {}
+
+    class Create extends Schema.TaggedClass<Create, { readonly _: unique symbol }>()(
+      "Create",
+      {}
+    ) {}
+
+    class Idle extends Schema.TaggedClass<Idle, { readonly _: unique symbol }>()(
+      "Idle",
+      {}
+    ) {}
+
+    const machine = Machine.make({
+      events: [Create],
+      initial: () => new Idle({}),
+      states: [Idle]
+    }).handlers("Idle")({
+      Create: () =>
+        Effect.gen(function*() {
+          yield* Machine.defer(DeferredDependency.use((deferred) => deferred.emit))
+          return new Idle({})
+        })
+    })
+
+    const planned = Machine.plan(machine, Machine.initial(machine), new Create({}))
+
+    expect<Machine.PlanErrorOf<typeof machine>>().type.toBe<never>()
+    expect<Machine.PlanServicesOf<typeof machine>>().type.toBe<never>()
+    expect<Machine.DeferredErrorOf<typeof machine>>().type.toBe<DeferredError>()
+    expect<Machine.DeferredServicesOf<typeof machine>>().type.toBe<DeferredDependency>()
+    expect<Machine.ErrorOf<typeof machine>>().type.toBe<DeferredError>()
+    expect<Machine.ServicesOf<typeof machine>>().type.toBe<DeferredDependency>()
+    expect<typeof planned>().type.toBe<
+      Effect.Effect<Machine.Plan<Machine.StateSchemasOf<typeof machine>, typeof machine.event, Idle>, Machine.UnhandledEventError>
+    >()
   })
 })
--- a/packages/effect/dtslint/unstable/machine/Machine.tst.ts
+++ b/packages/effect/dtslint/unstable/machine/Machine.tst.ts
@@ -1,6 +1,6 @@
 import * as Effect from "effect/Effect"
 import * as Schema from "effect/Schema"
-import * as ServiceMap from "effect/ServiceMap"
+import * as Context from "effect/Context"
 import * as Machine from "effect/unstable/machine/Machine"
 import { describe, expect, it } from "tstyche"
 
@@ -236,7 +236,7 @@ describe("Machine", () => {
       {}
     ) {}
 
-    class DeferredDependency extends ServiceMap.Service<DeferredDependency, {
+    class DeferredDependency extends Context.Service<DeferredDependency, {
       readonly emit: Effect.Effect<void, DeferredError>
     }>()("test/Machine/DeferredDependency") {}
 
--- a/packages/effect/dtslint/unstable/machine/Machine.tst.ts
+++ b/packages/effect/dtslint/unstable/machine/Machine.tst.ts
@@ -41,7 +41,7 @@ describe("Machine", () => {
 
     const UserMachine = Machine.make({
       events: [Create, Rename, Delete],
-      initial: new Uncreated({}),
+      initial: () => new Uncreated({}),
       states: [Uncreated, Created, Deleted]
     }).handlers({
       Uncreated: {
@@ -78,7 +78,7 @@ describe("Machine", () => {
 
     Machine.make({
       events: [Create],
-      initial: new Uncreated({ count: 0 }),
+      initial: () => new Uncreated({ count: 0 }),
       states: [Uncreated, Created]
     }).handlers({
       Uncreated: {
@@ -115,7 +115,7 @@ describe("Machine", () => {
 
     const machine = Machine.make({
       events: [Create, Rename],
-      initial: new Idle({}),
+      initial: () => new Idle({}),
       states: [Idle, Running]
     }).handlers({
       Idle: {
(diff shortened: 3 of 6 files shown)


mikearnaldi:
@davidkpiano materialize youself here!

davidkpiano:
Quick takes (not AI generated, real thoughts):

- Unifying state value + data = great idea, and even necessary (I hope to do this in XState v6 once I figure out the types + good API)
- Really like schema validation everywhere; adds an extra layer of robustness
- No viz obviously but I can fix that
- It's a flat FSM, not a full-on statechart, which is fine: but do you plan on supporting nested states eventually? Important to think about that early on re: the shape of `Schema.TaggedClass`

spaceemotion:
Thanks for starting a prototype on this! A couple feedback points that came to mind:

1. why do i need to declare an initial value? can I declare the initial state type instead, and then provide the initial value via `.start()`? (since some initial state may be context-dependent, like the current user ID).  that would allow to have the 'blueprint' of a machine once, and then start it with different initial state values.

2. how does it handle handlers that return failing effects?

davidkpiano:
> 1. why do i need to declare an initial value? can I declare the initial state type instead, and then provide the initial value via `.start()`? (since some initial state may be context-dependent, like the current user ID).  that would allow to have the 'blueprint' of a machine once, and then start it with different initial state values.

A notion of `input` is a good idea. I'll push back on allowing setting the start state to _any_ state though as that sorta defeats the notion of a finite state machine (like saying "start the game in the winning state")

spaceemotion:
> I'll push back on allowing setting the start state to any state though as that sorta defeats the notion of a finite state machine

right, yeah - i'd expect Effect to both check this via the types + at runtime. But having a FSM with a customizable initial state would reduce the need for an empty "boot" state that then needs to transition to its real start first.

mikearnaldi:
what about fixing the initial state to an empty `Initial` so that you'd manage transitions like:

```ts
{
  Initial: {
    Create: ...
  }
}
```

basically forcing state machines to start with an empty state

mikearnaldi:
> Quick takes (not AI generated, real thoughts):
> 
> * Unifying state value + data = great idea, and even necessary (I hope to do this in XState v6 once I figure out the types + good API)
> * Really like schema validation everywhere; adds an extra layer of robustness
> * No viz obviously but I can fix that
> * It's a flat FSM, not a full-on statechart, which is fine: but do you plan on supporting nested states eventually? Important to think about that early on re: the shape of `Schema.TaggedClass`

- forcing schemas is basically to allow for persistence by default and cluster integration where we would derive rpc, we don't do internal validations given inputs are typed
- viz, yeah that's currently a big miss
- yes currently it is a flat FSM I'm not too experiences in state charts, I'd love your input here re the API

davidkpiano:
Trying to fully strongly type statecharts will only achieve two things:

1. It'll make you extremely good in TypeScript
2. It'll make you give up on trying to fully strongly type statecharts

spaceemotion:
> It's a flat FSM, not a full-on statechart, which is fine: but do you plan on supporting nested states eventually?

> I'm not too experiences in state charts, I'd love your input here re the API

It would be good if a state chart could support both, nested re-usable machines.

For example, taking computer games here as reference: With XState and our mini-games we tend to have a reusable "pick and select" or "show dialogue choice" machine that we can spawn as a sub-machine.

I guess that would also require that we can we serialize another machine's state in another machine's state.

TheRealDevKat:
I really love this. :heart: Its exactly what I need.
Was working on botching a very basic FSM for a private project and talked with space about it, which apparently broke out this whole thing.

I will probably directly use this, as its way cleaner than anything I might have done. Would really love if this would become core.

spaceemotion:
Something that just came to mind:

XState has a `context` inside its machine. I am guessing that with Effect, we can just 'inject' context via the ServiceMap like usually, right? So when I start a machine, and it has dependencies, I just need to provide them up front.

However, what happens when I have a side-effect in a handler that accesses the context and mutates it, but I use `plan` to see what would happen for any given event. Wouldn't that mess with any outside state?

Not sure if there needs to be guardrails, or reversible elements here (e.g. via the Tx... features)

mikearnaldi:
> Something that just came to mind:
> 
> XState has a `context` inside its machine. I am guessing that with Effect, we can just 'inject' context via the ServiceMap like usually, right? So when I start a machine, and it has dependencies, I just need to provide them up front.
> 
> However, what happens when I have a side-effect in a handler that accesses the context and mutates it, but I use `plan` to see what would happen for any given event. Wouldn't that mess with any outside state?
> 
> Not sure if there needs to be guardrails, or reversible elements here (e.g. via the Tx... features)

same in XState, what happens if your handler mutates the context? also I am kind of uncertain about "plan" it feels nice in theory but in practice stuff will run in handlers so it's kind of never "pure"

TheRealDevKat:
I'm missing a way to have guarded transitions, like [xstate](https://stately.ai/docs/guards).
Basically I heavily use `Machine.enabled` to know in the UI if a transition is possible. But sometimes a transition might be technically possible but not really wanted, like trying to save the game when no game is running.
Right now I would need the same condition/guard (`context.gameRunning !== true`)  in the handler and in my UI, because currently `Machine.enabled` always has the transition.

Having the transition hidden from `Machine.enabled` would be the cleanest solution, especially if the guard can be an effect.

mikearnaldi:
> Trying to fully strongly type statecharts will only achieve two things:
> 
> 1. It'll make you extremely good in TypeScript
> 2. It'll make you give up on trying to fully strongly type statecharts



> Trying to fully strongly type statecharts will only achieve two things:
> 
> 1. It'll make you extremely good in TypeScript
> 2. It'll make you give up on trying to fully strongly type statecharts

Thanks for putting me in this insane rabbit hole by the way

SandroMaglione:
What's the objective of this module? A "complete" FSM, or moving towards a more complete model like `xstate`? Are you looking for a feedback on the current API, or how to expand it further/what features to add?

mikearnaldi:
> What's the objective of this module? A "complete" FSM, or moving towards a more complete model like `xstate`? Are you looking for a feedback on the current API, or how to expand it further/what features to add?

the objective is to get this integrated into cluster & atom, ideally we'd be able to model complex domains like `xstate` but the primary objective is to get to a type safe api and I am not sure if state charts are possible

mikearnaldi:
> What's the objective of this module? A "complete" FSM, or moving towards a more complete model like `xstate`? Are you looking for a feedback on the current API, or how to expand it further/what features to add?

oh and yes ofc, suggestions are welcome!

TheRealDevKat:
@mikearnaldi I created a POC for my proposed handler guard, that allows to use this syntax to guard specific events:

```typescript
Machine.make({/*...*/})
.handlers("Uncreated")({
  Create: ({ event }) => new Created({ user: { id: "admin", email: event.email } }) // still works
})
.handlers("Created")({
  Rename: {
    guard: ({ state }) => state.user.id !== "admin", // pure function
    handler: ({ state, event }) => new Created({ user: { ...state.user, email: event.email } })
  },
  Delete: {
    guard: ({ state }) => Effect.succeed(state.user.id !== "admin"), // as effect
    handler: ({ state }) => new Deleted({ userId: state.user.id })
  }
})
```

it's mostly AI created, as my typescript foo is not even close to good enough for writing (and sometimes understanding) effect source.

I'm not sure what the guidelines for AI commits (and PRs on PRs) are, so you can see it here: https://github.com/Effect-TS/effect-smol/compare/machine-prototype...TheRealDevKat:effect-smol:machine-prototype

SandroMaglione:
A few features from `xstate` I would consider necessary to make this usable:
- `context`: some way to store "global state"
- State-specific context: some way to type check that a state guarantees having a specific value (working in progress in `xstate` v6, it should be possible to achieve as part of the `Schema` for the states)
- `input`: some way to pass initial values for context at creation/init time (e.g. for react component)
- **`actors`**: necessary for composition, and sending events between them (type safe with schema, errors included)
- `actions`: some way to execute functions when an event is triggered, or after a transition

Then there may be a few other features around guards, `always` transitions, delayed transition and such.

More than open to contribute for some/all of these if okay.

mikearnaldi:
> A few features from `xstate` I would consider necessary to make this usable:
> 
> * `context`: some way to store "global state"
> * State-specific context: some way to type check that a state guarantees having a specific value (working in progress in `xstate` v6, it should be possible to achieve as part of the `Schema` for the states)
> * `input`: some way to pass initial values for context at creation/init time (e.g. for react component)
> * **`actors`**: necessary for composition, and sending events between them (type safe with schema, errors included)
> * `actions`: some way to execute functions when an event is triggered, or after a transition
> 
> Then there may be a few other features around guards, `always` transitions, delayed transition and such.
> 
> More than open to contribute for some/all of these if okay.

- `context` is not needed, effect has native context propagation
- `State-specific context` please make an example, I don't understand "type-check that a state has a specific value"
- `input` done already
- `actors` done already
- `actions` why? isn't it enough to have a transition effect that can allow any HOF to be applied?

mikearnaldi:
> @mikearnaldi I created a POC for my proposed handler guard, that allows to use this syntax to guard specific events:
> 
> ```ts
> Machine.make({/*...*/})
> .handlers("Uncreated")({
>   Create: ({ event }) => new Created({ user: { id: "admin", email: event.email } }) // still works
> })
> .handlers("Created")({
>   Rename: {
>     guard: ({ state }) => state.user.id !== "admin", // pure function
>     handler: ({ state, event }) => new Created({ user: { ...state.user, email: event.email } })
>   },
>   Delete: {
>     guard: ({ state }) => Effect.succeed(state.user.id !== "admin"), // as effect
>     handler: ({ state }) => new Deleted({ userId: state.user.id })
>   }
> })
> ```
> 
> it's mostly AI created, as my typescript foo is not even close to good enough for writing (and sometimes understanding) effect source.
> 
> I'm not sure what the guidelines for AI commits (and PRs on PRs) are, so you can see it here: [machine-prototype...TheRealDevKat:effect-smol:machine-prototype](https://github.com/Effect-TS/effect-smol/compare/machine-prototype...TheRealDevKat:effect-smol:machine-prototype)

wouldn't it be better to have it simply as a HOF around the handler? why do we need the guard to be separate?

TheRealDevKat:
> > @mikearnaldi I created a POC for my proposed handler guard, that allows to use this syntax to guard specific events:
> > -- snip --
> > it's mostly AI created, as my typescript foo is not even close to good enough for writing (and sometimes understanding) effect source.
> > I'm not sure what the guidelines for AI commits (and PRs on PRs) are, so you can see it here: [machine-prototype...TheRealDevKat:effect-smol:machine-prototype](https://github.com/Effect-TS/effect-smol/compare/machine-prototype...TheRealDevKat:effect-smol:machine-prototype)
> 
> wouldn't it be better to have it simply as a HOF around the handler? why do we need the guard to be separate?

I'm fine with that, as long as the guard can be called independently, so `Machine.enabled` can filter the events based on the result.

<details>
<summary>Obsolete suggestion</summary>

So maybe something like this?

```typescript
Machine.withGuard(guard, handler)
// and
handler.pipe(Machine.withGuard(guard))
```

<details>

TheRealDevKat:
Updated my fork, syntax is now this:
```typescript
Machine.make({/*...*/})
.handlers("Created")({
  Rename: Machine.when(
    ({ state }) => state.user.id !== "admin",
    ({ state, event }) => new Created({ user: { ...state.user, email: event.email } })
  ),
  Delete: ({ state }) => new Deleted({ userId: state.user.id })
})
```

SandroMaglione:
> > A few features from `xstate` I would consider necessary to make this usable:
> > 
> > * `context`: some way to store "global state"
> > * State-specific context: some way to type check that a state guarantees having a specific value (working in progress in `xstate` v6, it should be possible to achieve as part of the `Schema` for the states)
> > * `input`: some way to pass initial values for context at creation/init time (e.g. for react component)
> > * **`actors`**: necessary for composition, and sending events between them (type safe with schema, errors included)
> > * `actions`: some way to execute functions when an event is triggered, or after a transition
> > 
> > Then there may be a few other features around guards, `always` transitions, delayed transition and such.
> > More than open to contribute for some/all of these if okay.
> 
> * `context` is not needed, effect has native context propagation
> * `State-specific context` please make an example, I don't understand "type-check that a state has a specific value"
> * `input` done already
> * `actors` done already
> * `actions` why? isn't it enough to have a transition effect that can allow any HOF to be applied?

State-specific context seems to be there already using `Schema` for states (each state has a name + some state).

I vibe coded a sub-branch of this to validate and visualise how this would integrate inside effect-atom ([branch here](https://github.com/SandroMaglione/effect-smol/tree/sandro/machine)).

I will experiment a little on that sub-branch based on my xstate usage and report some feedback.

giovannilondero:
I just read the tests and would like to throw in here some points to consider that seem to be missing (based on my experience with XState, which I love), in addition to what's already been said above:
1. root events
2. parallel states
3. final states
4. entry/exit state events
5. immediate/delayed transitions
6. effects? I mean, you can do whatever you want in handlers, but I really like enqueing effects like in xstate/store (https://stately.ai/docs/xstate-store#effects-and-side-effects) so the handlers can stay pure
7. state tags (https://stately.ai/docs/tags)

davidkpiano:
I too am curious how we can do `(state, event) => [nextState, effects]` with a nice DX; that's the core philosophy of XState

jacob-roling:
I was recently inspired by davids talk to try building my own state machine actor primitive in Effect v4. I've been using it to great success in a CMS project I'm building for client side UI. The idea is 1. Define states 2. define events and 3. define handlers which are (state, event) => Effect(state) Here it is:

It handles:
- Typed, serializable states and events with Schema.
- Event name + data with tagged union.
- Tuneable mailbox capacity.
- Initial state passed in at instantiation time. 
- Provides a stream of state changes to subscribe to.
- Allows sending of typed events.
- Handlers are inferred from events (TS forces you to implement a handler for each event and narrows event types properly).
- You may use Match.value on the state in a handler to add guards. 
- Uses the transactional queue with message counting pattern (from drainable worker) for seamless synchronisation (calling drain on a worker awaits processing all events in queue until idle while still allowing more to be added concurrently).

https://gist.github.com/jacob-roling/57179d288c2c193a2c902c32e61cf10a

The API looks like this:

```ts
const actorFactory  = Actor.createFactory({
  name: "ActorName",
  state: State, // TaggedUnion
  events: Event, // TaggedUnion
  on: {
    EventName: (state, event) => Effect.succeed(newState)
  },
});
```
Then you have:
```ts
const actor = yield* actorFactory(initialState);
yield* actor.send(Event.cases.EventName({ data }));
const state = yield* actor.get;
yield* actor.drain; // Wait for actor to process all events currently in mailbox.
const stateStream: Stream<State> = actor.subscribe()
```

One thing I'm yet to figure out is how to pass actor references into other actors since you can't represent a reference with Schema since it isn't serializable. Though you can still do it through dependency injection. As long as those dependencies are in the context of the effect that creates the actor.

Qodestackr:
Am doing something with Xstate currently. I think current prototype still needs entry/exit effects ... actions that fire when you leave or enter a state not just when an event arrives. Currently am using implicits in my flow - I ente a DraftingOrder and want to kick off DraftOrderOrchestrator automatically, not wait for another event. It is abit clunky 

Saw this on YT channel and loved it 😅.

davidkpiano:
@mikearnaldi XState currently has a pure function transition API:

```ts
const [nextState, actions] = transition(machine, currentState, event)
```

What's stopping this (a pure, simple, boring pure function) from being used directly in Effect, rather than trying to reinvent it all in Effect? Just curious

jbmusso:
We're super excited by this pull request, and we would be happy to crash-test it on a live project and provide feedbacks. Let me know on Discord if that helps.

SandroMaglione:
> We're super excited by this pull request, and we would be happy to crash-test it on a live project and provide feedbacks. Let me know on Discord if that helps.

Same here ✋

Handfish:
Definitely excited to use the new state machine prototype. 

Here's what I've been using for the connection management for my web game:
https://github.com/Handfish/effstate/tree/sandbox/packages/effstate-v4

SandroMaglione:
I am working on top of this PR [in another branch here](https://github.com/SandroMaglione/effect-smol/tree/sandro/state-machines-effect).

I am using [`xstate` v6](https://github.com/statelyai/xstate/pull/5257) as a reference.

> What's stopping this (a pure, simple, boring pure function) from being used directly in Effect, rather than trying to reinvent it all in Effect? 

@davidkpiano I moved the API closer to the v6 model, where the transition is pure and you can queue actions (`effect`/`raise` for now):

```ts
const DeferredMachine = Machine.make({
  events: [Create],
  initial: () => new Uncreated({}),
  states: [Uncreated, Created]
}).handlers("Uncreated")({
  Create: (_, actions) => {
    actions.effect(DeferredLog.use((log) => log.push("created")))
    return new Created({ user: { id: "user-1", email: "a@example.com" } })
  }
})
```

This also covers `guards` by using if/else inside the transition function. I also added `entry`/`exit`. I am about to implement actors and actors composition as well.

@mikearnaldi feel free to use as a reference or copy if anything looks helpful

IMax153:
Hi @mikearnaldi,

PR #1946, opened 2026-04-01
improve Effect.repeat ergonomics

kitlangton:
Sometimes you want to repeat a successful effect while a response is still pending, but keep the response that finally succeeds.

```ts
const order = yield* Effect.repeat(fetchOrder(orderUrl), {
  schedule: Schedule.spaced("1 second"),
  while: (order) => order.status === "pending" || order.status === "processing"
})
```

That feels like the obvious thing to write.

Today, once a schedule is involved, `repeat` returns the schedule output instead of `order`. You can get this with `Schedule.passthrough(...)`, but in practice that can push you into the builder overload and make the code much uglier than just passing `schedule` and `while` directly.

My guess is that most people will be surprised by this. In the simple polling case, what you usually want is the final successful value, not the schedule output.

I do not think this has to be the final design. It might be better for `repeat` (or maybe when called with `while`?) to return the last successful value by default and move the schedule-output behavior to a separate API. This PR is just one small possible fix for the common case. Or possibly a different method could do this.

The ability to use Schedule's output type is very cool, but I feel like 90% of the time people will be confused why calling repeat with `while` and Schedule.spaced("1 second") returns a number instead of the result of the repeated effect.

### DIFF
--- a/.changeset/fair-squids-pull.md
+++ b/.changeset/fair-squids-pull.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add a `passthrough` option to `Effect.repeat` so scheduled repetitions can preserve the final successful effect value.
--- a/packages/effect/src/Effect.ts
+++ b/packages/effect/src/Effect.ts
@@ -7000,6 +7000,14 @@ export declare namespace Repeat {
   /**
    * @since 2.0.0
    * @category Repetition / Recursion
+   */
+  export type Output<A, O extends Options<A>> = O extends { until: Predicate.Refinement<A, infer B> } ? B
+    : O extends { while: Predicate.Refinement<A, infer B> } ? Exclude<A, B>
+    : A
+
+  /**
+   * @since 2.0.0
+   * @category Repetition / Recursion
    * @example
    * ```ts
    * import type { Effect } from "effect"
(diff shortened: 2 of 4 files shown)


tim-smart:
https://github.com/Effect-TS/effect-smol/pull/1952

Also another way to do it in v4:

```ts
import { Effect, Schedule } from "effect"

declare const fetchThing: Effect.Effect<{ foo: string }>

export const eff = fetchThing.pipe(
  Effect.repeat(($) =>
    $(Schedule.spaced(1000)).pipe(
      Schedule.passthrough,
      Schedule.while(({ input }) => input.foo === "bar")
    )
  )
)
```

kitlangton:
> #1952
> 
> Also another way to do it in v4:
> 
> ```ts
> import { Effect, Schedule } from "effect"
> 
> declare const fetchThing: Effect.Effect<{ foo: string }>
> 
> export const eff = fetchThing.pipe(
>   Effect.repeat(($) =>
>     $(Schedule.spaced(1000)).pipe(
>       Schedule.passthrough,
>       Schedule.while(({ input }) => input.foo === "bar")
>     )
>   )
> )
> ```

Woo! Awesome :) 

And yeah, I saw that! It's very cool but also a little scary to consider suggesting to my effect noob teammates 😜

PR #1948, opened 2026-04-01
chore: rename `dtslint` directories to `typetest`

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR renames `dtslint` directories to `typetest`.

I would like to suggest this change, because `dtslint` is the name of a legacy tool. `typetest` is similar to `test` or `benchmark` that are tool-agnostic. If you prefer, that could also be `tstest`, `testtypes`, `test-ts` or something else.


gcanti:
Thank you, I had been meaning to do this for some time 😅

mrazauskas:
> Thank you, I had been meaning to do this for some time 😅

Same here. It didn’t feel like a priority, but the day had come!

PR #1955, opened 2026-04-01
feat(sql-sqlite-do): add transaction support via DurableObjectStorage.transactionSync

### LINKED ISSUE #5987: effect/sql-sqlite-do doesn't support Cloudflare DO SQLite transaction model

3.19.14

Use `@effect/sql-sqlite-do` with any code that calls `sql.withTransaction`, such as `SqlEventJournal`:

```typescript
import { SqliteClient } from "@effect/sql-sqlite-do"
import * as SqlEventJournal from "@effect/sql/SqlEventJournal"
import * as EventLog from "@effect/experimental/EventLog"
import * as Config from "effect/Config"
import * as Layer from "effect/Layer"

// In a Durable Object constructor:
export class MyDurableObject extends DurableObject {
  constructor(ctx: DurableObjectState, env: Env) {
    super(ctx, env)

    const storage = ctx.storage.sql
    const sqlLayer = SqliteClient.layerConfig(Config.succeed({ db: storage }))

    // SqlEventJournal.make() creates tables fine, but...
    const journalLayer = SqlEventJournal.layer({
      entryTable: "events",
      remotesTable: "remotes"
    }).pipe(Layer.provide(sqlLayer))

    // When you later call journal.write(), it uses sql.withTransaction
    // which executes "BEGIN" - and that fails in DO SQLite
  }
}
```

The error occurs when `SqlEventJournal.write()` is called, which internally uses `sql.withTransaction`.

`sql.withTransaction` should work in Cloudflare Durable Object SQLite, or the `@effect/sql-sqlite-do` package should handle DO SQLite's transaction limitations.

Cloudflare DO SQLite does not support direct SQL transaction statements (`BEGIN`, `COMMIT`, `ROLLBACK`). From [Cloudflare's documentation](https://developers.cloudflare.com/durable-objects/api/sqlite-storage-api/):

> "sql.exec() cannot execute transaction-related statements like BEGIN TRANSACTION or SAVEPOINT."

However, DO SQLite provides **automatic write coalescing** - any series of synchronous writes are automatically atomic. So transactions can safely be no-ops.

### What do you see instead?

```
SqlError: Failed to execute statement
```

The underlying cause is that `BEGIN` is not a valid statement in DO SQLite.

**Root cause in code:**

In `@effect/sql-sqlite-do/src/SqliteClient.ts`, the `make` function calls `Client.make()` without overriding transaction commands:

```typescript
return Object.assign(
  (yield* Client.make({
    acquirer,
    compiler,
    transactionAcquirer,
    spanAttributes: [...],
    transformRows
    // Missing: beginTransaction, commit, rollback, savepoint, rollbackSavepoint
  })) as SqliteClient,
  ...
)
```

This uses the defaults from `@effect/sql/internal/client.ts`:
```typescript
beginTransaction = "BEGIN",
commit = "COMMIT",
rollback = "ROLLBACK",
```

### Additional information

Set no-op transaction commands since DO SQLite's automatic write coalescing provides atomicity:

```typescript
return Object.assign(
  (yield* Client.make({
    acquirer,
    compiler,
    transactionAcquirer,
    spanAttributes: [...],
    transformRows,
    // No-op for DO SQLite (automatic write coalescing handles atomicity)
    beginTransaction: "SELECT 1",
    commit: "SELECT 1",
    rollback: "SELECT 1",
    savepoint: (_name: string) => "SELECT 1",
    rollbackSavepoint: (_name: string) => "SELECT 1"
  })) as SqliteClient,
  ...
)
```

**Note:** `SqlEventLogServer.makeStorage()` works fine because it doesn't use `withTransaction`. The issue specifically affects `SqlEventJournal` and any user code using explicit transactions.

## Additional context

- Related: #3888 (D1 batch support - similar Cloudflare SQL limitation)
- Cloudflare docs: [DO SQLite Storage API](https://developers.cloudflare.com/durable-objects/api/sqlite-storage-api/)
- Cloudflare blog explaining why: ["If BEGIN TRANSACTION were permitted, any one Worker request anywhere in the world could effectively block your whole database"](https://blog.cloudflare.com/whats-new-with-d1/)

m9tdev:
## Summary

Adds transaction support to `@effect/sql-sqlite-do` using Cloudflare's `DurableObjectStorage.transactionSync()` API.

Closes Effect-TS/effect#5987

- New `storage` config option accepts the `DurableObjectStorage` instance — `db` is derived from `storage.sql`
- `withTransaction` wraps effects in a real SQLite transaction via `storage.transactionSync()` + `Effect.runSyncExitWith`
- Nested `withTransaction` calls run flat (no savepoint support — DO's `sql.exec()` rejects `SAVEPOINT`)
- Without `storage`, `withTransaction` dies with a clear error instead of silently failing
- Fully backward compatible — existing `{ db: ctx.storage.sql }` usage is unchanged

### Why

Cloudflare DO's `sql.exec()` [cannot execute transaction control statements](https://developers.cloudflare.com/durable-objects/api/sqlite-storage-api) (`BEGIN`, `COMMIT`, `ROLLBACK`, `SAVEPOINT`). The only way to get a transaction is via the callback-based `storage.transactionSync()` API. This cannot be implemented through `Client.makeWithTransaction` since that expects separate begin/commit/rollback steps.

### Usage

```typescript
// Before — transactions unsupported
SqliteClient.make({ db: ctx.storage.sql })

// After — transactions via transactionSync
SqliteClient.make({ storage: ctx.storage })
```

Effects inside `withTransaction` must be fully synchronous (no async, no `Effect.sleep`) — this is a fundamental constraint of DO's `transactionSync` API.

## Verification

Verified locally against a real Cloudflare DO runtime (miniflare/workerd):
https://github.com/m9tdev/effect-sql-sqlite-do-transaction-test

```
$ npm test

DO Transaction Integration Tests

  ✓ commit persists data
  ✓ rollback discards writes on failure
  ✓ read own writes within transaction
  ✓ nested transactions commit
  ✓ nested failure rolls back everything
  ✓ migrator runs migrations in a transaction
  ✓ schema model + sql.insert/update + makeRepository

7 passed, 0 failed
```

## Test plan

8 tests using real SQLite via `better-sqlite3` (new `test/Transaction.test.ts`):

- [x] Transaction commit persists data
- [x] Rollback on `Effect.fail` discards writes
- [x] Rollback on `Effect.die` (defect) discards writes
- [x] Nested `withTransaction` commits atomically
- [x] Nested failure rolls back entire transaction
- [x] Read-own-writes within transaction
- [x] Constraint violation triggers rollback
- [x] `withTransaction` without `storage` dies with clear error

### DIFF
--- a/.changeset/sql-sqlite-do-transaction-support.md
+++ b/.changeset/sql-sqlite-do-transaction-support.md
@@ -1,0 +1,13 @@
+---
+"@effect/sql-sqlite-do": minor
+---
+
+Add transaction support via `DurableObjectStorage.transactionSync`.
+
+Cloudflare DO's `sql.exec()` cannot execute transaction control statements (`BEGIN`, `COMMIT`, `ROLLBACK`, `SAVEPOINT`). Pass the `storage` option to enable real transactions:
+
+```typescript
+SqliteClient.make({ storage: ctx.storage })
+```
+
+Without `storage`, `withTransaction` now dies with a clear error message instead of silently failing.
--- a/packages/sql/sqlite-do/package.json
+++ b/packages/sql/sqlite-do/package.json
@@ -60,6 +60,8 @@
   },
   "devDependencies": {
     "@cloudflare/workers-types": "^4.20260131.0",
+    "@types/better-sqlite3": "^7.6.13",
+    "better-sqlite3": "^12.6.2",
     "effect": "workspace:^"
   },
   "peerDependencies": {
--- a/packages/sql/sqlite-do/src/SqliteClient.ts
+++ b/packages/sql/sqlite-do/src/SqliteClient.ts
@@ -2,8 +2,10 @@
  * @since 1.0.0
  */
 import type { SqlStorage } from "@cloudflare/workers-types"
+import { Clock } from "effect/Clock"
 import * as Config from "effect/Config"
 import * as Effect from "effect/Effect"
+import * as Exit from "effect/Exit"
 import * as Fiber from "effect/Fiber"
 import { identity } from "effect/Function"
 import * as Layer from "effect/Layer"
@@ -11,6 +13,7 @@ import * as Scope from "effect/Scope"
 import * as Semaphore from "effect/Semaphore"
 import * as ServiceMap from "effect/ServiceMap"
 import * as Stream from "effect/Stream"
+import * as Tracer from "effect/Tracer"
 import * as Reactivity from "effect/unstable/reactivity/Reactivity"
 import * as Client from "effect/unstable/sql/SqlClient"
 import type { Connection } from "effect/unstable/sql/SqlConnection"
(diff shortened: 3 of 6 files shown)


m9tdev:
Disclaimer: Created this with Claude Opus 4.6, but tested this in my own application.

I'm not sure if you like that there is an added dev dependencies, but I can remove those if needed.

IMax153:
Hi @m9tdev,

The GitHub issues linked to this PR have also been migrated to the `effect` repository:
- [Effect-TS/effect#5987](https://github.com/Effect-TS/effect/issues/5987)

Please use these migrated issue links in the replacement PR.

PR #1962, opened 2026-04-02
EFF-779 Add KeyValueStore.layerSql

tim-smart:
## Summary
- add `KeyValueStore.layerSql(options?)` table customization support via a new `LayerSqlOptions` interface (`table` defaults to `effect_key_value_store`)
- keep dialect-specific DDL / upsert behavior while preserving the layer error channel as `never`
- store SQL values as binary blobs across dialects and retain a `value_type` discriminator so strings and binary values round-trip correctly
- extract a reusable SQL KeyValueStore integration suite to `packages/effect/test/unstable/persistence/KeyValueStoreTest.ts`
- run the reusable suite against sqlite, postgres, and mysql2, including custom-table coverage for each dialect

## Validation
- pnpm lint-fix
- pnpm test packages/sql/sqlite-node/test/KeyValueStore.test.ts
- pnpm test packages/sql/pg/test/KeyValueStore.test.ts
- pnpm test packages/sql/mysql2/test/KeyValueStore.test.ts
- pnpm test packages/effect/test/unstable/persistence/KeyValueStore.test.ts
- pnpm check:tsgo
- pnpm docgen

### DIFF
--- a/packages/effect/src/unstable/persistence/KeyValueStore.ts
+++ b/packages/effect/src/unstable/persistence/KeyValueStore.ts
@@ -415,7 +415,7 @@ export const layerSql = (
           sql`
           CREATE TABLE IF NOT EXISTS ${table} (
             id VARCHAR(191) PRIMARY KEY,
-            value TEXT NOT NULL,
+            value BLOB NOT NULL,
             value_type SMALLINT NOT NULL
           )
         `,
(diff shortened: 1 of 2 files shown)


packages/sql/sqlite-node/test/KeyValueStore.test.ts
```diff
@@ -0,0 +1,75 @@
+import { NodeFileSystem } from "@effect/platform-node"
```
tim-smart: We also need to add tests for Postgres and MySQL. You can copy the reusable test suite format used for the persistence modules.

PR #1971, opened 2026-04-03
Add effectful support to CLI fallback prompts

joepjoosten:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds effectful fallback prompt support to unstable CLI params.

`withFallbackPrompt` can now accept either:
- a concrete `Prompt`
- an `Effect` that lazily produces a `Prompt`

This is implemented in the shared `Param` layer and exposed through both `Flag` and `Argument`, so prompts can be built dynamically only when a value is actually missing.

The PR also:
- adds runtime tests covering lazy prompt construction for flags and arguments
- adds type-level tests for the widened fallback prompt typing
- includes a changeset for the `effect` package

## Related

- Related Issue #
- Closes #

### DIFF
--- a/.changeset/add-schema-string-encoding.md
+++ b/.changeset/add-schema-string-encoding.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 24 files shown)


IMax153 (dismissed):
@joepjoosten - just needs a rebase and then we can run CI / get this merged.

PR #1984, opened 2026-04-06
fixes type inference when times option has provided in Effect.retry

### LINKED ISSUE #1982: `Effect.retry`: `while` Refinement narrowing is unsound when combined with `times`
## What is the problem

PR #1754 added Refinement narrowing support for the `while` option in `Effect.retry`, so that:

```ts
declare const myEffect: Effect.Effect<void, ErrorA | ErrorB>

const result = Effect.retry(myEffect, {
  while: (e): e is ErrorA => e._tag === "ErrorA",
})
// Correctly narrows error to: ErrorB
```

This works correctly in isolation — `while` means "keep retrying as long as the predicate holds", so the only errors that escape are those that **don't** match the predicate → `Exclude<E, E2>` is sound.

However, when `times` is also specified, the narrowing becomes **unsound**:

```ts
const result = Effect.retry(myEffect, {
  times: 3,
  while: (e): e is ErrorA => e._tag === "ErrorA",
})
// TypeScript infers: Effect<void, ErrorB>
// Actual possible error: ErrorA | ErrorB
```

With `times: 3`, the retry loop stops after 3 attempts **even if the error still matches the `while` predicate**. In that case, `ErrorA` can absolutely be the final error, but the type system says it can't.

## Root cause

In `Retry.Return` ([`[packages/effect/src/Effect.ts](https://github.com/Effect-TS/effect-smol/blob/main/packages/effect/src/Effect.ts)`](https://github.com/Effect-TS/effect-smol/blob/main/packages/effect/src/Effect.ts)), the narrowing logic checks for `schedule` first and skips narrowing if present:

```ts
O extends { schedule: Schedule<infer _O, infer _I, infer _E1, infer _R> } ? E   // no narrowing
: O extends { until: Predicate.Refinement<E, infer E2> } ? E2
: O extends { while: Predicate.Refinement<E, infer E2> } ? Exclude<E, E2>
: E
```

This correctly avoids narrowing when a `schedule` is present (since a schedule can terminate independently). But `times` is a separate option — at the type level it doesn't contribute a `schedule` field in `O`, even though internally it gets converted to one. So the presence of `times` does **not** prevent the `while` Refinement branch from matching.

The same issue applies to `Effect.repeat` with `while` + `times`.

## Expected behavior

When `times` is specified alongside `while` (with a Refinement), the error type should **not** be narrowed, because the retry can terminate due to exhaustion rather than the predicate.

## Suggested fix

Add a guard for `times` before the Refinement branches:

```ts
O extends { schedule: Schedule<infer _O, infer _I, infer _E1, infer _R> } ? E
: O extends { times: number } ? E   // times can exhaust retries, skip narrowing
: O extends { until: Predicate.Refinement<E, infer E2> } ? E2
: O extends { while: Predicate.Refinement<E, infer E2> } ? Exclude<E, E2>
: E
```

Note: the same consideration applies to `until` — if `times` is set, the `until` predicate may never get a chance to match the specific error subtype it narrows to, so `E2` alone may also be unsound. A more conservative fix would skip all Refinement narrowing when `times` is present.

jannabiforever:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When `Effect.retry` receives several options including `times`,
the error shouldn't be refined. It fixes this.

## Related

- Closes #1982

### DIFF
--- a/.changeset/kind-hounds-float.md
+++ b/.changeset/kind-hounds-float.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Make `Effect.retry` with `times` argument to propagate the original error.


tim-smart (approved):
Just needs a `pnpm changeset` :)

PR #1985, opened 2026-04-06
add .reverse() to idb select

tim-smart:


SandroMaglione:
What about adding a test for this as well?

PR #2043, opened 2026-04-16
F/schema batching

patroza:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/SchemaAST.ts
+++ b/packages/effect/src/SchemaAST.ts
@@ -1390,6 +1390,12 @@ export class Arrays extends Base {
       let issues: Arr.NonEmptyArray<Issue.Issue> | undefined
       const errorsAllOption = options.errors === "all"
 
+      const tasks: Array<{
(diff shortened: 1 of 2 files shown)


packages/effect/src/SchemaAST.ts
```diff
@@ -1390,31 +1390,75 @@ export class Arrays extends Base {
...
+          const tailj = tail[j]
+          tasks.push({
+            ast: tailj.ast,
+            index,
+            exit: toExitEffect(tailj.parser(Option.some(input[index]), options))
+          })
+        }
+      }
+
+      const exits = yield* Effect.forEach(tasks, (task) => task.exit, { concurrency: "unbounded" })
```
tim-smart, line 1442: We can't use Effect.forEach here as it is a big perf de-opt.

PR #2048, opened 2026-04-17
fix(ai-openai): accept "in_memory" for prompt_cache_retention in response schemas

matthew-sessions:
## Summary

- OpenAI's live API returns `prompt_cache_retention: "in_memory"` (underscore), but the OpenAPI spec only documents `"in-memory"` (hyphen). Every real response that includes this field fails strict decoding with `InvalidOutputError` at `["prompt_cache_retention"]`, breaking `createResponse`, `createChatCompletion`, and all response-stream events that carry the nested `Response` payload.
- Generated schemas (`CreateChatCompletionRequest`, `Response`, `CreateResponse`, `ResponseCompleted/Created/Failed/InProgress/Incomplete/QueuedEvent`) now accept `"in-memory" | "24h" | "in_memory"` so responses decode cleanly. The spec-documented hyphen form is preserved for requests.
- `codegen.yaml` gains a `replacements` rule so future regenerations from the Stainless spec keep the fix.

## Repro (before the fix)

Minimal `createResponse` mock with `prompt_cache_retention: "in_memory"` (what OpenAI actually returns today):

```
[classify-task] Classification failed
  [classifyTask] {"description":"Expected \"in-memory\" | \"24h\", got \"in_memory\"
  at [\"prompt_cache_retention\"]","_tag":"InvalidOutputError",...}
```

## Test plan

- [x] `pnpm --filter @effect/ai-openai test OpenAiClient` — 24/24 passing (2 new tests added covering both `"in_memory"` and `"in-memory"` in a decoded `Response` body).
- [x] `pnpm check:tsgo` — passes.
- [x] `pnpm lint-fix` — clean.
- [x] Changeset added (`.changeset/fix-openai-prompt-cache-retention-in-memory.md`, patch bump for `@effect/ai-openai`).

### DIFF
--- a/.changeset/fix-openai-prompt-cache-retention-in-memory.md
+++ b/.changeset/fix-openai-prompt-cache-retention-in-memory.md
@@ -1,0 +1,7 @@
+---
+"@effect/ai-openai": patch
+---
+
+Accept `"in_memory"` (underscore) as a valid `prompt_cache_retention` value in generated response schemas.
+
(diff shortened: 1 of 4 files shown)


IMax153:
@matthew-sessions thank you for the PR but this work will be superceded by #1933 which includes the fix for the schemas you mentioned.

PR #2049, opened 2026-04-17
fix(rpc): add makeMsgPack for configurable msgpackr options

bohdanbirdie:
## Problem

`RpcSerialization.msgPack` silently fails on Cloudflare Workers when decoding messages with 3+ same-structure objects. The `Packr`/`Unpackr` are constructed with `useRecords: true`, which enables msgpackr's record path. When the JIT threshold is reached, msgpackr calls `new Function()` — blocked by CF Workers during request handling — and the `catch { return [] }` in the decode path silently swallows the resulting `EvalError`.

## Solution

Add `RpcSerialization.makeMsgPack(options)` factory accepting `msgpackr.Options`. CF Workers users pass `{ useRecords: false }`:

```ts
Layer.succeed(RpcSerialization)(
  RpcSerialization.makeMsgPack({ useRecords: false })
)
```

Existing `msgPack` / `layerMsgPack` exports are unchanged. Also rethrows non-incomplete decode errors instead of returning `[]`.

Same fix as v3: https://github.com/Effect-TS/effect/pull/6161
Reproduction: https://github.com/bohdanbirdie/repro-effect-rpc-msgpack-cf-workers

### DIFF
--- a/packages/effect/src/unstable/rpc/RpcSerialization.ts
+++ b/packages/effect/src/unstable/rpc/RpcSerialization.ts
@@ -400,47 +400,67 @@ interface JsonRpcResponse {
 type JsonRpcMessage = JsonRpcRequest | JsonRpcResponse
 
 /**
+ * Create a MessagePack serialization with custom msgpackr options.
+ *
+ * On Cloudflare Workers with `allow_eval_during_startup` (default for
+ * `compatibility_date >= 2025-06-01`), pass `{ useRecords: false }` to
+ * prevent msgpackr's JIT code generation via `new Function()`, which is
+ * blocked during request handling.
+ *
+ * @since 4.0.0
+ * @category serialization
+ * @example
+ * ```ts
+ * import { Layer } from "effect"
+ * import { RpcSerialization } from "effect/unstable/rpc"
+ *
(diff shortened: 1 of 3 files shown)


packages/effect/src/unstable/rpc/RpcSerialization.ts
```diff
@@ -400,47 +400,67 @@ interface JsonRpcResponse {
...
-          incomplete = undefined
-        }
-        try {
-          return unpackr.unpackMultiple(buf)
-        } catch (error_) {
-          const error = error_ as any
-          if (error.incomplete) {
-            incomplete = buf.subarray(error.lastPosition)
-            return error.values ?? []
+export const makeMsgPack = (options: Msgpackr.Options): RpcSerialization["Service"] =>
```
tim-smart:
```suggestion
export const makeMsgPack = (options?: Msgpackr.Options): RpcSerialization["Service"] =>
```

bohdanbirdie:
@tim-smart would it be possible to merge the same PR for v3? I applied the same changes you suggested here
https://github.com/Effect-TS/effect/pull/6161

PR #2062, opened 2026-04-20
fix http schemaBodyJson parseOptions annotation

aldotestino:
## Summary
- Fix `HttpIncomingMessage.schemaBodyJson` to annotate codec JSON schemas with `parseOptions` instead of `options`, so `ParseOptions` are correctly resolved by the schema parser.
- Add a regression test in `HttpServerRequest.test.ts` that verifies `onExcessProperty: \"preserve\"` keeps excess JSON keys when decoding via `schemaBodyJson`.
- Add a patch changeset for `effect` documenting the fix.

## Verification
- `pnpm lint-fix`
- `pnpm test packages/effect/test/unstable/http/HttpServerRequest.test.ts`
- `pnpm check:tsgo`
- `pnpm docgen` (from `packages/effect`)

### DIFF
--- a/.changeset/fix-http-incoming-message-parse-options.md
+++ b/.changeset/fix-http-incoming-message-parse-options.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `HttpIncomingMessage.schemaBodyJson` to forward parse options via the `parseOptions` annotation key.
--- a/packages/effect/src/unstable/http/HttpIncomingMessage.ts
+++ b/packages/effect/src/unstable/http/HttpIncomingMessage.ts
@@ -46,10 +46,11 @@ export interface HttpIncomingMessage<E = unknown> extends Inspectable.Inspectabl
  * @category schema
  */
 export const schemaBodyJson = <S extends Schema.Top>(schema: S, options?: ParseOptions | undefined) => {
-  const decode = Schema.decodeEffect(Schema.toCodecJson(schema).annotate({ parseOptions: options }))
+  const decode = Schema.decodeEffect(Schema.toCodecJson(schema))
   return <E>(
     self: HttpIncomingMessage<E>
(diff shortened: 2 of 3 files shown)


gcanti:
IMO we should avoid annotating the schema in the first place (both in `schemaBodyJson` and in `schemaBodyUrlParams`)

```ts
export const schemaBodyJson = <S extends Schema.Top>(schema: S, options?: ParseOptions | undefined) => {
  const decode = Schema.decodeEffect(Schema.toCodecJson(schema))
  return <E>(
    self: HttpIncomingMessage<E>
  ): Effect.Effect<S["Type"], E | Schema.SchemaError, S["DecodingServices"]> =>
    Effect.flatMap(self.json, (json) => decode(json, options))
}

//

export const schemaBodyUrlParams = <
  A,
  I extends Readonly<Record<string, string | ReadonlyArray<string> | undefined>>,
  RD,
  RE
>(
  schema: Schema.Codec<A, I, RD, RE>,
  options?: ParseOptions | undefined
) => {
  const decode = UrlParams.schemaRecord.pipe(
    Schema.decodeTo(schema),
    Schema.decodeEffect
  )
  return <E>(self: HttpIncomingMessage<E>): Effect.Effect<A, E | Schema.SchemaError, RD> =>
    Effect.flatMap(self.urlParamsBody, (urlParams) => decode(urlParams, options))
}
```

gcanti:
Btw why is it named "schemaBodyJson" instead of "decodeBodyJson"? A "schema*" prefixed API makes me think of a function that returns a schema

tim-smart (dismissed):

tim-smart:
> Btw why is it named "schemaBodyJson" instead of "decodeBodyJson"? A "schema*" prefixed API makes me think of a function that returns a schema

All the apis the integrate with schema are prefixed. decode / encode is too ambiguous imo.

PR #2069, opened 2026-04-21
fix: floor TestClock nanoseconds before BigInt conversion

mikearnaldi:
## Summary
- floor `TestClock.currentTimeNanosUnsafe()` before converting fractional millisecond instants to `BigInt`
- add a regression test covering a fractional `setTime(...)` timestamp

Port of: https://github.com/Effect-TS/effect/pull/6194

### DIFF
--- a/packages/effect/test/TestClock.test.ts
+++ b/packages/effect/test/TestClock.test.ts
@@ -53,4 +53,11 @@ describe("TestClock", () => {
       yield* TestClock.setTime(Duration.toMillis(Duration.hours(11)))
       assert.isTrue(elapsed)
     }))
+
+  it.effect("setTime - floors nanoseconds for fractional millisecond instants", () =>
+    Effect.gen(function*() {
+      const testClock = yield* TestClock.make()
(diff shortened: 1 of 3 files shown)


packages/effect/test/TestClock.test.ts
```diff
@@ -53,4 +53,11 @@ describe("TestClock", () => {
       yield* TestClock.setTime(Duration.toMillis(Duration.hours(11)))
       assert.isTrue(elapsed)
     }))
+
+  it.effect("setTime - floors nanoseconds for fractional millisecond instants", () =>
+    Effect.gen(function*() {
+      const testClock = yield* TestClock.make()
+      yield* testClock.setTime(34199023438.000004)
+      assert.strictEqual(testClock.currentTimeNanosUnsafe(), 34199023438000004n)
```
tim-smart:
```suggestion
      yield* testClock.setTime(34199023438.0000004)
      assert.strictEqual(testClock.currentTimeNanosUnsafe(), 34199023438000000n)
```

Need to test for the out-of-bounds case, otherwise flooring will have no effect.

PR #2080, opened 2026-04-22
chore: update tstyche to v7.1.0

mrazauskas:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR updates `tstyche` to v7.1.0. This release removes the experimental `when()` helper, which was undocumented and unfinished.

I initially hoped it would help avoid `// @ts-expect-error` in complex cases, but it actually made those test cases harder to read. Using `// @ts-expect-error` is sufficient and much clearer for readers.

As you'll notice in this PR, `.not.toBeCallableWith()` can be used instead in some cases, which is another reason to remove `when()` as redundant.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "rollup-plugin-visualizer": "^7.0.1",
     "terser": "^5.46.1",
     "ts-patch": "^3.3.0",
-    "tstyche": "^7.0.0",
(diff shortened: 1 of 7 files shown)


gcanti:
yeah, it's much nicer indeed, thanks

PR #2095, opened 2026-04-27
fix node request body caching

### LINKED ISSUE #2096: From Discord: HttpServerRequest body stream cannot be read multiple times in HttpRouter route handlers
# Summary
Users observed that calling both `request.arrayBuffer` and `request.text` sequentially on the same `HttpServerRequest` in an HttpRouter route leads to the server hanging and eventually timing out with an InterruptError. This happens because reading the request body stream consumes it, preventing subsequent reads. While HttpClientRequest supports cached reads allowing multiple accesses to the body, HttpServerRequest in the router context does not cache the stream, causing the problem. A minimal reproduction scenario demonstrates how attempting to read the body twice within a single POST route handler blocks the request, unlike when the same is done outside the router. The issue appears specific to HttpRouter’s request stream handling, suggesting the need for caching or mechanisms to allow multiple reads of the request body in route handlers.

# Discord thread

https://discord.com/channels/795981131316985866/1498330174071115806

IMax153:
## Summary
- cache Node HTTP server request bytes once and derive `text` and `arrayBuffer` from the same body
- add a regression test covering repeated reads of the same request body
- include a patch changeset for `@effect/platform-node`

Closes #2096

### DIFF
--- a/.changeset/fix-node-request-body-caching.md
+++ b/.changeset/fix-node-request-body-caching.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-node": patch
+---
+
+Fix Node HTTP server request body caching so `text`, `json`, and `arrayBuffer` can all be read from the same request body without consuming the stream more than once.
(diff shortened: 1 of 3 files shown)


tim-smart:
Closed by #2098

PR #2097, opened 2026-04-27
feat(reactivity): add `.exhaustive()` finalizer to AsyncResult builder

Leka74:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds an `exhaustive()` finalizer to `AsyncResult.builder`.

Makes it easier & safer to handle all cases, instead of relying on `.render()` which can throw for unhandled failures.

## Related

- Related Issue: N/A
- Closes: N/A

### DIFF
--- a/packages/effect/src/unstable/reactivity/AsyncResult.ts
+++ b/packages/effect/src/unstable/reactivity/AsyncResult.ts
@@ -612,57 +612,59 @@ export const builder = <A extends AsyncResult<any, any>>(self: A): Builder<
   never,
   A extends Success<infer _A, infer _E> ? _A : never,
   A extends Failure<infer _A, infer _E> ? _E : never,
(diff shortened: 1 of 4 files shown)


packages/effect/src/unstable/reactivity/AsyncResult.ts
```diff
@@ -628,6 +628,10 @@ export type Builder<Out, A, E, I> =
     orNull(): Out | null
     render(): [A | I] extends [never] ? Out : Out | null
   }
+  & ([A | E | I] extends [never] ? {
+      exhaustive(): Out
```
tim-smart, line 648: It isn't truly exhaustive so I think it is a bit misleading.

PR #2100, opened 2026-04-27
Change the current it.effect.each to it.effect.for

jvliwanag:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

On vanilla non-effect Vitest, there are two ways to provide a list of test cases:

* [`it.each`](https://vitest.dev/api/test.html#test-each) which treat each input array entry as var args
* [`it.for`](https://vitest.dev/api/test.html#test-for) which directly assigns each input array entry as the first arg, leaving room for a second arg - the vitest test context

Currently, the behavior of `it.effect.each` behaves more like vanilla `it.for` in that it's not var arg, but currently does not pass  the test context.

The pull request changes the ff:

* Add the vitest test context on the second arg
* Rename `it.effect.each` to `it.effect.for` so it aligns more with vitest naming.

## Related

https://github.com/Effect-TS/effect/issues/5839 notes that the current `it.effect.each` is undocumented still.

### DIFF
--- a/.changeset/jolly-grapes-mate.md
+++ b/.changeset/jolly-grapes-mate.md
@@ -1,0 +1,5 @@
+---
+"@effect/vitest": minor
+---
+
+Rename it.effect.each to it.effect.for for Vitest alignment and add support for accessing the test context.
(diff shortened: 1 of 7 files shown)


IMax153:
Hi @jvliwanag,

PR #2107, opened 2026-04-29
expose otel export flush mechanism

harrysolovay:
## Summary

- Add `OtlpExporter.Exporters` and `OtlpExporter.layerExporters` to register OTLP exporters and flush them explicitly.
- Allow OTLP logger/tracer/exporter configuration to use `exportInterval: "disabled"` for manual flush mode.
- Make OTLP exports semaphore-protected and fully consume HTTP response bodies after export requests.
- Lazily initialize the HTTP client response finalization registry to avoid side effects in workerd (see https://gist.github.com/danieljvdm/726f9dcd8bc808eac5b82a502329e9e6).
- Add test coverage for manually flushing disabled OTLP exporters.

## Rationale

Managed runtimes in environments such as workerd cannot reliably depend on background interval fibers to export telemetry. For example, a Durable Object may go into hibernation before the next scheduled OTLP export interval runs.

This change introduces a manual flush model for those environments:

- telemetry is buffered during request handling
- scheduled interval exports can be disabled
- the runtime integration can call `OtlpExporter.Exporters.flush`
- and platforms such as Cloudflare can pass that flush effect to `ctx.waitUntil(...)` so logs/spans are exported after the response is returned but before the runtime is allowed to shut down the computation.

This makes OTLP exporting compatible with request-scoped and hibernation-prone runtimes without forcing exports to happen inline on the critical response path.

## Notes

`exportInterval: "disabled"` should be understood as manual flush mode, not as disabling telemetry export entirely. Consumers using this mode should provide `OtlpExporter.layerExporters` and explicitly call `flush` from their platform lifecycle hook.

### DIFF
--- a/.changeset/fuzzy-hounds-flush.md
+++ b/.changeset/fuzzy-hounds-flush.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Allow OTLP metrics exporters to disable interval exports and rely on manual flushing.
--- a/.changeset/otlp-exporter-flush.md
+++ b/.changeset/otlp-exporter-flush.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add shared OTLP exporter flushing support and allow disabling scheduled exports.
--- a/packages/effect/src/unstable/http/HttpClient.ts
+++ b/packages/effect/src/unstable/http/HttpClient.ts
@@ -1368,7 +1368,23 @@ export const layerMergedContext = <E, R>(
 // internal
 // -----------------------------------------------------------------------------
 
-const responseRegistry = (() => {
+interface ResponseRegistry {
+  readonly register: (response: HttpClientResponse.HttpClientResponse, controller: AbortController) => void
+  readonly unregister: (response: HttpClientResponse.HttpClientResponse) => void
+}
+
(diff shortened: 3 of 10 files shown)


patroza:
Otel auto flushes on layer scope closing. Is there no way for you to call dispose on the managedruntime at these events?

harrysolovay:
Calling dispose per Worker request would use the wrong lifecycle. It would close the whole managed runtime, forcing request-agnostic layers like config, HttpClient, Drizzle, and OTLP exporters to be rebuilt on every request. We want those services to live for the isolate / Durable Object instance.

The issue is that OTLP finalizers are tied to that long-lived layer scope. They run when the runtime is disposed, which is useful for Node-style process shutdown, but not for Worker request completion. A Worker event can finish, freeze, or terminate without the runtime being disposed.

OtlpExporter.layerExporters registers long-lived exporters, and OtlpExporter.flush lets the Worker adapter drain them without disposing the runtime. That can be paired with ctx.waitUntil so Cloudflare keeps the event alive long enough to send logs/traces.

In short: dispose would flush, but by shutting down the whole app runtime. Workers need per-event flushing without tearing down shared services.

IMax153:
Hi @harrysolovay,

Issue #2105, opened 2026-04-28
EntityProxyServer.layerRpcHandlers crashes entity proxy RPC calls because handler entries use `services` instead of `context`


jan-wilhelm:
### What version of Effect is running?

4.0.0-beta.56

### What steps can reproduce the bug?

Using `effect@4.0.0-beta.56`, create an entity, expose it through `EntityProxy.toRpcGroup`, provide handlers with `EntityProxyServer.layerRpcHandlers`, and call the generated RPC client.

A runnable reproduction repo is available at: https://github.com/jan-wilhelm/effect-entity-proxy-context-repro

Minimal reproduction:

```js
import { Effect } from "effect";
import { Entity, EntityProxy, EntityProxyServer, Sharding } from "effect/unstable/cluster";
import { Rpc, RpcClient, RpcServer } from "effect/unstable/rpc";

const TestEntity = Entity.make("TestEntity", [Rpc.make("NoPayload")]);
const TestEntityRpcs = EntityProxy.toRpcGroup(TestEntity);

let server;

const program = Effect.gen(function* () {
  const sharding = {
    makeClient: () =>
      Effect.succeed((entityId) => ({
        NoPayload: (payload, options) =>
          Effect.log(
            `entity ${entityId} payload:${String(payload)} discard:${String(options?.discard)}`
          ),
      })),
    isShutdown: Effect.succeed(false),
    pollStorage: Effect.void,
  };

  const clientSide = yield* RpcClient.makeNoSerialization(TestEntityRpcs, {
    supportsAck: true,
    onFromClient: ({ message }) => server.write(0, message),
  });

  server = yield* RpcServer.makeNoSerialization(TestEntityRpcs, {
    onFromServer: (response) => clientSide.write(response),
  }).pipe(
    Effect.provide(EntityProxyServer.layerRpcHandlers(TestEntity)),
    Effect.provideService(Sharding.Sharding, sharding)
  );

  yield* clientSide.client["TestEntity.NoPayloadDiscard"]({
    entityId: "id",
    payload: undefined,
  });
});

Effect.runPromise(Effect.scoped(program)).then(
  () => console.log("ok"),
  (error) => {
    console.error(error);
    process.exit(1);
  }
);
```

For example:

```sh
bun add effect@4.0.0-beta.56
bun repro.mjs
```

### What is the expected behavior?

The generated entity proxy RPC should call the entity client handler and complete successfully. In the repro above, it should log something like:

```txt
[...]

PR #2111, opened 2026-04-30
fix(cluster): EntityProxyServer.layerHttpApi uses params instead of path

thiagofelix:
## Type

   - [ ] Refactor
   - [ ] Feature
   - [x] Bug Fix
   - [ ] Optimization
   - [ ] Documentation Update

   ## Description

   `EntityProxyServer.layerHttpApi` destructured `path.entityId` from the HttpApiEndpoint handler request, but `HttpApiBuilder` provides parsed path parameters under the `params`
 key (matching `HttpApiEndpoint.Request` type).

   This caused `TypeError: undefined is not an object (evaluating 'path.entityId')` on every entity HTTP API call (both sync and discard variants).

   ### Fix

   Replace `path` → `params` in all 6 occurrences within `layerHttpApi`.

   ### Regression Test

   Added `"layerHttpApi correctly passes path params to entity client"` test in `packages/platform-node/test/RpcServer.test.ts` — spins up a real HTTP server with
 `EntityProxy.toHttpApiGroup`, POSTs to the entity endpoint, and asserts the `entityId` path parameter is correctly forwarded to the entity client.

   ### Notes

   - `layerRpcHandlers` is unaffected (uses `{ entityId, payload }` directly)
   - Related: #2105 fixed a separate `services` vs `context` bug in `layerRpcHandlers`

   ## Related

   - Related Issue #433

### DIFF
--- a/.changeset/fix-entity-proxy-server-path-params.md
+++ b/.changeset/fix-entity-proxy-server-path-params.md
@@ -3,10 +3,3 @@
 ---
 
 Fix `EntityProxyServer.layerHttpApi` using `path.entityId` instead of `params.entityId`
-
-`HttpApiEndpoint.Request` provides parsed path parameters under the `params` key,
-but `EntityProxyServer.layerHttpApi` was destructuring `path` — causing
-`TypeError: undefined is not an object (evaluating 'path.entityId')` on every
-entity HTTP call.
-
-This bug has existed since `EntityProxyServer.layerHttpApi` was introduced in #433.
--- a/.changeset/fix-entity-proxy-server-path-params.md
+++ b/.changeset/fix-entity-proxy-server-path-params.md
@@ -1,0 +1,12 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/fix-entity-proxy-server-path-params.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Fix `EntityProxyServer.layerHttpApi` using `path.entityId` instead of `params.entityId`
+
+`HttpApiEndpoint.Request` provides parsed path parameters under the `params` key,
+but `EntityProxyServer.layerHttpApi` was destructuring `path` — causing
+`TypeError: undefined is not an object (evaluating 'path.entityId')` on every
+entity HTTP call.
+
+This bug has existed since `EntityProxyServer.layerHttpApi` was introduced in #433.
```
tim-smart:
```suggestion
Fix `EntityProxyServer.layerHttpApi` using `path.entityId` instead of `params.entityId`
```

tim-smart (changes requested):
There are a couple issues with the test from CI

tim-smart (approved):

PR #2113, opened 2026-05-01
feat(Schema): allow defaults to fail with SchemaError

patroza:
Allow Schema constructor and decoding defaults to fail with `SchemaError`.

The `Effect` passed to `Schema.withConstructorDefault`, `Schema.withDecodingDefault`, and `Schema.withDecodingDefaultKey` now accepts `SchemaError` in its error channel. When a default fails, the parser unwraps the underlying `SchemaIssue.Issue` and propagates it as a parse failure with the surrounding path attached. This makes it easy to use another schema's `makeEffect` / `decode*` as the default value.


packages/effect/src/Schema.ts
```diff
@@ -4300,10 +4300,10 @@ export interface withConstructorDefault<S extends Top & WithoutConstructorDefaul
 export function withConstructorDefault<S extends Top & WithoutConstructorDefault>(
   // `S["~type.make.in"]` instead of `S["Type"]` is intentional here because
   // it makes easier to define the default value if there are nested defaults
-  defaultValue: Effect.Effect<S["~type.make.in"]>
+  defaultValue: Effect.Effect<S["~type.make.in"], SchemaError>
```
gcanti, line 4309: is using `Issue.Issue` instead of `SchemaError` more elegant and consistent?

gcanti (changes requested):

packages/effect/src/Schema.ts
patroza: > is using `Issue.Issue` instead of `SchemaError` more elegant and consistent?

At first I was considering it, but then all apis in Schema operate on SchemaError, only AST & Friends deal with SchemaIssue.

The main use case I have is `S.String.pipe(S.withConstructorDefault(SomethingElse.makeEffect()))` - which returns `SchemaError`.
Once you need to add a `.pipe(Effect.mapError(_ => _.issue)))` it's starting to loose it's benefits.

One consideration could be to support both, but I think being aligned in the Schema module on SchemaError seems to make sense.

gcanti: ok, then I'd use `Effect.mapErrorEager` instead of `Effect.mapError`

gcanti:
@patroza What about `withDecodingDefaultType` and `withDecodingDefaultTypeKey`? Did you leave them unchanged intentionally, or was it an oversight?

patroza:
@gcanti oversight sadly; there are so many! Fixed :)

gcanti (approved):

patroza:
@gcanti thanks! added a sibling :) https://github.com/Effect-TS/effect-smol/pull/2176

PR #2127, opened 2026-05-06
fix(Stream): avoid mutable Channel.runFold accumulator in mkUint8Array

### LINKED ISSUE #2126: Stream.mkUint8Array crashes in Bun compiled+minified binaries: mutable Channel.runFold accumulator
## Bug

`Stream.mkUint8Array` crashes with `Attempting to define property on object that is not extensible` when used inside a Bun `--compile --minify` binary.

## Root cause

In `4.0.0-beta.59`, `mkUint8Array` was refactored to use `Channel.runFold` with a **mutable accumulator**:

```ts
Channel.runFold(
  self.channel,
  () => ({ bytes: 0, arrays: [] }),
  (acc, chunk) => {
    acc.bytes += chunk[i].length  // mutates acc
    acc.arrays.push(chunk[i])     // mutates acc
    return acc
  }
)
```

In Bun compiled **AND** minified binaries, the accumulator object becomes non-extensible between fold iterations. The second iteration receives a frozen `acc`, so `acc.bytes +=` and `acc.arrays.push` throw. The exact JSC mechanism is not confirmed yet, but the behavior is reliably reproducible. *Took me a good 12 hours to trace down (and probably a couple of million tokens)*

Before `4.0.0-beta.59`, `mkUint8Array` returned a **new** `Uint8Array` on each fold iteration (immutable pattern), matching the convention used in `Multipart.collectUint8Array`. The beta.59 refactor introduced in-place mutation for performance, which breaks in Bun compiled mode.

This does NOT reproduce with:

- `bun run` (interpreted mode)
- `bun build --compile` without `--minify`
- Any Node.js runtime

If you compile without minification the issue does not happen. I still need to check out exactly why this behavior is such an issue for the minifier and how to best raise this with the [Bun team](https://github.com/oven-sh/bun?tab=contributing-ov-file). *But fixing this issue here would help the [`opencode` team](https://github.com/anomalyco/opencode) massively already.*

## Impact

All callers of `Stream.mkUint8Array` in compiled Bun binaries crash after the first chunk. The crash surfaces as a defect (unhandled exception), not a typed error.

Discovered via `opencode` issue [#25873](https://github.com/anomalyco/opencode/issues/25873) because those maniacs _(in the best possible sense)_ live on the edge and were jumping from [4.0.0-beta.57](https://www.npmjs.com/package/effect/v/4.0.0-beta.57) to [4.0.0-beta.59](https://www.npmjs.com/package/effect/v/4.0.0-beta.59).

opencode's `snapshot/index.ts` calls `Stream.mkUint8Array` to collect `git cat-file --batch` output, causing **every second** (and following) LLM tool call in a session to fail with `Attempted to assign to readonly property`. 
_You gotta love those "only manifest in every not-the-first calls" issue..._

Anyways, `opencode` makes heavy use of `bun` and does minify their binaries and uses the beta version of this package.

## Same mutable-accumulator pattern elsewhere

`Stream.runCollect` ([Stream.ts line 9906](https://github.com/Effect-TS/effect-smol/blob/aae8797b9cb383be0c182dd58d03d787c354238b/packages/effect/src/Stream.ts#L9906)) and `Channel.runCollect` ([Channel.ts line 7221](https://github.com/Effect-TS/effect-smol/blob/aae8797b9cb383be0c182dd58d03d787c354238b/packages/effect/src/Channel.ts#L7221)) use the same mutable fold pattern (`acc.push(...); return acc`) and may be affected in Bun compiled binaries.

Note: `Multipart.collectUint8Array` is NOT affected because it already uses the immutable pattern (returns a new `Uint8Array` each iteration).

## Fix

_At first I tried with a `runForEach` + `Effect.suspend` with local variables to avoid the mutable fold accumulator but that's way slower because of repeated copies._ Reverted to the pre-beta.59 immutable `Channel.runFold` pattern where each fold iteration returns a new `Uint8Array`. See PR #2127.

I checked if there are related issues or PRs but none found.

stephanschielke:
Fixes #2126

## What

`Stream.mkUint8Array` was refactored in `4.0.0-beta.59` to use `Channel.runFold` with a mutable accumulator (`acc.bytes +=`, `acc.arrays.push`). In Bun `--compile --minify` binaries, mutating an object that is passed repeatedly as the accumulator to a fold step callback throws `Attempting to define property on object that is not extensible`.

This does NOT reproduce with:
- `bun run` (interpreted mode)
- `bun build --compile` without `--minify`
- Any Node.js runtime

## Fix

Revert to the pre-beta.59 immutable `Channel.runFold` pattern where each fold iteration returns a **new** `Uint8Array` instead of mutating a shared `{ bytes, arrays }` object. This matches the established convention used by `Multipart.collectUint8Array` (line 472) and what `mkUint8Array` itself used before beta.59.

## Verification

All commands run from the repo root:

- `pnpm --filter effect check` -- zero type errors
- `pnpm lint-fix` -- zero warnings, zero errors
- `pnpm --filter effect test run -- Stream.test.ts` -- all Stream tests pass. 7 pre-existing failures in unrelated files (EffectKeepAlive, Command, OtlpMetrics) -- same count on upstream/main without this change.

## Tests

Added three tests for `mkUint8Array`:
- basic concatenation across two chunks
- many-chunk regression (10 single-byte chunks -- the exact failing scenario)
- empty stream edge case

## Real-world impact

opencode v1.14.34+ crashes on every second LLM tool call in compiled binaries because `snapshot/index.ts` calls `Stream.mkUint8Array(handle.stdout)` to collect `git cat-file --batch` output. See anomalyco/opencode#25873 and anomalyco/opencode#25867.

## Note

`Stream.runCollect` (line 9907) also mutates its fold accumulator (`acc.push(chunk[i])`). It may be affected by the same Bun compiled-mode behavior. Not fixing here to keep this PR focused on the confirmed crash site.

### DIFF
--- a/.changeset/fix-stream-mkuint8array-mutable-fold.md
+++ b/.changeset/fix-stream-mkuint8array-mutable-fold.md
@@ -1,0 +1,7 @@
+---
+"effect": patch
+---
+
+Fix `Stream.mkUint8Array` crash in Bun compiled+minified binaries.
+
+Reverted to the pre-beta.59 immutable `Channel.runFold` pattern where each fold iteration returns a new `Uint8Array` instead of mutating a shared accumulator. The beta.59 refactor introduced in-place mutation (`acc.bytes +=`, `acc.arrays.push`) which throws in Bun `--compile --minify` binaries.
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -10339,31 +10339,20 @@ export const mkString = <E, R>(self: Stream<string, E, R>): Effect.Effect<string
  * @category Destructors
  */
 export const mkUint8Array = <E, R>(self: Stream<Uint8Array, E, R>): Effect.Effect<Uint8Array, E, R> =>
-  Effect.map(
-    Channel.runFold(
-      self.channel,
-      (): {
-        bytes: number
-        readonly arrays: Array<Uint8Array>
-      } => ({
-        bytes: 0,
-        arrays: []
-      }),
-      (acc, chunk) => {
-        for (let i = 0; i < chunk.length; i++) {
-          acc.bytes += chunk[i].length
-          acc.arrays.push(chunk[i])
-        }
-        return acc
+  Channel.runFold(
+    self.channel,
+    () => new Uint8Array(0),
+    (acc, chunk) => {
+      let chunkLength = 0
+      for (let i = 0; i < chunk.length; i++) {
+        chunkLength += chunk[i].length
       }
-    ),
-    ({ arrays, bytes }) => {
-      const result = new Uint8Array(bytes)
-      let offset = 0
-      for (let i = 0; i < arrays.length; i++) {
-        const array = arrays[i]
-        result.set(array, offset)
-        offset += array.length
+      const result = new Uint8Array(acc.length + chunkLength)
+      result.set(acc, 0)
+      let offset = acc.length
+      for (let i = 0; i < chunk.length; i++) {
+        result.set(chunk[i], offset)
+        offset += chunk[i].length
       }
       return result
     }
--- a/packages/effect/test/Stream.test.ts
+++ b/packages/effect/test/Stream.test.ts
@@ -205,6 +205,29 @@ describe("Stream", () => {
       }))
   })
 
+  describe("mkUint8Array", () => {
+    it.effect("concatenates chunks into a single Uint8Array", () =>
+      Effect.gen(function*() {
+        const result = yield* Stream.mkUint8Array(
+          Stream.make(new Uint8Array([1, 2]), new Uint8Array([3, 4]))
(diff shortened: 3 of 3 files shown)


stephanschielke:
Note on CI: there are 7 pre-existing test failures on `main` that are unrelated to this change:

- `test/EffectKeepAlive.test.ts` (2 failures -- timing assertions)
- `test/unstable/cli/Command.test.ts` (4 failures -- `toMatchInlineSnapshot` context error)
- `test/unstable/observability/OtlpMetrics.test.ts` (1 failure -- gauge delta assertion)

Verified locally by running the test suite on upstream/main without this change -- same 7 failures. With this change, the count stays at 7 (zero regressions introduced).

IMax153:
Closing this PR as the issue is not on our end.

PR #2133, opened 2026-05-06
fix(ai-openai): don't spread library-only config fields into request body

### LINKED ISSUE #2125: [@effect/ai-openai] OpenAiLanguageModel leaks local-only Config fields (`fileIdPrefixes`, `strictJsonSchema`) into request body, causing OpenAI 400
## Summary

`OpenAiLanguageModel.makeRequest` spreads the entire `Config.Service` object into the OpenAI request body. The `Config.Service` shape includes local-only library fields (`fileIdPrefixes`, `strictJsonSchema`) alongside real OpenAI API parameters. Those local fields get JSON-stringified into the `/responses` request and OpenAI rejects them with `400 Bad Request: Unknown parameter: 'fileIdPrefixes'`.

## Repro

```ts
import { OpenAiClient, OpenAiLanguageModel } from "@effect/ai-openai"
import { LanguageModel, Prompt } from "effect/unstable/ai"
import { Config, Effect, Layer } from "effect"
import { FetchHttpClient } from "effect/unstable/http"

const layer = OpenAiLanguageModel.layer({
  config: { fileIdPrefixes: ["file-"] },  // <-- needed so file_id strings dispatch to `input_file`
  model: "gpt-4o-mini",
}).pipe(
  Layer.provide(OpenAiClient.layerConfig({ apiKey: Config.redacted("OPENAI_API_KEY") })),
  Layer.provide(FetchHttpClient.layer),
)

  const result = yield* LanguageModel.generateText({
    prompt: Prompt.make([
      { role: "user", content: [
        { type: "text", text: "Summarize this PDF" },
        Prompt.filePart({ data: "file-abc123", mediaType: "application/pdf" }),
      ]},
    ]),
  })
  console.log(result.text)
})

Effect.runPromise(program.pipe(Effect.provide(layer)))
// → AiError: Invalid request. Unknown parameter: 'fileIdPrefixes'.
//   (POST https://api.openai.com/v1/responses) [code: unknown_parameter]
```

## Root cause

`packages/ai/openai/src/OpenAiLanguageModel.ts` line 379 (HEAD as of 2026-05):

```ts
const request: Mutable<typeof OpenAiSchema.CreateResponse.Encoded> = {
  ...config,        // ← spreads `fileIdPrefixes`, `strictJsonSchema`, etc. into the request body
  input: messages,
  include: include.size > 0 ? Array.from(include) : undefined,
  text: {
    verbosity: config.text?.verbosity ?? undefined,
    format: responseFormat
  }
}
```

`Config.Service` is declared (line 64-103) as:

```ts
Partial<Omit<typeof OpenAiSchema.CreateResponse.Encoded, "input" | "tools" | "tool_choice" | "stream" | "text">>
& {
  readonly fileIdPrefixes?: ReadonlyArray<string> | undefined
  readonly text?: { readonly verbosity?: "low" | "medium" | "high" | undefined } | undefined
  readonly strictJsonSchema?: boolean | undefined
}
```

So `Config.Service` mixes real API params with three library-internal fields (`fileIdPrefixes`, `strictJsonSchema`, and a stripped-down `text`). The spread doesn't filter — both groups go into the JSON body. The `Mutable<typeof OpenAiSchema.CreateResponse.Encoded>` cast on line 379 doesn't catch this because TypeScript spread compatibility is too loose.

The body is then sent via `OpenAiClient.createResponse` → `HttpClientRequest.post("/responses", { body: HttpBody.jsonUnsafe(payload) })` (`packages/ai/openai/src/OpenAiClient.ts:203-204`), which is plain `JSON.stringify` — no schema validation/filtering.

## Suggested fix

Filter local-only keys before the spread:

```ts
const { fileIdPrefixes: _fip, strictJsonSchema: _sjs, ...apiConfig } = config
const request: Mutable<typeof OpenAiSchema.CreateResponse.Encoded> = {
  ...apiConfig,
  input: messages,
  // ...
}
```

Or better — encode the request via `OpenAiSchema.CreateResponse` so unknown fields are stripped at the schema boundary.

## Workaround we're using

Custom `transformClient` on `OpenAiClient.layerConfig` that intercepts outbound `/responses` requests, parses the JSON body, and deletes `fileIdPrefixes` / `strictJsonSchema` before send.

## Versions

- `@effect/ai-openai@4.0.0-beta.60`
- `effect@4.0.0-beta.60`
- Bun 1.3.13

Zelys-DFKH:
Closes #2125

## Summary

- `Config.Service` holds two library-internal fields (`fileIdPrefixes`, `strictJsonSchema`) that have no meaning on the OpenAI API. `makeRequest` spreads the whole config into the request body, so both fields land in the JSON and OpenAI rejects the call with `400 Unknown parameter`.
- Destructure those two fields out before spreading, keeping all real API params intact. Same fix in `@effect/ai-openai-compat` where the fields would pass through `extractCustomRequestProperties` to the chat completions endpoint.
- Added a regression test to each package: pass `fileIdPrefixes`, `strictJsonSchema`, and `temperature` in the model config, run `generateText`, inspect the captured body. Library-only fields absent, `temperature` present. Both tests fail without the fix.

## Checks

- `pnpm exec dprint fmt` on changed files
- `pnpm --filter "@effect/ai-openai" test`
- `pnpm --filter "@effect/ai-openai-compat" test`
- `pnpm check:tsgo`

### DIFF
--- a/.changeset/ai-openai-config-field-leak.md
+++ b/.changeset/ai-openai-config-field-leak.md
@@ -1,0 +1,6 @@
+---
+"@effect/ai-openai": patch
+"@effect/ai-openai-compat": patch
+---
+
+Fix `OpenAiLanguageModel` leaking library-only config fields (`fileIdPrefixes`, `strictJsonSchema`) into request body, causing OpenAI 400 errors.
--- a/packages/ai/openai-compat/src/OpenAiLanguageModel.ts
+++ b/packages/ai/openai-compat/src/OpenAiLanguageModel.ts
@@ -372,8 +372,9 @@ export const make = Effect.fnUntraced(function*({ model, config: providerConfig 
         config,
         options
       })
+      const { fileIdPrefixes: _fip, strictJsonSchema: _sjs, ...apiConfig } = config
       const request: CreateResponse = {
-        ...config,
+        ...apiConfig,
         input: messages,
         include: include.size > 0 ? Array.from(include) : null,
         text: {
--- a/packages/ai/openai-compat/test/OpenAiLanguageModel.test.ts
+++ b/packages/ai/openai-compat/test/OpenAiLanguageModel.test.ts
@@ -1027,6 +1027,40 @@ describe("OpenAiLanguageModel", () => {
         assert.isTrue(capturedRequest.url.endsWith("/chat/completions"))
       }))
   })
+
+  describe("config", () => {
+    it.effect("does not leak library-only fields into request body", () =>
+      Effect.gen(function*() {
+        let capturedRequest: HttpClientRequest.HttpClientRequest | undefined
+
+        const layer = OpenAiClient.layer({ apiKey: Redacted.make("sk-test-key") }).pipe(
+          Layer.provide(Layer.succeed(
+            HttpClient.HttpClient,
+            makeHttpClient((request) => {
+              capturedRequest = request
+              return Effect.succeed(jsonResponse(request, makeChatCompletion()))
+            })
+          ))
+        )
+
+        yield* LanguageModel.generateText({ prompt: "test" }).pipe(
+          Effect.provide(OpenAiLanguageModel.model("gpt-4o-mini", {
+            fileIdPrefixes: ["file-"],
+            strictJsonSchema: false,
+            temperature: 0.5
+          })),
+          Effect.provide(layer)
+        )
+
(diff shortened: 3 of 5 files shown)


tim-smart (approved):

Zelys-DFKH:
Thanks for the quick review and merge!

PR #2144, opened 2026-05-09
fix(effect): preserve Vary: Origin in CORS preflight when Access-Control-Request-Headers is present

kitlangton:
## Summary

`HttpMiddleware.cors` was silently dropping `Vary: Origin` from preflight responses whenever the request also had `Access-Control-Request-Headers` and `allowedOrigins` was a predicate or a multi-entry list.

`headersFromRequestOptions` builds the response headers via object spread:

```ts
return Headers.fromRecordUnsafe({
  ...allowOrigin(origin),                          // { vary: "Origin", ... }
  ...allowCredentials,
  ...exposeHeaders,
  ...allowMethods,
  ...allowHeaders(accessControlRequestHeaders),    // { vary: "Access-Control-Request-Headers", ... }
  ...maxAge
})
```

Because both helpers set the same lowercase `vary` key, the second spread wins and the response goes out with `Vary: Access-Control-Request-Headers` only. With dynamic origin echoing, that means a shared cache can serve a preflight cached for one origin to a request from a different origin.

## Fix

- Stop returning `vary` from `allowOrigin` / `allowHeaders`.
- Compute the `Vary` header explicitly in `headersFromRequest` / `headersFromRequestOptions` from a list of contributing entries, joining with `, `.

The non-preflight path was correct (only `allowOrigin` contributes to Vary), but it now goes through the same explicit computation for consistency.

## Tests

- New reproducer test that fails on `main` and passes with the fix.
- Test that confirms `Vary: Origin` (and only `Origin`) when `allowedHeaders` is configured statically.
- Test that confirms no `Vary` header for the wildcard origin case.

## Test plan

- [x] `pnpm vitest run test/unstable/http/HttpMiddleware.test.ts` (4/4 passing)
- [x] `pnpm check:tsgo`
- [x] `pnpm lint-fix`
- [x] `pnpm docgen` (effect package)
- [x] Changeset added (`.changeset/fix-cors-vary-merge.md`, patch)


IMax153:
Hi @kitlangton,

PR #2145, opened 2026-05-09
fix(ai-openai): allow null for fields OpenAI returns null for

lindit:
## Summary

OpenAI returns `null` for several response fields the OpenAPI spec marks as optional-only, causing `HttpClientResponse.schemaBodyJson` to fail decoding with:

```
SchemaError(Expected <type>, got null at [<field>])
```

This PR adds JSON Patch entries to `packages/ai/openai/codegen.yaml` marking each field nullable (matching the precedent already set for `ModelResponseProperties.user`, `.safety_identifier`, `.prompt_cache_key`, and `Response.usage`):

- **`OpenAIFile.status_details`** — Files API (`createFile` / `retrieveFile` / `listFiles`). The fix flows through `Upload.file`, `ListFiles200`, `CreateFile200`, and `RetrieveFile200`.
- **`Batch.model`** — `null` on freshly created batches that haven't been validated yet.
- **`Batch.output_file_id`** — `null` until the batch completes successfully.
- **`Batch.error_file_id`** — `null` unless errors occurred.

After the patch, each field is emitted as `Schema.optionalKey(Schema.Union([Schema.<T>, Schema.Null]))`.

## Note on diff size

`Generated.ts` was last regenerated on 2026-04-22, so this regeneration also picks up ~3 weeks of unrelated upstream OpenAI spec drift (new `AdminApiKeyCreateResponse` type, tightened object literals, `int64` → `unixtime` format swaps, etc.). This is in line with prior regen commits (e.g. `c5a73274` was 12,529 / 9,373; this one is 1,896 / 678). Happy to split the regen into a separate prep PR if preferred.

## Note on `Batch.errors` (generator bug, not patched)

While sweeping `Batch`, OpenAI's own spec example shows `"errors": null`. I tried adding `nullable: true` for `Batch.errors` too, but the generator handles `nullable: true` on **object-typed** properties incorrectly — it produces a degenerate `Schema.Union([Schema.Struct(...)])` with no `Schema.Null` member, and the corresponding TS type is unchanged. Primitive-typed properties (the four in this PR) work correctly. Worth filing as a separate generator bug; the workaround is to express it via `anyOf: [{...object schema...}, {type: 'null'}]` instead of `nullable: true`, which is verbose to encode as a JSON Patch.

## Test plan

- [x] `pnpm --filter @effect/ai-openai check` passes
- [x] `pnpm --filter @effect/ai-openai test` passes (90 tests)
- [x] `pnpm --filter @effect/ai-openai build` passes
- [x] `pnpm --filter "./packages/ai/**" check` passes (all 4 AI packages)
- [ ] Manual verification against the live OpenAI API: `createFile` decodes when `status_details: null`, `createBatch` decodes when `model: null`


IMax153:
Hi @lindit,

PR #2146, opened 2026-05-09
fix(ai-openai): allow null for Batch.model, output_file_id, error_file_id

lindit:
## Summary

Follow-up to #2145. Same class of bug, different schema: OpenAI returns `null` for several `Batch` fields that the OpenAPI spec marks as optional-only.

- `Batch.model` — `null` on freshly created batches that haven't been validated yet (observed downstream).
- `Batch.output_file_id` — `null` until the batch completes successfully (likely; not yet observed in the wild but obvious candidate).
- `Batch.error_file_id` — `null` unless errors occurred (likely; not yet observed but obvious candidate).

`createBatch` / `retrieveBatch` / `listBatches` fail decoding with:

```
SchemaError(Expected string, got null at ["model"])
```

This PR adds three JSON Patch entries to `packages/ai/openai/codegen.yaml` setting `nullable: true` on each (mirroring the precedent set by #2145 and the existing `ModelResponseProperties.user/safety_identifier/prompt_cache_key` patches). After regen, each field is emitted as:

```ts
"model": Schema.optionalKey(
  Schema.Union([Schema.String, Schema.Null]).annotate({ ... })
)
```

## Note: stacked on #2145

This branch is built on top of `fix/openai-files-status-details-null` (#2145), so GitHub will show that PR's commit until #2145 merges to `main`. The `Generated.ts` diff *unique to this PR* is small and surgical (3 fields, 13 ins / 9 del):

```diff
-  readonly "model"?: string
+  readonly "model"?: string | null
...
-  readonly "output_file_id"?: string
-  readonly "error_file_id"?: string
+  readonly "output_file_id"?: string | null
+  readonly "error_file_id"?: string | null
```

## Note: `Batch.errors` not patched (generator bug)

While sweeping, I noticed OpenAI's own example shows `"errors": null` (Batch component, line 40127 of the spec). I tried adding `nullable: true` for `Batch.errors` too, but the generator handles `nullable: true` on **object-typed** properties incorrectly — it produces a degenerate `Schema.Union([Schema.Struct(...)])` with no `Schema.Null` member, and the corresponding TS type is unchanged. The string-typed fields in this PR work fine because the generator handles `nullable: true` on primitive-typed properties correctly.

I dropped that patch from the PR per the original guidance not to touch the generator. Worth filing as a separate generator bug; the workaround is to express it via `anyOf: [{...object schema...}, {type: 'null'}]` instead of `nullable: true`, which is significantly more verbose to encode as a JSON Patch.

## Test plan

- [x] `pnpm --filter @effect/ai-openai check` passes
- [x] `pnpm --filter @effect/ai-openai test` passes (90 tests)
- [x] `pnpm --filter @effect/ai-openai build` passes
- [ ] Manual verification: call `createBatch` against the live OpenAI API and confirm response decodes when `model: null`


lindit:
Closing — squashing into #2145.

PR #2151, opened 2026-05-09
feat(effect): add HttpMiddleware.compression for gzip/deflate response compression

kitlangton:
## Summary

Adds an HTTP response compression middleware so Effect HttpApi users don't have to roll their own. Effect-smol's `HttpMiddleware` ships logging, tracing, and CORS, but no compression — Hono, Express, Fastify, and nginx all ship one.

This is platform-agnostic: it uses the Web Standard `CompressionStream` (available in Node 18+, Bun, Deno, and browser-style runtimes) instead of `node:zlib`, so it works anywhere effect-smol runs.

## What's added

- **`HttpMiddleware.compression(options?)`** — middleware that conditionally compresses response bodies with gzip or deflate.
- **`HttpRouter.compression(options?)`** — matching `Layer<never, never, HttpRouter>` registration, mirroring the `cors` pattern.
- **`HttpServerResponse.removeHeader`** — small public addition; the compression middleware uses it to clear stale `Content-Length` after swapping in a Stream body of unknown length.

### Defaults (mirroring Hono's `compress`)

- `threshold`: 1024 bytes
- `compressibleContentType`: regex covering `text/*`, common `application/*` (json/xml/javascript/wasm/…), `image/svg+xml`, `+json`/`+xml`/`+yaml`/`+text` suffixes, etc. `text/event-stream` is always excluded so SSE chunks flow immediately.
- `encodings`: `["gzip", "deflate"]` — gzip preferred when both are accepted. `br` is intentionally not included; it would need separate handling and isn't supported by `CompressionStream` everywhere.
- `skip?: (request) => boolean` — user predicate (e.g., to bypass specific paths).

### Skip conditions

- `HEAD` requests
- Response already has `Content-Encoding`
- Response has `Cache-Control: no-transform`
- Content type not in the allowlist, or is `text/event-stream`
- Known `Content-Length` < `threshold`
- Body is `Empty` / `Raw` / `FormData` (only `Uint8Array` and `Stream` are compressed)

### When compressing

- Replaces body via `Stream` (or `Uint8Array` if the original body was `Uint8Array`)
- Sets `Content-Encoding`
- Drops stale `Content-Length` for stream bodies (compressed length unknown)
- Merges `Accept-Encoding` into any existing `Vary` header (preserving e.g. `Vary: Origin` from CORS)
- Weakens strong `ETag` to `W/...`

## Test plan

15 new unit tests covering all branches:

- [x] Compresses `Uint8Array` body with gzip; round-trip decompresses correctly
- [x] Prefers gzip when both gzip and deflate are accepted
- [x] Compresses `Stream` body and clears stale `Content-Length`
- [x] Skips when `Accept-Encoding` is missing
- [x] Skips when no encoding is supported (e.g., `br` only)
- [x] Skips bodies smaller than threshold
- [x] Skips non-compressible content types (e.g., `image/png`)
- [x] Skips `text/event-stream`
- [x] Skips when response already has `Content-Encoding`
- [x] Skips when `Cache-Control: no-transform`
- [x] Skips `HEAD` requests
- [x] Respects user `skip` predicate
- [x] Merges `Accept-Encoding` into existing `Vary` header
- [x] Weakens strong `ETag` (`"abc"` → `W/"abc"`)
- [x] Leaves weak `ETag` unchanged
- [x] `pnpm check:tsgo`, `pnpm lint-fix`, `pnpm docgen`
- [x] All 107 `test/unstable/http/` tests pass
- [x] Changeset added (`minor`)

## Open questions for the maintainers

- Naming: I went with `HttpMiddleware.compression` to match the existing module style; happy to rename to `compress` if that's preferred.
- Brotli: skipped intentionally for a first cut; `CompressionStream` doesn't accept `"br"` in Node, so it would need a Node-only path. Easy follow-up.
- Configuration shape: the `compressibleContentType` option is a single `RegExp` (matching Hono). Open to making it a predicate `(contentType) => boolean` if more flexibility is wanted.

### DIFF
--- a/.changeset/add-http-compression-middleware.md
+++ b/.changeset/add-http-compression-middleware.md
@@ -1,0 +1,9 @@
+---
+"effect": minor
+---
+
+Add `HttpMiddleware.compression` (and matching `HttpRouter.compression` layer) for gzip/deflate response compression.
+
+The middleware uses the Web Standard `CompressionStream`, so it works in Node, Bun, Deno, and browser-style runtimes without a `node:zlib` dependency. It supports both `Uint8Array` and `Stream` response bodies, and mirrors Hono's `compress` defaults: skips `HEAD` requests, responses that already have `Content-Encoding`, `Cache-Control: no-transform`, `text/event-stream`, non-compressible content types, and bodies smaller than the configured threshold (default 1024 bytes). When compression is applied, `Vary: Accept-Encoding` is merged into any existing `Vary` header and strong `ETag`s are weakened.
+
+Also adds `HttpServerResponse.removeHeader`, used internally to drop stale `Content-Length` after a stream body swap.
--- a/packages/effect/src/unstable/http/HttpMiddleware.ts
+++ b/packages/effect/src/unstable/http/HttpMiddleware.ts
@@ -368,9 +368,16 @@ export type CompressionEncoding = "gzip" | "deflate"
 
 const compressionEncodingsDefault: ReadonlyArray<CompressionEncoding> = ["gzip", "deflate"]
 
-// Mirrors Hono's compressible content-type allowlist.
+// Curated set of compressible content types. Each entry is flagged
+// `compressible: true` in the IANA-derived mime-db dataset. Already-compressed
+// formats (woff/woff2, image/png, image/jpeg, video/*, audio/*, etc.) are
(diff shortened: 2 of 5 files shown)


IMax153:
Hi @kitlangton,

PR #2158, opened 2026-05-11
fix(schema): preserve symbol link of obj made by Schema Struct, enable VSCode "Go to definitions"(F12) jump

### LINKED ISSUE #2141: Schema: when trying navigate to schema field definition using `go to definition F12` on VSCode, shows "No definition found for 'xx'"

4.0.0-beta.62

```typescript
const OneSchema = Schema.Struct({
  cannotGoToDefinition: Schema.Boolean
})

function main() {
  const one = OneSchema.make({ cannotGoToDefinition: false })
  const cannotGoToDefinition = one.cannotGoToDefinition // VSCode F12 shows: No definition found for 'cannotGoToDefinition '
}
```

IDE like vscode should be able to navigate to the schema field definition

### What do you see instead?

Image

### Additional information

Typescript version: ^6.0.2

I tested on Effect v3, it works as expected

ColaFanta:
## Type
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Change `Type_<>` implementation, from using `Exclude<F, O | M>` type util to `keyof F as xx`. 

This keeps IDE provenance link andenables clicking "Go to definition (F12)" in VSCode on an object made from Schema Struct jumps to the correct Struct field definition. `keyof` is a must because this ensures all field keys are derived from input thus holds the correct reference.

Also Fixed type error of `TaggedClass` and `TaggedErrorClass` that appears after the above change applied.

There is another PR #1862 mentioned this fix but no further response. But implementation in this PR seems more straight forward and can be another option.

## Related

- Closes #2141

### DIFF
--- a/.changeset/floppy-items-admire.md
+++ b/.changeset/floppy-items-admire.md
@@ -1,5 +1,5 @@
 ---
-"effect": major
+"effect": patch
 ---
 
 Change `Type_<>` implementation, from using `Exclude<F, O | M>` type util to `keyof F as xx`, this implementation keeps IDE provenance link. This enables clicking "Go to definition (F12)" in VSCode on an object made from Schema Struct jumps to the correct Struct field definition.
--- a/.changeset/floppy-items-admire.md
+++ b/.changeset/floppy-items-admire.md
@@ -1,0 +1,5 @@
+---
+"effect": major
+---
+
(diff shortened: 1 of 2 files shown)


gcanti:
This is a nice improvement, thank you @ColaFanta

PR #2161, opened 2026-05-11
Fix RpcServer HTTP finalizer request IDs

wking-io:
## Summary
- Move `requestIds` initialization before the HTTP protocol request finalizer closes over it.
- Prevent an early request-scope finalizer from throwing a temporal dead zone error before request decoding starts.

## Context
While using `RpcServer.toHttpEffect(...)` with JSON-RPC in a Cloudflare Worker runtime, the request scope finalizer can run before the generator reaches the later `requestIds` declaration. In the built output this surfaced as `Cannot access 'oe' before initialization`.

### DIFF
--- a/.changeset/fix-rpc-http-requestids-finalizer.md
+++ b/.changeset/fix-rpc-http-requestids-finalizer.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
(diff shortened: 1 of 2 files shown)


tim-smart:
Just needs a `pnpm changeset` :)

PR #2162, opened 2026-05-11
Fix RpcServer toHttpEffect lifetime

wking-io:
## Summary
- Detach the `toHttpEffect` server loop from the ambient scope so the returned HTTP effect can be used after construction scope cleanup.
- Keep non-framed HTTP clients registered until the server loop writes `Exit` and `ClientEnd`, allowing `Queue.collect` to finish.
- Add a regression test for non-framed JSON-RPC where `toHttpEffect` is constructed in a short-lived scope before handling a request.

## Context
A downstream Cloudflare Worker setup using Alchemy constructs `RpcServer.toHttpEffect(...)` for `RpcSerialization.jsonRpc()`. After fixing the separate finalizer `requestIds` TDZ crash, the official adapter decoded the request and sent EOF, then hung in the non-framed `Queue.collect` path until workerd canceled the request. Runtime logs showed the server loop never emitted `Exit`/`ClientEnd` before the ambient request scope finalized.

This change makes `toHttpEffect` own the server loop lifetime independently of the construction scope and avoids treating non-framed one-shot HTTP requests as disconnects before their collected response is produced.

## Test plan
- `pnpm vitest run packages/effect/test/rpc/RpcServer.test.ts`
- `pnpm --filter effect check`

### DIFF
--- a/packages/effect/test/rpc/RpcServer.test.ts
+++ b/packages/effect/test/rpc/RpcServer.test.ts
@@ -1,0 +1,43 @@
+import { assert, describe, it } from "@effect/vitest"
+import { Effect, Schema } from "effect"
+import { HttpEffect } from "effect/unstable/http"
+import { Rpc, RpcGroup, RpcSerialization, RpcServer } from "effect/unstable/rpc"
+
+const TestGroup = RpcGroup.make(
+  Rpc.make("Ping", { success: Schema.String })
+)
+
+describe("RpcServer", () => {
+  it.effect("toHttpEffect keeps non-framed json-rpc server alive outside construction scope", () =>
+    Effect.gen(function*() {
+      const httpEffect = yield* RpcServer.toHttpEffect(TestGroup).pipe(
+        Effect.provide(TestGroup.toLayer({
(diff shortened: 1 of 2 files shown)


packages/effect/test/rpc/RpcServer.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+describe("RpcServer", () => {
+  it.effect("toHttpEffect keeps non-framed json-rpc server alive outside construction scope", () =>
+    Effect.gen(function*() {
+      const httpEffect = yield* RpcServer.toHttpEffect(TestGroup).pipe(
+        Effect.provide(TestGroup.toLayer({
+          Ping: () => Effect.succeed("pong")
+        })),
+        Effect.provideService(RpcSerialization.RpcSerialization, RpcSerialization.jsonRpc()),
+        Effect.scoped
```
tim-smart, line 18: This `Effect.scoped` will shutdown the server immediately.

IMax153:
Hi @wking-io,

PR #2163, opened 2026-05-11
remove Effect.Yieldable

tim-smart:
- **wip**
- **wip**
- **wip**


bwbuchanan:
@tim-smart Looks like the migration docs need an update because they're still referencing `Yieldable`.

Support for yielding an `Option` in `Effect.gen` looks to have been removed, and I'm not sure whether this was intentional. Do I need to wrap all my option values in `Effect.fromOption` before yielding them now?

Also same for `Result`, needs to be transformed, but `Data.TaggedError` is still yieldable.

The deletion in [‎packages/effect/test/Effect.test.ts](https://github.com/Effect-TS/effect-smol/pull/2163/changes#diff-0336f4733fc93a063e5a763040ab0a4e5cfaa9dd834fb17b510bbabb30e223f3) suggests it was intentional to remove the yieldability of `Option` and `Result` but it would nice to have confirmation of this before undertaking a large migration.

patroza:
> @tim-smart Looks like the migration docs need an update because they're still referencing `Yieldable`.
> 
> Support for yielding an `Option` in `Effect.gen` looks to have been removed, and I'm not sure whether this was intentional. Do I need to wrap all my option values in `Effect.fromOption` before yielding them now?
> 
> Also same for `Result`, needs to be transformed, but `Data.TaggedError` is still yieldable.
> 
> The deletion in `[‎packages/effect/test/Effect.test.ts](https://github.com/Effect-TS/effect-smol/pull/2163/changes#diff-0336f4733fc93a063e5a763040ab0a4e5cfaa9dd834fb17b510bbabb30e223f3) suggests it was intentional to remove the yieldability of `Option`and`Result` but it would nice to have confirmation of this before undertaking a large migration.

Same

tim-smart:
> @tim-smart Looks like the migration docs need an update because they're still referencing `Yieldable`.
> 
> Support for yielding an `Option` in `Effect.gen` looks to have been removed, and I'm not sure whether this was intentional. Do I need to wrap all my option values in `Effect.fromOption` before yielding them now?
> 
> Also same for `Result`, needs to be transformed, but `Data.TaggedError` is still yieldable.
> 
> The deletion in [‎packages/effect/test/Effect.test.ts](https://github.com/Effect-TS/effect-smol/pull/2163/changes#diff-0336f4733fc93a063e5a763040ab0a4e5cfaa9dd834fb17b510bbabb30e223f3) suggests it was intentional to remove the yieldability of `Option` and `Result` but it would nice to have confirmation of this before undertaking a large migration.

A proper migration doc will be worked on closer to release, at the moment it is largely incomplete.

Option / Result is no longer directly yieldable, so yes you would use Effect.from* in a generator.

PR #2180, opened 2026-05-13
EFF-809: Add platform Crypto service

IMax153:
## Summary
- Add a platform-agnostic `Crypto` service with random bytes, UUIDv4/UUIDv7 generation, random helpers, and digest operations.
- Add Node, Bun, and Browser platform layers plus service aggregate wiring.
- Remove `Random.nextUUIDv4` and document/test the new Crypto service replacement.
- Address review feedback by making `DigestAlgorithm` a string literal union and moving derived random helpers / UUIDv4 onto the `Crypto` service via `Crypto.make`.
- Address follow-up feedback by adding `Crypto.randomUUIDv7`, derived from platform `randomBytes(16)` and `Clock.currentTimeMillis` for the UUIDv7 timestamp.
- Keep Node digests on `node:crypto` `createHash` and use Web Crypto digest algorithm strings directly in the browser implementation.

## Verification
- `pnpm lint-fix`
- `pnpm codegen` (ran; unrelated generated barrel churn was restored, with focused Crypto barrel docs kept)
- `pnpm test packages/effect/test/Crypto.test.ts packages/effect/test/Random.test.ts packages/platform-browser/test/BrowserCrypto.test.ts packages/platform-node/test/NodeCrypto.test.ts`
- `pnpm check:tsgo`

### DIFF
--- a/packages/effect/src/Crypto.ts
+++ b/packages/effect/src/Crypto.ts
@@ -4,9 +4,9 @@
  * `@effect/platform-bun`, and `@effect/platform-browser` provide concrete
  * implementations backed by the host platform's cryptography APIs.
  *
- * Use `Crypto` for cryptographic randomness, UUIDv4 generation, and message
- * digests. Use `Random` only when you need deterministic pseudo-random values,
- * such as in tests or simulations.
+ * Use `Crypto` for cryptographic randomness, UUIDv4 generation, random values,
+ * and message digests. The base `Random` service is not cryptographically
+ * secure unless you replace it with a cryptographically secure implementation.
  *
  * @example
  * ```ts
@@ -16,7 +16,8 @@
  *   Crypto.Crypto,
  *   Crypto.make({
  *     randomBytes: (size) => Effect.succeed(new Uint8Array(size)),
- *     randomUUIDv4: Effect.succeed("00000000-0000-4000-8000-000000000000"),
+ *     nextIntUnsafe: () => 1,
+ *     nextDoubleUnsafe: () => 0.5,
  *     digest: (_algorithm, data) => Effect.succeed(data)
  *   })
  * )
@@ -37,7 +38,8 @@
  *   Crypto.Crypto,
  *   Crypto.make({
  *     randomBytes: (size) => Effect.succeed(new Uint8Array(size)),
- *     randomUUIDv4: Effect.succeed("00000000-0000-4000-8000-000000000000"),
+ *     nextIntUnsafe: () => 1,
+ *     nextDoubleUnsafe: () => 0.5,
  *     digest: (_algorithm, data) => Effect.succeed(data)
  *   })
  * )
@@ -53,6 +55,7 @@ import * as Context from "./Context.ts"
 import * as Data from "./Data.ts"
 import * as Effect from "./Effect.ts"
 import type { PlatformError } from "./PlatformError.ts"
+import type * as Random from "./Random.ts"
 
 const TypeId = "~effect/platform/Crypto"
 
@@ -93,8 +96,8 @@ export const DigestAlgorithm = Data.taggedEnum<DigestAlgorithm>()
  * Platform-agnostic cryptographic operations.
  *
  * `Crypto` implementations must use cryptographically secure platform APIs.
- * UUIDv4 generation should use native platform UUID APIs such as
- * `crypto.randomUUID` where available.
+ * The module-level random generators are backed by the random methods on this
+ * service.
  *
  * @example
  * ```ts
@@ -104,7 +107,8 @@ export const DigestAlgorithm = Data.taggedEnum<DigestAlgorithm>()
  *   Crypto.Crypto,
  *   Crypto.make({
  *     randomBytes: (size) => Effect.succeed(new Uint8Array(size)),
- *     randomUUIDv4: Effect.succeed("00000000-0000-4000-8000-000000000000"),
+ *     nextIntUnsafe: () => 1,
+ *     nextDoubleUnsafe: () => 0.5,
  *     digest: (_algorithm, data) => Effect.succeed(data)
  *   })
  * )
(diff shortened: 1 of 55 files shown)


packages/platform-browser/src/BrowserCrypto.ts
```diff
@@ -0,0 +1,157 @@
...
+  })
+
+const systemError = (method: string, description: string, cause: unknown) =>
+  PlatformError.systemError({
+    module: "Crypto",
+    method,
+    _tag: "Unknown",
+    description,
+    cause
+  })
```
tim-smart: Don't really need these helpers

packages/platform-browser/src/BrowserCrypto.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+const makeWith = (
+  getBackend: (method: string) => Effect.Effect<CryptoBackend, PlatformError.PlatformError>,
+  getUnsafeBackend: () => CryptoBackend | undefined
+): EffectCrypto.Crypto => {
+  const unsafeRandomBytes = (size: number): Uint8Array => {
+    const crypto = getUnsafeBackend()
+    if (typeof crypto?.getRandomValues !== "function") {
+      throw new Error("crypto.getRandomValues is not available")
```
tim-smart: Remove the throw

packages/platform-node-shared/src/NodeCrypto.ts
```diff
@@ -0,0 +1,85 @@
...
+  })
+
+const systemError = (method: string, description: string, cause: unknown) =>
+  PlatformError.systemError({
+    module: "Crypto",
+    method,
+    _tag: "Unknown",
+    description,
+    cause
+  })
```
tim-smart: Remove the error wrappers

packages/platform-node-shared/src/NodeCrypto.ts
```diff
@@ -0,0 +1,85 @@
...
+  return value / 2 ** 53
+}
+
+const nextIntUnsafe = (): number =>
+  Math.floor(nextDoubleUnsafe() * (Number.MAX_SAFE_INTEGER - Number.MIN_SAFE_INTEGER + 1)) + Number.MIN_SAFE_INTEGER
+
+const digest: EffectCrypto.Crypto["digest"] = (algorithm, data) =>
+  Effect.map(
+    Effect.tryPromise({
+      try: () => NodeCrypto.webcrypto.subtle.digest(toSubtleAlgorithm(algorithm), new Uint8Array(data)),
```
tim-smart: This should use the node crypto hash apis instead of the webcrypto wrapper.

tim-smart (changes requested):

erlangxk:
https://github.com/nodejs/node/pull/62553

https://www.postgresql.org/docs/current/release-18.html, it supports UUIDv7 too.

is it possible to include UUIDv7 generation?

packages/effect/src/Crypto.ts
```diff
@@ -0,0 +1,304 @@
...
+  readonly Sha512: {}
+}>
+
+/**
+ * Constructors and matchers for `DigestAlgorithm` values.
+ *
+ * @since 4.0.0
+ * @category constructors
+ */
+export const DigestAlgorithm = Data.taggedEnum<DigestAlgorithm>()
```
tim-smart: DigestAlgorithm should just be a string literal union.

packages/effect/src/Crypto.ts
```diff
@@ -0,0 +1,304 @@
...
+ */
+export const make = (impl: Omit<Crypto, typeof TypeId>): Crypto => Crypto.of({ ...impl, [TypeId]: TypeId })
+
+/**
+ * Generates cryptographically secure random bytes.
+ *
+ * @since 4.0.0
+ * @category accessors
+ */
+export const randomBytes = (size: number): Effect.Effect<Uint8Array, PlatformError, Crypto> =>
```
tim-smart: Remove all the accessors and add everything to the service interface, to prevent requirement leakage.

Update the `make` constructor to add the derived methods, similar to `ChildProcessSpawner`.

tim-smart (changes requested):

packages/effect/src/Crypto.ts
```diff
@@ -0,0 +1,269 @@
...
+  /**
+   * Uses the cryptographically secure random generator to shuffle the supplied
+   * iterable.
+   */
+  readonly randomShuffle: <A>(elements: Iterable<A>) => Effect.Effect<Array<A>>
+
+  /**
+   * Generates a cryptographically secure UUIDv4 string.
+   */
+  readonly randomUUIDv4: Effect.Effect<string, PlatformError>
```
tim-smart: We should add `randomUUIDv7` too, which uses the `Clock` service for the timestamp

IMax153:
@tim-smart another thing I was thinking about in this PR is we should make sure that the JSDocs line up with the range of random numbers generated and keep that consistent in the crypto service.

For example, `Random.next` says it generates between 0 (inclusive) and 1 (inclusive), which is incorrect since I believe `Math.random()` generates in the half-open range `[0, 1)`.

Also, I think the `nextDoubleUnsafe` impl for NodeCrypto can be simplified to:

```ts
const nextDoubleUnsafe = (): number => {
  const array = new Uint32Array(1)
  NodeCrypto.getRandomValues(array)
  return array[0] / (0xFFFFFFFF + 1)
}
```

In addition, the `nextIntUnsafe` method _could_ be simplified to the following:

```ts
function getSecureRandomNumber() {
  const buffer = crypto.randomBytes(4)
  return buffer.readUInt32BE(0)
} 
```

but that would imply a random range of 0 to 2^32 - 1, which is different than the range that `Random.nextInt` produces right now (-9007199254740991 to 9007199254740991). 

In summary, I think we first need to standardize the random ranges we are targeting and then we can bring all the methods into sync and clean them up. @tim-smart lmk what you think.

tim-smart (dismissed):

PR #2187, opened 2026-05-14
fix(HttpApi): handle empty optional JSON payloads

### LINKED ISSUE #2188: HttpApi optional JSON payload rejects empty request body
`HttpApiEndpoint` optional JSON payloads reject empty request bodies before the handler runs.

Repro PR: https://github.com/Effect-TS/effect-smol/pull/2187

Minimal shape:

```ts
const Payload = Schema.Struct({
  name: Schema.optional(Schema.String)
})

HttpApiEndpoint.post("create", "/create", {
  payload: Schema.optional(Payload),
  success: Schema.String
})
```

For:

```http
POST /create
Content-Type: application/json
Content-Length: 0
```

Expected: handler receives `ctx.payload === undefined`.

Actual: HttpApi returns `400` before the handler.

Maybe this is intentional semantics, but I found it somewhat surprising and awkward to work around. It seems related to JSON codec rewriting: `Schema.toCodecJson` represents `undefined` as JSON `null`, while an empty HTTP body is transport absence and seems like it should preserve the endpoint's optional-body semantics.

kitlangton:
Closes #2188.

## Summary
- Adds a regression test for an optional JSON payload endpoint receiving an empty request body.
- Includes one possible implementation: if the payload Type accepts `undefined`, decode an empty JSON body as decoded `undefined` instead of routing it through the JSON codec's encoded `null` representation.
- Adds a patch changeset.

## Notes
The reproducer is the important part; the implementation is one candidate fix and I am happy to adjust it if there is a preferred semantic/modeling approach.

## Test
- `pnpm vitest packages/platform-node/test/HttpApi.test.ts -t \"optional JSON payload accepts an empty request body\"`

### DIFF
--- a/.changeset/httpapi-empty-optional-payload.md
+++ b/.changeset/httpapi-empty-optional-payload.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `HttpApiBuilder` empty JSON body decoding for optional payloads. See [#2188](https://github.com/Effect-TS/effect-smol/issues/2188).
--- a/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
@@ -497,6 +497,7 @@ type PayloadDecoder =
   }
   | {
     readonly _tag: "Json" | "FormUrlEncoded" | "Uint8Array" | "Text"
+    readonly undefinedOnEmpty: boolean
     readonly nullOnEmpty: boolean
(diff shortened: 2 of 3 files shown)


tim-smart:
Schema.optional is for properties. For HttpApi payloads you can use `payload: [SomePayload, HttpApiSchema.NoContent]`

kitlangton:
Closing — `payload: [SomePayload, HttpApiSchema.NoContent]` is the idiomatic way to handle this. Thanks @tim-smart!

PR #2197, opened 2026-05-16
feat(rpc): Add parseOptions

patroza:
Effect-TS/effect#6361 RPC schema parsing in v4 was not configurable at the RPC boundary, forcing schema-level annotations for behaviors like concurrency. This change adds first-class parse concurrency support to both client and server so schema parsing concurrency can be controlled per RPC instance.

- **API surface**
  - Added `parseOptions?: Pick<AST.ParseOptions, "concurrency">` to:
    - `RpcClient.make(...)`
    - `RpcServer.make(...)`
    - server forwarding APIs (`layer`, `layerHttp`, `toHttpEffect`, `toHttpEffectWebsocket`)

- **Client-side wiring**
  - Threaded concurrency parse options through client schema operations used during request/response processing:
    - payload encoding
    - stream chunk decoding
    - exit decoding
  - Refactored schema cache creation to be parse-options aware per client instance.
  - Normalized runtime `parseOptions` before schema usage so only `{ concurrency }` is forwarded, even if extra properties are provided at runtime.

- **Server-side wiring**
  - Threaded concurrency parse options through server schema operations used during request/response processing:
    - request payload decoding
    - stream chunk encoding
    - exit encoding
    - defect encoding
  - Normalized runtime `parseOptions` before schema usage so only `{ concurrency }` is forwarded, even if extra properties are provided at runtime.

- **Tests**
  - Updated RPC parse-options tests to validate concurrency-only configuration on both client and server paths.
  - Added coverage that passes extra runtime parse options (for example `onExcessProperty`) and verifies only concurrency is effectively applied.

- **Release note**
  - Added a changeset describing RPC parse concurrency capability.

Example usage:

```ts
const client = yield* RpcClient.make(group, {
  parseOptions: {
    concurrency: "unbounded"
  }
})

const server = yield* RpcServer.make(group, {
  parseOptions: {
    concurrency: 16
  }
})
```


IMax153:
Hi @patroza,

PR #2198, opened 2026-05-16
feat(Schema): preserve encodedKey across field reuse, spread, and mapFields

patroza:
## Summary

In effect v3, a struct field's encoded key rename traveled with the field. In effect-smol, renames only existed as a top-level `encodeKeys` mapping on the struct AST, so:

- `Struct.fields.x` exposed the un-renamed encoded shape
- Spreading shared field maps lost the rename
- `mapFields` / `Class.mapFields` dropped the rename
- Reusing a field across structs forced you to repeat the rename in every parent

This PR moves the rename from a struct-level concern to a **field-local annotation** and threads it through the type system so it survives reuse, spreading, `mapFields`, `encodeKeys`, and `flip ∘ flip`.

## Approach

Two alternatives were considered:

1. **Expose `encodedFields` / `encodedMapFields` on the struct** or adding `fields`/`mapFields` to the return value of `S.encodedKeys`. Keeps the rename at the struct level, still does not help when a single field schema is consumed by an unrelated struct.
2. **Make the rename a property of the field schema itself.** A new `Schema.encodedKey("foo")` combinator attaches an `encodedKey` annotation to the field's `Key` annotation bag. `Struct(...)` discovers those annotations at construction time and folds them into an `encodeKeys(...)` mapping automatically. Field reuse, spread, and `mapFields` then "just work" because the rename rides on the field.

Option 2 won — same author ergonomics, no new struct API, and the rename is preserved through any operation that preserves the field schema.

### Type-level plumbing

A new `EncodedKey extends PropertyKey` parameter is appended to `Bottom`:

```ts
export interface Bottom<
  ...,
  out EncodedOptionality extends Optionality = "required",
  out EncodedKey extends PropertyKey = never
> { readonly "~encoded.key": EncodedKey }


jdharrisnz:
Nice to see this come back. Would this work on classes too or only structs?

IMax153:
Hi @patroza,

PR #2204, opened 2026-05-17
feat(bun-test): add @effect/bun-test package

emilienbidet:
## Summary

Adds a new `@effect/bun-test` workspace package that mirrors the `@effect/vitest`
API surface (`it.effect`, `it.live`, `it.layer`, `it.prop`, `flakyTest`,
`layer`, …) but runs on Bun's built-in
[`bun:test`](https://bun.sh/docs/cli/test) runner instead of Vitest.

Companion PR to Effect-TS/effect#6236; addresses the same request as
Effect-TS/effect#5964.

## Why

Quoting Effect-TS/effect#5964: Bun's test runner is significantly faster than
Vitest, and the Vitest dependency of `@effect/vitest` is a thin wrapper around
`it()` and `describe()` — the actual Effect test machinery (TestClock,
TestConsole, fast-check integration) is runner-agnostic. This PR ports that
thin wrapper to `bun:test` for the v4 line in `effect-smol`.

## What's in the package

`packages/bun-test/` mirrors `packages/vitest/`:

- `src/index.ts` — public API, types, re-exports of `bun:test` primitives
- `src/internal/internal.ts` — Effect runner + tester/layer/prop wrappers
  (uses `Effect.fnUntraced`, `Layer.buildWithMemoMap`, `Layer.mergeAll(TestConsole.layer, TestClock.layer())`
  to match the v4 vitest package shape)
- `src/utils.ts` — assertion helpers (copied from `@effect/vitest` `utils.ts`,
  `vassert.instanceOf` swapped for `node:assert`)
- `test/index.test.ts` — ported from `packages/vitest/test/index.test.ts`
- `tsconfig.json`, `docgen.json`, `bunfig.toml`, `package.json`, `README.md`,
  `LICENSE`

A changeset is included (`.changeset/add-effect-bun-test.md`).

## Differences from `@effect/vitest`

Bun's test runner has a smaller surface than Vitest. The PR documents these
gaps in the package README; they were not papered over with runtime shims:

- `addEqualityTesters` is a **no-op** — `bun:test`'s `expect` has no
  `addEqualityTesters` API.
- **`TestContext`** is synthesised inside the test wrapper rather than passed
  in by the runner, because `bun:test` doesn't pass a context object.
- **`re-export from \"bun:test\"`** doesn't work in Bun (1.2.x), so the public
  primitives are re-exported via `const` bindings rather than `export *`.

## Test plan

- [x] `pnpm install` at the workspace root resolves cleanly
- [x] `pnpm --filter=@effect/bun-test check` passes
- [x] `bun test` inside `packages/bun-test` is green
  - 24 pass, 4 skip, 0 fail
- [ ] Maintainers to decide whether CI should grow a Bun job for this package

## Notes for review

- The `layer > release` test had its trailing `afterAll(...)` reordered so it
  runs **after** the layer's own `afterAll` (Bun runs `afterAll` hooks in
  registration order).
- The `it.live.fails("interrupts on timeout", ...)` test from the vitest suite
  was dropped because Bun has no fiber-interrupt-on-failure hook equivalent.
- No upstream files outside `packages/bun-test/`, `.changeset/`, and
  `pnpm-lock.yaml` were touched.


ebramanti:
Would love to see this land, I have had to vendor this locally in my project in the meantime

artile:
+1 to this

IMax153:
Hi @emilienbidet,

emilienbidet:
Replacement PR opened against `Effect-TS/effect:main` as requested: https://github.com/Effect-TS/effect/pull/7521

The code was re-ported against the current v4 `main` — the property-testing layer now uses `effect/unstable/arbitrary/Arbitrary` like the current `@effect/vitest` (this PR's version was FastCheck-based), and timeouts now actually interrupt the Effect fiber (the wrapper aborts the synthesized test context's `AbortSignal`), so the two timeout-interruption tests from the vitest suite could be ported instead of dropped.

PR #2205, opened 2026-05-17
feat: add support for merging external events into `Prompt.custom`

lloydrichards:
## Type

- [x] Feature

## Description

I've been wanting a way to link in external data into the `Prompt.custom` in a way that triggers a re-render without needing to press a key stroke. In the current implementation I need to fake a key stroke using `process.stdin` as a hack-y work around or build me own render loop from scratch.  Ideally what I wanted was an API where I could just pass in a stream of events and be able to trigger the action with the next state:

```ts
const eventQueue = yield * Queue.make<number>()

const prompt = Prompt.custom(
  { count: 0 },
  Queue.asDequeue(eventQueue), // <-- provide the event queue as a dequeue to the prompt
  {
    render: (state) => Effect.succeed(`Count: ${state.count}`),
    process: (input, state) =>
      Effect.succeed(
        Match.value(input).pipe(
          // handle user input
          Match.tag("Input", () => Action.Submit({ value: state.count })),
          // handle external events from the queue
          Match.tag("Event", (input) => Action.NextFrame({ state: { count: state.count + input.value } })),
          Match.exhaustive
        )
      ),
    clear: () => Effect.succeed("")
  }
)
```

I've been digging around in the Prompt.ts for a while now so I already knew the `Terminal.readInput` was a Queue and that it shouldn't be too hard to just merge two Queues together.

What I settled on was a type overload where there is the current implementation without the event doing the same thing (non-breaking) and the new type that adds the new `events` Queue.Dequeued argument ~and extends the handlers with the `received` callback~.  This way adding the queue can be a progressive enhancement on an already functioning user input only prompt.

Some things I needed to mull over:
- The `Prompt` isn't responsible for cleaning up the `Queue`, this needs to be handled by the user
- There is a very small possibility of an event and a user input happening simultaneously and which one is picked first is indeterministic (I think)
- Queue needs to not fail (`Queue.Dequeue<A, never>` so the user is responsible outside of the Prompt to handle these causes.

## Related

I wanted to make an issue first but needed to open the repo to poke around in the types and next thing I know I was already halfway through implementing it 🤦‍♂️

Open to other suggestions or requests to change the approach.

---

Edit: Have changed the api a bit to put the event before the handler as well as removed the new received callback in favour of overloading the process with a taggedEnum when event us present.

### DIFF
--- a/packages/effect/src/unstable/cli/Prompt.ts
+++ b/packages/effect/src/unstable/cli/Prompt.ts
@@ -94,6 +94,18 @@ export interface ActionDefinition extends Data.TaggedEnum.WithGenerics<2> {
 }
 
 /**
+ * Represents the input that should be processed by a `Prompt` based upon user
+ * input or an external event received during the current frame.
+ *
+ * @category models
+ * @since 4.0.0
+ */
+export type ProcessInput<A> = Data.TaggedEnum<{
+  readonly Input: { readonly input: Terminal.UserInput }
+  readonly Event: { readonly value: A }
+}>
+
+/**
  * Represents the set of handlers used by a `Prompt`.
  *
  * **Details**
@@ -104,7 +116,7 @@ export interface ActionDefinition extends Data.TaggedEnum.WithGenerics<2> {
  * @category models
  * @since 4.0.0
  */
-export interface Handlers<State, Output> {
+export interface Handlers<State, Output, Input = Terminal.UserInput> {
   /**
    * A function that is called to render the current frame of the `Prompt`.
    */
@@ -117,7 +129,7 @@ export interface Handlers<State, Output> {
    * `Prompt.Action` that should be taken.
    */
   readonly process: (
-    input: Terminal.UserInput,
+    input: Input,
     state: State
   ) => Effect.Effect<Action<State, Output>, never, Environment>
   /**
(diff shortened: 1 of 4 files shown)


lloydrichards:
here are a couple concrete examples of where this would be used:

- [log viewer](https://github.com/lloydrichards/effect-boxes/blob/main/apps/scratchpad/09-log-viewer.ts) <- being able to pass in streams logs to show in a scrollable viewport
- [perlin flow](https://github.com/lloydrichards/effect-boxes/blob/main/apps/scratchpad/08-perlin-flow-field.ts) <- add a a tickEvent which can move the animation loop along

packages/effect/src/unstable/cli/Prompt.ts
```diff
@@ -762,19 +762,42 @@ export const confirm = (options: ConfirmOptions): Prompt<boolean> => {
...
+    events: Queue.Dequeue<A, never>
+  ): Prompt<Output>
+} = <State, Output, A>(
   initialState: State | Effect.Effect<State, never, Environment>,
-  handlers: Handlers<State, Output>
+  handlers: Handlers<State, Output> & {
+    readonly receive?: (value: A, state: State) => Effect.Effect<Action<State, Output>, never, Environment>
+  },
+  events?: Queue.Dequeue<A, never>
 ): Prompt<Output> => {
```
IMax153, line 790: - I would prefer avoiding a separate `receive` handler and instead merge the event types into the process input, and force the user to handle that logic there. Otherwise, we have two separate handlers that are doing the same thing. (required)
- The overload that takes the events `Queue` probably should have the `Queue` passed before the handlers so that the type of the events gets fixed before you specify the handlers. (maybe)
- One other thing to consider is using an options object to pass the external events `Queue` to "future proof" addition of further custom options. (maybe)

IMax153 (changes requested):

packages/effect/src/unstable/cli/Prompt.ts
lloydrichards: I made the changes: 
- remove the `receive` callback
- combine the new `ProcessInput` taggedEnum with the `Terminal.UserInput` for the `process`
- moved the `events` Queue to before the handler
- added the typeTests

I've left out the option object since it kind of conflicts with the moving of the `events` Queue since I would expect the option key to be an optional arg at the end of a function call but would make the `A` type defined afterward.  Happy to reconsider this if you think it would be better to look like:

```ts
const prompt = Prompt.custom(
  { count: 0 },
  {
    render: (state) => Effect.succeed(`Count: ${state.count}`),
    process: (input, state) =>
      Effect.succeed(
        Match.value(input).pipe(
          Match.tag("Input", () => Action.Submit({ value: state.count })),
          Match.tag("Event", (input) => Action.NextFrame({ state: { count: state.count + input.value } })),
          Match.exhaustive
        )
      ),
    clear: () => Effect.succeed("")
  },
  {
    events: Queue.asDequeue(eventQueue)
  }
)
```

packages/effect/typetest/unstable/cli/Prompt.tst.ts
```diff
@@ -0,0 +1,85 @@
...
+        }
+      )
+    })
+
+    it("returns Prompt<Output>", () => {
+      const prompt = Prompt.custom(
+        0,
+        {
+          render: () => Effect.succeed(""),
+          process: (_input, state) => Effect.succeed({ _tag: "Submit" as const, value: state }),
```
IMax153: We are not doing anything currently here to ensure the type of `_input` is an untagged user input.

IMax153 (changes requested):

packages/effect/typetest/unstable/cli/Prompt.tst.ts
lloydrichards: There is an explicit expect in the "without events, process receives Terminal.UserInput" test above, but I've gone and added a `Match` against the output so there is also a type check against the types in action

IMax153 (dismissed):

PR #2210, opened 2026-05-18
EFF-807 Add Stream.broadcastN

tim-smart:
## Summary
- Add `Stream.broadcastN`, a fixed-size broadcast API ported from Effect v3 `Stream.broadcast`
- Pre-subscribe downstream streams before starting upstream and unsubscribe individual streams on exit
- Publish `Take` values, including final exits, so downstream streams observe completion and failure
- Address review feedback by exposing `broadcastN` with a single options object (for example `{ n: 2, capacity: 8 }`) and by using a `makePubSub` helper copied from `Channel.ts`
- Add Stream tests for fan-out and failure propagation plus a changeset

## Validation
- `pnpm lint-fix`
- `pnpm test Stream.test.ts`
- `pnpm check:tsgo`

### DIFF
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -8513,7 +8513,7 @@ const makePubSub = <A>(
  * the same elements as the source stream.
  *
  * The source stream starts after all downstream streams have been subscribed.
- * With the default suspend strategy, the source can only advance `maximumLag`
+ * With the default suspend strategy, the source can only advance `capacity`
  * chunks ahead of the slowest downstream stream. If a downstream stream is
  * interrupted, it unsubscribes from the broadcast so it no longer contributes
  * backpressure.
@@ -8526,7 +8526,7 @@ const makePubSub = <A>(
  * const program = Effect.scoped(
  *   Effect.gen(function*() {
  *     const [left, right] = yield* Stream.make(1, 2, 3).pipe(
- *       Stream.broadcastN(2, 8)
+ *       Stream.broadcastN({ n: 2, capacity: 8 })
  *     )
  *
  *     const values = yield* Effect.all([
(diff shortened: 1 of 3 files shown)


packages/effect/src/Stream.ts
```diff
@@ -8486,6 +8487,111 @@ export const aggregateWithin: {
     )
   }))))
 
+type BroadcastNOptions = {
```
tim-smart: Remove the type helper and inline the types

packages/effect/src/Stream.ts
```diff
@@ -8486,6 +8487,111 @@ export const aggregateWithin: {
...
+  readonly capacity: "unbounded"
+  readonly replay?: number | undefined
+} | {
+  readonly capacity: number
+  readonly strategy?: "sliding" | "dropping" | "suspend" | undefined
+  readonly replay?: number | undefined
+}
+
+const makeBroadcastNPubSub = <A, E>(
```
tim-smart: This can copy the `makePubSub` function from `Channel.ts`

packages/effect/src/Stream.ts
```diff
@@ -8552,21 +8548,44 @@ const makeBroadcastNPubSub = <A, E>(
...
     n: N,
-    maximumLag: number | BroadcastNOptions
+    maximumLag: number | {
+      readonly capacity: "unbounded"
+      readonly replay?: number | undefined
+    } | {
+      readonly capacity: number
+      readonly strategy?: "sliding" | "dropping" | "suspend" | undefined
+      readonly replay?: number | undefined
+    }
```
tim-smart: I think the options should be merged into object, so `n` and `maximumLag` become a single `options` parameter.

PR #2217, opened 2026-05-18
Fix sqlite-do Durable Object transactions

### LINKED ISSUE #2216: @effect/sql-sqlite-do: back withTransaction by DurableObjectStorage.transaction
## Attribution

Opened by Pi (GPT-5.5 high), guided by Dillon Mulroy <dillon@cloudflare.com>.

Related v3 issue: https://github.com/Effect-TS/effect/issues/5987

## Repro

```ts
import { SqliteClient, SqliteMigrator } from "@effect/sql-sqlite-do"
import * as Reactivity from "effect/unstable/reactivity/Reactivity"

export class MyDurableObject extends DurableObject {
  constructor(ctx: DurableObjectState, env: Env) {
    super(ctx, env)

      const sql = yield* SqliteClient.make({ db: ctx.storage.sql })

      yield* SqliteMigrator.run({
        loader: SqliteMigrator.fromRecord({
          "1_init": sql`CREATE TABLE links (slug text primary key, href text not null)`
        })
      }).pipe(Effect.provideService(SqliteClient.SqliteClient, sql))
    }).pipe(Effect.scoped, Effect.provide(Reactivity.layer))

    ctx.blockConcurrencyWhile(() => Effect.runPromise(program))
  }
}
```

Workerd fails because `SqliteMigrator.run(...)` calls `sql.withTransaction(...)`, and sqlite-do's inherited `withTransaction` emits transaction SQL:

```txt
BEGIN
SAVEPOINT effect_sql_1
ROLLBACK TO SAVEPOINT effect_sql_1
COMMIT
ROLLBACK
```

Cloudflare Durable Object SQLite rejects transaction SQL:

```txt
sql.exec() cannot execute transaction-related statements like BEGIN TRANSACTION or SAVEPOINT.
Use ctx.storage.transaction() or ctx.storage.transactionSync() instead.
```

## Current v4 source

```ts
// packages/effect/src/unstable/sql/Migrator.ts
const completed = yield* pipe(
  sql.withTransaction(run),
  Effect.catchTag("MigrationError", (_) =>
    _.kind === "Locked"
      ? Effect.as(Effect.logDebug(_.message), [])
      : Effect.fail(_))
)
```

```ts
// packages/sql/sqlite-do/src/SqliteMigrator.ts
export const run = Migrator.make({})
```

```ts
// packages/effect/src/unstable/sql/SqlClient.ts
const beginTransaction = options.beginTransaction ?? "BEGIN"
const commit = options.commit ?? "COMMIT"
const savepoint = options.savepoint ?? ((name: string) => `SAVEPOINT ${name}`)
const rollback = options.rollback ?? "ROLLBACK"
const rollbackSavepoint = options.rollbackSavepoint ?? ((name: string) => `ROLLBACK TO SAVEPOINT ${name}`)
```

## Why the existing hooks are not enough

`beginTransaction`, `commit`, `rollback`, `savepoint`, `rollbackSavepoint`, `transactionAcquirer`, and `transactionService` model split-phase transactions.

DO SQLite needs callback-scoped transactions:

```ts
ctx.storage.transaction(() => runEffectSqlInsideCallback())
// or sync-only:
ctx.storage.transactionSync(() => runSyncSqlInsideCallback())
```

This is unsafe:

```ts
SqliteClient.make({
  db: ctx.storage.sql,
  beginTransaction: "SELECT 1",
  commit: "SELECT 1",
  rollback: "SELECT 1",
  savepoint: () => "SELECT 1",
  rollbackSavepoint: () => "SELECT 1"
})
```

because `Migrator.ts` inserts rows into the migrations table before running migration effects. A no-op transaction can leave a failed migration recorded as complete.

## Proposed shape

Make sqlite-do transactions storage-backed when storage is provided; keep `db`-only query compatibility.

```ts
import type { DurableObjectStorage, SqlStorage } from "@cloudflare/workers-types"

export interface SqliteClientConfig {
  readonly db?: SqlStorage
  readonly storage?: DurableObjectStorage
  readonly spanAttributes?: Record<string, unknown>
  readonly transformResultNames?: (str: string) => string
  readonly transformQueryNames?: (str: string) => string
}

const db = options.storage?.sql ?? options.db
```

Then override the returned client like `libsql` / `mssql` already do:

```ts
const base = yield* Client.make({ acquirer, compiler, transactionAcquirer, spanAttributes, transformRows })

return Object.assign(base, {
  [TypeId]: TypeId,
  config: options,
  withTransaction: options.storage === undefined
    ? unsupportedDoSqliteTransaction
    : storageBackedWithTransaction({
        storage: options.storage,
        connection,
        transactionService: base.transactionService,
        spanAttributes
      })
})
```

Implementation requirement: preserve Effect failure semantics while forcing Cloudflare rollback.

```ts
const Rollback = Symbol("@effect/sql-sqlite-do/Rollback")

// sketch, not exact API
const storageBackedWithTransaction =
  <R, E, A>(effect: Effect.Effect<A, E, R>): Effect.Effect<A, E | SqlError, R> =>
    Effect.withFiber((fiber) => {
      const services = fiber.context

      if (Context.getOption(services, transactionService)._tag === "Some") {
        return unsupportedNestedTransaction
      }

      let exit: Exit.Exit<A, E | SqlError> | undefined

      return Effect.tryPromise({
        try: () => storage.transaction(async () => {
          exit = await Effect.runPromiseExit(
            Effect.provideContext(
              effect,
              Context.add(services, transactionService, [connection, 0] as const)
            )
          )
          if (Exit.isFailure(exit)) throw Rollback
          return exit.value
        }),
        catch: (cause) => cause === Rollback ? Rollback : new SqlError({ reason: classifyError(cause, "Failed transaction", "transaction") })
      }).pipe(
        Effect.flatMap(() => exit === undefined ? Effect.die("missing transaction exit") : exit)
      )
    })
```

## Nested transactions

Do not emit savepoints for DO SQLite.

Safe default:

```ts
if (alreadyInTransaction) {
  return Effect.fail(new SqlError({
    reason: new UnknownError({
      operation: "transaction",
      message: "Nested transactions are not supported by Cloudflare Durable Object SQLite",
      cause: undefined
    })
  }))
}
```

Reason: Cloudflare docs do not expose savepoint-equivalent semantics for nested `storage.transaction(...)`. A nested no-op would break normal `withTransaction` savepoint semantics when an inner transaction fails and is caught by the outer transaction.

## Acceptance

```ts
yield* sql.withTransaction(
  Effect.gen(function*() {
    yield* sql`INSERT INTO t VALUES ('a')`
    return yield* Effect.fail("rollback")
  })
).pipe(Effect.ignore)

const rows = yield* sql`SELECT * FROM t`
// rows === []
```

```ts
yield* SqliteMigrator.run({ loader })
// works in Cloudflare Durable Object SQLite when SqliteClient was constructed with storage
```

dmmulroy:
Closes #2216

Author: Pi (GPT 5.5 on high)
Guided by: Dillon Mulroy <dillon@cloudflare.com>

## Changes
- accept `storage?: DurableObjectStorage` in `SqliteClientConfig`
- route `withTransaction` through `storage.transaction(...)`
- fail db-only / nested transactions without emitting `BEGIN` / `SAVEPOINT`
- add sqlite-do transaction + migrator regression tests

## Validation
- `pnpm test packages/sql/sqlite-do/test/Client.test.ts` — blocked: `vitest` not installed
- `pnpm lint-fix` — blocked: `oxlint` not installed
- `pnpm check:tsgo` — blocked: local tool/build-mode issue

### DIFF
--- a/.changeset/sqlite-do-durable-object-transactions.md
+++ b/.changeset/sqlite-do-durable-object-transactions.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-sqlite-do": patch
+---
+
+Support Cloudflare Durable Object SQLite transactions by allowing `SqliteClient` to be configured with `DurableObjectStorage` and routing `withTransaction` through `storage.transaction`.
--- a/packages/sql/sqlite-do/src/SqliteClient.ts
+++ b/packages/sql/sqlite-do/src/SqliteClient.ts
@@ -85,6 +85,10 @@ export interface SqliteClient extends Client.SqlClient {
  */
 export const SqliteClient = Context.Service<SqliteClient>("@effect/sql-sqlite-do/SqliteClient")
 
+const SqliteTransaction = Context.Service<Client.TransactionConnection, Client.TransactionConnection.Service>(
+  "@effect/sql-sqlite-do/SqliteClient/SqliteTransaction"
+)
+
 /**
  * Configuration for a Cloudflare Durable Object SQLite client, including either a `SqlStorage` handle or the full `DurableObjectStorage` for transaction support, span attributes, and query/result name transforms.
  *
@@ -115,14 +119,13 @@ const makeUnsupportedWithTransaction = (message: string): Client.SqlClient["with
 
 const makeStorageBackedWithTransaction = (
   storage: DurableObjectStorage,
-  transactionService: Client.SqlClient["transactionService"],
   connection: Connection,
   semaphore: Semaphore.Semaphore
 ): Client.SqlClient["withTransaction"] =>
 <R, E, A>(effect: Effect.Effect<A, E, R>): Effect.Effect<A, E | SqlError, R> =>
   Effect.withFiber((fiber) => {
     const services = fiber.context
-    const connOption = Context.getOption(services, transactionService)
+    const connOption = Context.getOption(services, SqliteTransaction)
     if (connOption._tag === "Some") {
       return Effect.fail(
         unsupportedTransaction(
@@ -139,7 +142,7 @@ const makeStorageBackedWithTransaction = (
             Effect.runPromiseExit(
               Effect.provideContext(
                 effect,
-                Context.add(services, transactionService, [connection, 0] as const)
+                Context.add(services, SqliteTransaction, [connection, 0] as const)
               )
             ).then((exit) => {
               if (Exit.isFailure(exit)) {
@@ -270,6 +273,7 @@ export const make = (
       acquirer,
       compiler,
       transactionAcquirer,
+      transactionService: SqliteTransaction,
       spanAttributes: [
         ...(options.spanAttributes ? Object.entries(options.spanAttributes) : []),
         [ATTR_DB_SYSTEM_NAME, "sqlite"]
@@ -281,7 +285,7 @@ export const make = (
       [TypeId]: TypeId as TypeId,
       config: options,
       withTransaction: options.storage
-        ? makeStorageBackedWithTransaction(options.storage, client.transactionService, connection, semaphore)
+        ? makeStorageBackedWithTransaction(options.storage, connection, semaphore)
         : makeUnsupportedWithTransaction(
           "Transactions require Durable Object storage; pass ctx.storage as the storage option"
         )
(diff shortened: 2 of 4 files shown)


tim-smart:
It feels like there is a cleaner way of doing this. Will take a look when I have a moment to look at the cf docs.

dmmulroy:
happy to refactor or change the impl - anything in particular you want changed?

PR #2224, opened 2026-05-18
fix(ai-openai): widen OpenAIFile.expires_at and status_details to accept null

aniravi24:
## Type

Bug Fix

## Description

`OpenAIFile.expires_at` and `OpenAIFile.status_details` are declared in the generated schema as `Schema.optionalKey(T)` (key may be omitted, but if present must be `T`). In practice, OpenAI's `POST /v1/files` response returns these fields with literal `null` values — at minimum for uploads with `purpose: "user_data"` that carry no expiration policy / no status detail. This makes the schema reject valid responses with:

```
SchemaError: Expected number, got null at ["expires_at"]
SchemaError: Expected string, got null at ["status_details"]
```

Any caller of `OpenAiClient.createFile` (or `retrieveFile` / `listFiles`) is affected — the decode fails even though the HTTP upload itself succeeded.

The mismatch is in the upstream Stainless-hosted OpenAPI doc (these fields are marked optional but not nullable, while the API runtime returns null). Since `Generated.ts` is produced via `effect-ai-codegen` against that doc, and `codegen.yaml` already carries similar overrides for known field-level nullability discrepancies (see the existing `ModelResponseProperties.user` patch and others), this PR follows the same pattern: two JSON-Patch overrides in `packages/ai/openai/codegen.yaml`.

Notes on the patch shape:

- `status_details` uses `nullable: true` (the straightforward case for a `string`).
- `expires_at` uses an `anyOf` against `{type: "null"}` rather than `nullable: true`, because the field also carries `format: unixtime` — feeding that through `JsonSchema.apply_nullable` produces a `type: [integer, null]` array that the generator collapses to `Schema.optionalKey(Schema.Never)` rather than a usable union. The `anyOf` form round-trips cleanly and matches what `AdminApiKey.last_used_at` already does in the same spec.

After regeneration the standalone `OpenAIFile` and its inline copy nested in `Upload` both emit `Schema.optionalKey(Schema.Union([..., Schema.Null]))`, and the downstream type aliases (`CreateFile200`, `RetrieveFile200`, `ListFiles200`, `ListFilesResponse.data`) pick it up structurally.

`Generated.ts` is intentionally not committed here — keeping the diff small, and the nightly `chore/ai-codegen-update` PR will refresh it (matches the convention used by similar codegen-yaml-only fixes).

## Related

n/a

### DIFF
--- a/packages/ai/openai/codegen.yaml
+++ b/packages/ai/openai/codegen.yaml
@@ -16,6 +16,8 @@ patches:
   - '[{"op":"remove","path":"/components/schemas/WebSearchActionSearch/required/1"}]'
   - '[{"op":"add","path":"/components/schemas/ModelResponseProperties/properties/prompt_cache_retention/anyOf/0/enum/1","value":"in_memory"}]'
   - '[{"op":"add","path":"/components/schemas/PromptCacheRetentionEnum/enum/1","value":"in-memory"}]'
+  - '[{"op":"replace","path":"/components/schemas/OpenAIFile/properties/expires_at","value":{"anyOf":[{"type":"integer","format":"unixtime","description":"The Unix timestamp (in seconds) for when the file will expire."},{"type":"null"}]}}]'
+  - '[{"op":"add","path":"/components/schemas/OpenAIFile/properties/status_details/nullable","value":true}]'
   # Add missing keepalive stream event support
   - '[{"op":"add","path":"/components/schemas/ResponseKeepAliveEvent","value":{"type":"object","title":"Keep alive","description":"A keepalive event emitted during long-running response streams.","properties":{"type":{"type":"string","enum":["keepalive"],"description":"The type of the keepalive event. Always `keepalive`.","default":"keepalive","x-stainless-const":true},"sequence_number":{"type":"integer","description":"The sequence number of this keepalive event."}},"required":["type","sequence_number"]}}]'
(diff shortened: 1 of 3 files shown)


packages/ai/openai/codegen.yaml
```diff
@@ -16,6 +16,14 @@ patches:
   - '[{"op":"remove","path":"/components/schemas/WebSearchActionSearch/required/1"}]'
   - '[{"op":"add","path":"/components/schemas/ModelResponseProperties/properties/prompt_cache_retention/anyOf/0/enum/1","value":"in_memory"}]'
   - '[{"op":"add","path":"/components/schemas/PromptCacheRetentionEnum/enum/1","value":"in-memory"}]'
+  # OpenAIFile.expires_at and OpenAIFile.status_details are returned as literal null
```
IMax153: Please remove these comments.

IMax153 (changes requested):
@aniravi24 - you need to re-run the code generation to regenerate the schemas.

This can be done with:

```ts
node ./packages/tools/ai-codegen/src/bin.ts generate
```

aniravi24:
@IMax153 seems to be a fairly large diff, let me know if that's what you want!

IMax153:
> @IMax153 seems to be a fairly large diff, let me know if that's what you want!

Yea that's fine - can you verify if the changes you need are included in the updated `Generated.ts` _without_ the patch? If so, we can remove the patch.

aniravi24:
@IMax153 without the patch, the fields are still showing up as optional when at runtime they can be nullable. It's still required.

IMax153 (approved):

PR #2230, opened 2026-05-19
perf(runtime): per-instance [evaluate] dispatch, zero module-level mutable state

front-depiction:
## Summary

The Effect run-loop's primary dispatch site at `internal/effect.ts` (the `current[evaluate](this)` call inside `FiberImpl.runLoop`) walks ~14 distinct primitive types per running fiber. The current implementation routes that dispatch through:

1. A shared `[evaluate]` method on `SharedComputationProto` that does `switch (this[opTag]) { ... }`,
2. Each case loading from one of 14 **module-level mutable `let` bindings** (`EVAL_Sync`, `EVAL_OnSuccess`, etc.),
3. Invoked via `.call(this, fiber)`.

This PR collapses that machinery. Each Primitive carries its evaluator function as an **own-property** set at construction time. The shared protos become bare `{ ...EffectProto }`. The 14 `EVAL_*` `let` bindings, the `registerComputationOp` dispatcher, the `registerStackPushEvaluators` cross-module init hook, and the giant switch all go away.

The Effect runtime now has **zero module-level mutable dispatch state**. Every evaluator reference flows by parenthood (closure capture) from the factory that creates it.

## Why this is a perf win (it's actually about the JIT, not the IC)

The intuitive framing — "kill the megamorphic call IC at the dispatch site" — turned out wrong in measurement. JSC's polymorphic call IC handles ≤14 distinct call targets cheaply via specialized stubs. The real cost was elsewhere:

- **Module-level `let` bindings defeat JSC's constant folding.** Every `EVAL_OnSuccess.call(...)` inside the switch had to re-read the mutable binding. JSC cannot prove the value is stable, so it cannot fold the call target.
- **Per-instance own-property evaluator reads a closed-over `const`-equivalent.** JSC can fold this at JIT compile time — the function reference is bound at factory definition and never mutates.
- **Three stack-push ops (`OnSuccess`, `OnFailure`, `OnSuccessAndFailure`) have identical evaluator bodies.** Sharing one function reference across them tightens dispatch-site polymorphism from 14 distinct targets to 12.

The OCAP-style framing isn't a side benefit — eliminating the ambient mutable registry is what unlocks the JIT optimization. Capability discipline and JIT-friendliness pointed at the same fix.

## Benchmark (n=300 baseline pooled, n=90 candidate pooled, quiesced Apple M4 Pro, Bun 1.3.12, Welch's two-tailed t-test α=0.05)

| Scenario | Δ% | p |
|---|---:|---:|
| succeed-flatMap chain (depth 1000) | **-39.12%** | 2e-150 |
| mixed-primitive chain (flatMap/map/tap × 250) | **-24.49%** | 4e-80 |
| fork-join 100 fibers | **-19.35%** | 5e-37 |
| Stream range + map + filter + drain | **-14.31%** | 2e-33 |
| fail + catchCause roundtrip | **-9.85%** | 2e-34 |
| service provide + yield (shallow) | **-4.69%** | 8e-8 |
| service provide + yield (deep) | **-30.80%** | 2e-104 |
| Layer build deep (6-service graph) | **-12.92%** | 4e-59 |
| Representative program (services + Stream + 16 fork/join + catchCause) | **-14.04%** | 2e-47 |
| Effect AST allocator (deep tree, no run) | -0.87% | 0.24 (NEUTRAL) |

**9 WINS / 1 NEUTRAL / 0 LOSSES.**

The executing-workload heap profile is essentially unchanged from main.

## Parity

`pnpm -F effect test` produces `5972 pass / 7 fail / 5 skip` — exact match to main at the time this branch was cut (`8d71e54`).

`pnpm check:tsgo` clean.

Public API completely unchanged. Observable behavior unchanged — only the internal node representation and dispatch machinery moved.

## Test plan

- [x] `pnpm check:tsgo` clean
- [x] `pnpm -F effect test` matches baseline (5972/7/5)
- [x] Bench at n=300 pooled samples on a quiesced machine confirms wins on 9 of 10 scenarios with no significant regression
- [x] Bench scenarios include realistic workloads (Layer build, representative mixed program) not just microbenches
- [ ] Reviewer to confirm: any test or downstream package that may rely on the now-removed `registerStackPushEvaluators` export from `internal/core.ts`? (Internal export only; no public-API uses found in this repo.)

## What changed (file inventory)

- `packages/effect/src/internal/core.ts` — unified `SharedComputationProto` / `SharedExitProto` (now bare `...EffectProto` with no `[evaluate]` on the proto); `makePrimitive`, `makeOnSuccess`, `makeOnFailure`, `makeOnSuccessAndFailure`, `exitSucceed`, `exitFailCause` all set `[evaluate]` as the last own-property at construction; `EVAL_*` lets, `registerComputationOp`, `registerStackPushEvaluators` removed; the three stack-push ops share one `evaluateStackPush` function reference.
- `packages/effect/src/internal/effect.ts` — three local `evaluateOnSuccess` / `evaluateOnFailure` / `evaluateOnSuccessAndFailure` definitions removed; `registerStackPushEvaluators` import + call removed.
- `.changeset/perf-runtime-zero-mutable-dispatch.md` — patch changeset.

Net diff is small — both files shrank. The bulk of the change is *removal*.

## Notes for reviewers

- Instance property add-order is held canonical across all 14 internal computation primitives so `SharedComputationProto` retains a single hidden class: `[identifier], [opTag], [args], [contA], [contE], [contAll], [evaluate]`. Adding new ops in the future must preserve this order.
- `core.ts` already imports `FiberImpl` as a type-only import; the evaluator's runtime access to `fiber._stack` is duck-typed, so no runtime cross-module dependency is introduced.
- The bench harness used to measure this (under `bench/` in this branch) follows the discipline in `bun-benchmarking` style: warmup, `Bun.gc(true)` between runs, raw samples preserved for Welch's t-test, heap delta + object count captured. Happy to discuss the methodology or hand the harness over if useful upstream.
- Currently only improves in JSC,  likely megamorphism issues in V8

### DIFF
--- a/.changeset/perf-runtime-zero-mutable-dispatch.md
+++ b/.changeset/perf-runtime-zero-mutable-dispatch.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+perf(runtime): per-instance [evaluate] dispatch via shared protos; zero module-level mutable dispatch state. Closes a JIT-friendliness gap in the run-loop's primary dispatch site: evaluator references now flow by closure capture (parenthood) instead of through a module-level mutable registry, enabling JSC to const-fold the call target. -39% on succeed-flatMap chains, -31% on deep service-yield, -19% on fork-join, with 9/10 benchmark scenarios improving and no regressions at n=300 pooled samples.
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ scratchpad/**/*
 
 # Repositories
 .repos/
+bench/results/
--- a/bench/compare-pooled.ts
+++ b/bench/compare-pooled.ts
@@ -1,0 +1,364 @@
+/**
+ * Pooled Welch's two-tailed t-test comparison.
+ *
+ * Each side aggregates N JSON bench files (each containing RUNS=30 post-warmup
+ * samples). The N files are concatenated per scenario into a single pooled
+ * sample vector, then Welch's t-test is run on the pooled vectors.
+ *
+ * Also emits a self-consistency check: pairwise compare every (baseline_i,
+ * baseline_j) and flag any pair with p<0.05 AND |Δ|>=3%. That tells the user
+ * whether the machine was quiet across runs.
+ *
+ * Usage:
+ *   bun run bench/compare-pooled.ts \
+ *     --label-base baseline --base FILE [FILE ...] \
+ *     --label-cand candidate --cand FILE [FILE ...] \
+ *     [--out comparison.json]
+ */
+
+import * as fs from "node:fs"
+
+interface Stats {
+  min: number
+  max: number
+  mean: number
+  median: number
+  p95: number
+  p99: number
+  stddev: number
+  rsd: number
+  opsPerSec: number
+}
+
+interface Scenario {
+  name: string
+  iterationsPerRun: number
+  samples: Array<number>
+  stats: Stats
+  heap: { heapDeltaMean: number; objectCountDeltaMean: number }
+}
+
+interface BenchFile {
+  scenarios: Array<Scenario>
+}
+
+// --- statistics ------------------------------------------------------------
+
+const betacf = (a: number, b: number, x: number): number => {
+  const MAXIT = 200
+  const EPS = 3e-7
+  const FPMIN = 1e-30
+  const qab = a + b
+  const qap = a + 1
+  const qam = a - 1
+  let c = 1
+  let d = 1 - qab * x / qap
+  if (Math.abs(d) < FPMIN) d = FPMIN
+  d = 1 / d
+  let h = d
+  for (let m = 1; m <= MAXIT; m++) {
+    const m2 = 2 * m
+    let aa = m * (b - m) * x / ((qam + m2) * (a + m2))
+    d = 1 + aa * d
+    if (Math.abs(d) < FPMIN) d = FPMIN
+    c = 1 + aa / c
+    if (Math.abs(c) < FPMIN) c = FPMIN
+    d = 1 / d
+    h *= d * c
+    aa = -(a + m) * (qab + m) * x / ((a + m2) * (qap + m2))
+    d = 1 + aa * d
+    if (Math.abs(d) < FPMIN) d = FPMIN
+    c = 1 + aa / c
+    if (Math.abs(c) < FPMIN) c = FPMIN
+    d = 1 / d
+    const del = d * c
+    h *= del
+    if (Math.abs(del - 1) < EPS) break
+  }
+  return h
+}
+
+const lgamma = (x: number): number => {
+  const cof = [
+    76.18009172947146,
+    -86.50532032941677,
+    24.01409824083091,
+    -1.231739572450155,
+    0.1208650973866179e-2,
+    -0.5395239384953e-5
+  ]
+  let y = x
+  let tmp = x + 5.5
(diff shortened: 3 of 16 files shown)


front-depiction:
## Cross-engine measurement results — converting to research note

Marking this as draft. After opening the PR I ran the same bench harness against Node 24 / V8 (—expose-gc) and the win does **not** generalize. Posting the negative result here for the record.

### Headline result

Same change, two engines, n=90 pooled per condition, Welch's two-tailed t-test α=0.05:

| Scenario | Bun (JSC) | **Node (V8)** |
|---|---:|---:|
| 01-succeed-flatmap-1000 | **-39.12%** | **+64.38%** |
| 02-mixed-primitive-chain | **-24.49%** | **+60.31%** |
| 03-fork-join-100 | **-19.35%** | +6.43% |
| 04-stream-range-map-filter-drain | **-14.31%** | -0.23% |
| 05-fail-catchCause-roundtrip | **-9.85%** | +0.07% |
| 06a-service-provide-yield (shallow) | **-4.69%** | **-13.84%** |
| 06b-service-provide-yield (deep) | **-30.80%** | **-16.03%** |
| 07-layer-build-deep | **-11.24%** | **+13.17%** |
| 08-representative-program | **-14.04%** | **+13.72%** |
| 09-effect-allocator | -0.87% | **+87.71%** |

JSC: 9 WIN / 1 NEUTRAL / 0 LOSS. V8: 3 WIN / 2 NEUTRAL / 5 LOSS, with the largest hot-path regressions reaching +60–87%.

### Diagnosis

The dispatch site `current[evaluate](this)` in this PR sees ~12 distinct call targets (one per primitive type, since `[evaluate]` is now an instance own-property closed over by each factory). JSC's polymorphic call IC handles ≤12 targets gracefully (per webkit.org/blog/10308); V8's call IC goes megamorphic at >4 (per mrale.ph/blog/2015 / mathiasbynens.be/notes/shapes-ics).

I then tried a follow-up variant (V1.6, branch `candidate-V1.6-design-b` in the same fork) that kept the monomorphic outer dispatch (single shared `[evaluate]` switch on `SharedComputationProto`) while eliminating module-level `let` bindings via closure-captured `const` evaluator references. **Same V8 regression pattern** — 01 +54.84%, 02 +51.92%, 09 +72.94%. That falsifies the hypothesis that the JSC win was about JIT const-folding; the win is structurally tied to JSC's IC handling, not portable.

### What this means

The existing main-branch design (15 distinct per-primitive prototypes; 14 module-level `let EVAL_*` bindings + switch dispatch in `SharedComputationProto`; per-instance own-property continuations) is **V8-optimal** for this codebase. The module-level mutable bindings aren't a code smell from V8's perspective — TurboFan profile-specializes them after warmup, and the resulting dispatch is faster than any closure-captured-const alternative we could find. There is a real JSC-only win (10-40% on hot paths) but capturing it would symmetrically regress V8 users, who are the majority of Effect's production base.

### Reusable artifacts

If useful upstream:
- `bench/` directory: cross-runtime harness (Bun + Node 24 `--experimental-strip-types`), pooled Welch comparator, 10 scenarios (incl. layer-heavy + representative program). Methodology mirrors `@effect-atom/atom`'s bench discipline.

Happy to remove anything that's not useful or close the PR entirely. Posting as a research note in case the JSC/V8 IC asymmetry on per-op dispatch is interesting context for future runtime work.

tim-smart:
There is quite a big slow down on node.js unfortunately.

PR #2246, opened 2026-05-20
fix(schema): preserve named-field decoded value on StructWithRest key overlap

### LINKED ISSUE #2245: StructWithRest: index signatures shadow named-field decoders at runtime

4.0.0-beta.69

In Schema.StructWithRest, a Schema.Record index signature whose key domain overlaps with a named property signature silently overwrites the named decoder's result when the key is present in input. The decoded TypeScript type continues to advertise the named field's type, so consumers code against a runtime shape that violates the type contract.

#### Repro

```ts
const S = Schema.StructWithRest(
   Schema.Struct({ count: Schema.NumberFromString }),
   [Schema.Record(Schema.String, Schema.String)]
 )
 Schema.decodeUnknownSync(S)({ count: "42" })
 // → { count: "42" }    ← raw string, expected number 42
```

decode({ count: "42" }) should return `{ count: 42 }`.

### What do you see instead?

decode({ count: "42" }) returns `{ count: "42" }` .

### Additional information

_No response_

Leka74:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When a `Schema.StructWithRest` index signature's key domain overlapped a named property signature, the index signature's decoded value overwrote the property signature's transformed value. The declared type narrowed via intersection but the runtime did not.

```ts
const S = Schema.StructWithRest(
  Schema.Struct({ count: Schema.NumberFromString }),
  [Schema.Record(Schema.String, Schema.Unknown)]
)

Schema.decodeUnknownSync(S)({ count: "42" })
// before: { count: "42" }  (raw string)
// after:  { count: 42 }    (transformed number)
```

The index signature still runs for overlapping keys, its refinement checks (e.g. `Schema.Finite`) still apply and surface errors  but its decoded value is no longer written when a property signature has already produced one.

disclaimer: fix applied with Opus 4.7

## Related

- Closes #2245

### DIFF
--- a/.changeset/fix-structwithrest-shadow.md
+++ b/.changeset/fix-structwithrest-shadow.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `Schema.StructWithRest` index signatures overwriting named-field decoded values on overlapping keys.
--- a/packages/effect/src/SchemaAST.ts
+++ b/packages/effect/src/SchemaAST.ts
@@ -1763,10 +1763,10 @@ export class Objects extends Base {
         readonly options: ParseOptions
         readonly out: Record<PropertyKey, unknown>
         issues: Array<Issue.Issue> | undefined
-      }, [key: PropertyKey, is: IndexSignature]>()({
+      }, [key: PropertyKey, is: IndexSignature, validateOnly: boolean]>()({
         onItem: Effect.fnUntracedEager(function*(
           s,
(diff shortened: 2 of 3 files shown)


gcanti:
Closing in favor of https://github.com/Effect-TS/effect-smol/pull/2312

PR #2247, opened 2026-05-20
fix(cli): use predicate dual for Argument.variadic to support no-options pipe form

### LINKED ISSUE #6228: Argument.variadic without options parses to an object instead of an array

effect@4.0.0-beta.59

 ```ts
   import { BunRuntime, BunServices } from "@effect/platform-bun"
   import { Argument, Command } from "effect/unstable/cli"
   import * as Effect from "effect/Effect"

   const bad = Command.make(
     "bad",
     {
       files: Argument.string("file").pipe(Argument.variadic),
     },
     ({ files }) =>
       Effect.sync(() => {
         console.log({ files, isArray: Array.isArray(files) })
       })
   )

   const good = Command.make(
     "good",
     {
       files: Argument.string("file").pipe(Argument.variadic({})),
     },
     ({ files }) =>
       Effect.sync(() => {
         console.log({ files, isArray: Array.isArray(files) })
       })
   )

   const command = Command.make("repro").pipe(Command.withSubcommands([bad, good]))

   BunRuntime.runMain(
     Command.run(command, { version: "0.0.0" }).pipe(Effect.provide(BunServices.layer))
   )
 ```

 Run:

 ```sh
   bun repro.ts bad a b c
   bun repro.ts good a b c
 ```

Using `effect@4.0.0-beta.59`, the documented no-options form parses `files` as `{}` instead of an array. Passing an explicit empty options object parses correctly.

 ```ts
   { files: ["a", "b", "c"], isArray: true }
 ```

### What do you see instead?

 For:

 ```ts
   Argument.string("file").pipe(Argument.variadic)
 ```

 the output is:

 ```ts
   { files: {}, isArray: false }
 ```

 For:

 ```ts
   Argument.string("file").pipe(Argument.variadic({}))
 ```

 the output is:

 ```ts
   { files: ["a", "b", "c"], isArray: true }
 ```

### Additional information

 The installed implementation appears to export `variadic` as a two-argument dual combinator:

 ```ts
   export const variadic = dual(2, (self, options) => Param.variadic(self, options))
 ```

Since `Param.variadic` already handles `options` as optional, the exported wrapper likely needs to support the no-options form shown in the docs.

Zelys-DFKH:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Closes Effect-TS/effect#6228.

The no-options pipe form silently returns `{}` instead of an array:

```ts
Argument.string("files").pipe(Argument.variadic)
// files is {}, not string[]
```

`variadic` is declared as `dual(2, body)`. When `.pipe(fn)` calls `fn(self)` with one argument, `dual(2, ...)` treats it as the `options` argument and returns a partially-applied function. `self` never reaches the body.

Switching to the predicate overload fixes it:

```ts
// before
dual(2, body)

// after
dual((args) => Param.isParam(args[0]), body)
```

If the first argument is a `Param`, it's data-first. Otherwise, data-last. `Channel.ts`, `FiberHandle.ts`, and `Graph.ts` use this same pattern for the same reason.

All four call forms now work:

```ts
Argument.string("files").pipe(Argument.variadic)              // pipe, no options
Argument.string("files").pipe(Argument.variadic({ min: 1 }))  // pipe, with options
Argument.variadic(Argument.string("files"))                    // direct, no options
Argument.variadic(Argument.string("files"), { min: 1 })        // direct, with options
```

Added two regression tests, one per pipe form.

### DIFF
--- a/.changeset/fix-argument-variadic-pipe-form.md
+++ b/.changeset/fix-argument-variadic-pipe-form.md
@@ -3,26 +3,3 @@
 ---
 
 Fix `Argument.variadic` no-options pipe form returning `{}` instead of an array
-
-`Argument.variadic` was declared as `dual(2, ...)`, which requires two arguments for
-the data-first call form. When used as `.pipe(Argument.variadic)` (no options, no call),
-the piped value was incorrectly treated as the options argument, returning a partially-applied
-function instead of the expected `Argument<ReadonlyArray<A>>`.
-
-Replaced `dual(2, body)` with the predicate-based overload `dual((args) => Param.isParam(args[0]), body)`,
-which detects the data-first form by checking whether the first argument is a `Param`. All four
-call forms now work correctly:
-
-```ts
-// piped, no options
-Argument.string("files").pipe(Argument.variadic)
-
-// piped, with options
-Argument.string("files").pipe(Argument.variadic({ min: 1 }))
-
-// direct, no options
-Argument.variadic(Argument.string("files"))
-
-// direct, with options
-Argument.variadic(Argument.string("files"), { min: 1 })
-```
--- a/.changeset/fix-argument-variadic-pipe-form.md
+++ b/.changeset/fix-argument-variadic-pipe-form.md
@@ -1,0 +1,28 @@
+---
+"effect": patch
+---
+
+Fix `Argument.variadic` no-options pipe form returning `{}` instead of an array
+
+`Argument.variadic` was declared as `dual(2, ...)`, which requires two arguments for
+the data-first call form. When used as `.pipe(Argument.variadic)` (no options, no call),
+the piped value was incorrectly treated as the options argument, returning a partially-applied
+function instead of the expected `Argument<ReadonlyArray<A>>`.
+
+Replaced `dual(2, body)` with the predicate-based overload `dual((args) => Param.isParam(args[0]), body)`,
+which detects the data-first form by checking whether the first argument is a `Param`. All four
+call forms now work correctly:
+
+```ts
+// piped, no options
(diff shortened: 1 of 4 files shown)


packages/effect/test/unstable/cli/Arguments.test.ts
```diff
@@ -175,6 +175,46 @@ describe("Command arguments", () => {
       assert.deepStrictEqual(result.files, ["file1.txt", "file2.txt", "file3.txt"])
     }).pipe(Effect.provide(TestLayer)))
 
+  it("should handle variadic in pipe form without options", () =>
+    Effect.gen(function*() {
+      let result: { readonly files: ReadonlyArray<string> } | undefined
+
+      // Regression: dual(2,...) with optional arg treats the piped self as options, returning {} instead of an array.
+      const testCommand = Command.make("test", {
```
tim-smart:
```suggestion
      const testCommand = Command.make("test", {
```

.changeset/fix-argument-variadic-pipe-form.md
```diff
@@ -0,0 +1,28 @@
...
+
+// piped, with options
+Argument.string("files").pipe(Argument.variadic({ min: 1 }))
+
+// direct, no options
+Argument.variadic(Argument.string("files"))
+
+// direct, with options
+Argument.variadic(Argument.string("files"), { min: 1 })
+```
```
tim-smart:
```suggestion
Fix `Argument.variadic` no-options pipe form returning `{}` instead of an array
```

PR #2255, opened 2026-05-21
fix(effect): resume hang in iterateEager when paused-batch observer drains last

patroza:
introduced in #2010 `iterateEagerImpl` observers captured a per-`go()` `paused` local. Observers forked during a paused call always took the `if (paused)` branch and skipped the `else if (done && fibers.size === 0)` resume path. When the last fiber to complete carried `paused = true`, `resume!` was never invoked, leaving the outer fiber suspended on the async callback and retaining `parentFiber`, items, and schema state.

Split the observer into two independent checks and read-clear `resume` on use so calls are idempotent and the closure reference is released early. Apply the same read-clear pattern to the resume paths inside `go()`.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/test/Effect.test.ts
+++ b/packages/effect/test/Effect.test.ts
@@ -280,6 +280,26 @@ describe("Effect", () => {
         assert.deepStrictEqual(done, [1, 2, 3])
       }))
 
+    it.effect("bounded fail interrupts paused-batch fibers", () =>
+      Effect.gen(function*() {
+        const completed: Array<number> = []
+        const handle = yield* Effect.forEach([1, 2, 3, 4], (i) =>
+          Effect.suspend(() =>
+            i === 2
+              ? Effect.fail("boom").pipe(Effect.delay(50))
+              : Effect.sync(() => {
+                completed.push(i)
+                return i
+              }).pipe(Effect.delay(i === 1 ? 500 : 100))
+          ), {
+          concurrency: 2
+        }).pipe(Effect.forkChild)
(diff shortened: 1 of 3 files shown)


packages/effect/test/Effect.test.ts
```diff
@@ -280,6 +280,26 @@ describe("Effect", () => {
         assert.deepStrictEqual(done, [1, 2, 3])
       }))
 
+    it.effect("bounded fail interrupts paused-batch fibers", () =>
```
tim-smart, line 283: This test passes without the changes :)

patroza: yea, the leak also was not solved. it seemed to make sense at the time, all the data was supportive too.. but alas :)

The problem appears when using schemas who use resolvers with batching/caching where I manage a cache per request https://github.com/effect-app/libs/blob/6ee5b1707e63eb04340c4d79191425bcc9a5c6d3/packages/infra/src/Store/ContextMapContainer.ts#L33
It looks like the problem goes away when I use a global request resolver cache instead.
will share some dumps or repro if I catch it.

patroza: @tim-smart this global instead appears to be our problem :D https://github.com/Effect-TS/effect-smol/pull/2263

PR #2265, opened 2026-05-25
fix schema arbitrary constraint combiners for BigInt, Date, and BigDecimal

Gabrola:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Summary

Fix schema arbitrary constraint derivation for BigInt, Date, and BigDecimal checks.

- Merge constraint bounds with type-specific combiners instead of shared number-based reducers.
- Add BigDecimal arbitrary constraints and generate bigint bounds for each selected decimal scale.
- Skip BigDecimal scales whose rounded bounds cross before passing constraints to fast-check.
- Add regression coverage for BigInt, Date context handling, and BigDecimal rescaling.

## Why

The previous constraint merging path did not handle non-number bounds reliably, and BigDecimal generation reused decimal bounds without adapting them to the generated scale. That made some valid decimal values effectively unreachable and could produce invalid fast-check bigint ranges for narrow decimal intervals.

### DIFF
--- a/packages/effect/src/Schema.ts
+++ b/packages/effect/src/Schema.ts
@@ -7191,10 +7191,6 @@ export const isBetweenBigInt = makeIsBetween({
   })
 })
 
-const nextBigDecimal = (n: BigDecimal_.BigDecimal) => BigDecimal_.make(n.value + 1n, n.scale)
-
-const previousBigDecimal = (n: BigDecimal_.BigDecimal) => BigDecimal_.make(n.value - 1n, n.scale)
-
 /**
  * Validates that a BigDecimal is greater than the specified value (exclusive).
  *
@@ -7206,7 +7202,8 @@ export const isGreaterThanBigDecimal = makeIsGreaterThan({
   annotate: (exclusiveMinimum) => ({
     toArbitraryConstraint: {
       bigDecimal: {
-        min: nextBigDecimal(exclusiveMinimum)
+        min: exclusiveMinimum,
+        minExcluded: true
       }
     }
   }),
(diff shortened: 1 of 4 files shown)


packages/effect/src/Schema.ts
```diff
@@ -7058,8 +7057,7 @@ export const isGreaterThanBigInt = makeIsGreaterThan({
     },
     toArbitraryConstraint: {
       bigint: {
-        min: exclusiveMinimum,
-        minExcluded: true
```
Gabrola: fast-check doesn't have `minExcluded`/`maxExcluded` on bigints

gcanti: you are right, working on a fix that also includes dates and integers: https://github.com/Effect-TS/effect-smol/pull/2266

Gabrola: @gcanti that's perfect, I'll readjust my fix to match

packages/effect/test/schema/toArbitrary.test.ts
```diff
@@ -608,6 +608,42 @@ describe("Arbitrary generation", () => {
...
+    it("isBetweenBigDecimal", () => {
+      verifyGeneration(
+        Schema.BigDecimal.check(
+          Schema.isBetweenBigDecimal({
+            minimum: BigDecimal.make(100n, 0),
+            maximum: BigDecimal.make(200n, 0)
+          })
+        )
+      )
+    })
```
Gabrola, line 637: This one is a little peculiar. Before the fix, fast-check was only generating bigints between -77n and 77n, so this test was just running in an infinite loop

gcanti:
@Gabrola I've added a generic `ordered` constraint instead of specialized constraints for each sortable data type. Otherwise, we would need to introduce a new specialized constraint every time we want to support another sortable type.

PR #2276, opened 2026-05-26
EFF-817 Fix AtomRef notify listener resubscribe

### LINKED ISSUE #2275: AtomRef.notify skips listeners when a callback re-subscribes itself
`ReadonlyRefImpl` ([AtomRef.ts#L140-L161](https://github.com/Effect-TS/effect-smol/blob/main/packages/effect/src/unstable/reactivity/AtomRef.ts#L140-L161)) iterates `listeners` in `notify` while `subscribe` (push) and the unsubscribe returned by `subscribe` (swap-with-last + pop) mutate the same array. If a listener unsubscribes itself and re-subscribes from inside its callback — a common shape for derived/reactive abstractions that re-collect deps on each notification — the swap-pop reshuffles the array under `notify`'s loop. Some listeners get skipped, others fire twice.

### Repro

With `effect@4.0.0-beta.70`:

```ts
import { AtomRef } from "effect/unstable/reactivity"

const ref = AtomRef.make(0)
const calls = { A: 0, B: 0, C: 0 }

let unsubA = ref.subscribe(function A() {
  calls.A++
  unsubA()
  unsubA = ref.subscribe(A)
})
ref.subscribe(() => { calls.B++ })
ref.subscribe(() => { calls.C++ })

ref.set(1)
console.log(calls)
// Expected: { A: 1, B: 1, C: 1 }
// Actual:   { A: 2, B: 1, C: 0 }
```

### Trace

Initial `listeners = [A, B, C]`. `notify` enters at `i=0`, calls `A`. `A`'s self-unsubscribe swap-pops `C` into slot 0 → `[C, B]`. Its re-subscribe pushes a fresh `A` at the end → `[C, B, A]`. Loop continues to `i=1` (`B`, fires) and `i=2` (`A` again, fires + repeats the swap dance). `C` sits at index 0 the whole time but `i=0` is already behind us, so `C` never runs.

### Suggested fix

Snapshot at the start of `notify`:

```ts
notify(a: A) {
  const snap = this.listeners.slice(0, this.listenerCount)
  for (let i = 0; i < snap.length; i++) snap[i](a)
}
```

One-line change, matches the file's minimalist style. The same fragility transitively affects `MapRefImpl` / `PropRefImpl` (their `subscribe` registers a wrapper on the parent), so the parent-side fix covers them.

Alternative shapes if per-notify allocation is a concern: tombstone-and-compact (mark slots `null` on unsub, skip them during iteration, compact after), or a re-entrancy counter that defers mutations until `notify` finishes.

Happy to send a PR if it'd be welcome.

---

Found while debugging a UI framework built on `effect/unstable/reactivity` ([m9tdev/efx#14](https://github.com/m9tdev/efx/pull/14)). Two JSX expressions in the same component reading the same ref were enough to trigger it.

tim-smart:
## Summary
- Store AtomRef listeners in a Set and snapshot them at the start of notification so self re-subscription cannot mutate the active iteration.
- Keep once-per-notification semantics even when listeners unsubscribe and re-subscribe themselves; Set iteration is live, so notify still uses a snapshot.
- Add a regression test for a listener that unsubscribes and re-subscribes itself during notification.
- Add a patch changeset for effect.

## Validation
- pnpm test packages/effect/test/AtomRef.test.ts
- pnpm lint-fix
- pnpm check:tsgo (fails due existing typetest Effect.tst.ts missing OtherError errors at lines 323, 342, 417, 454)

### DIFF
--- a/packages/effect/src/unstable/reactivity/AtomRef.ts
+++ b/packages/effect/src/unstable/reactivity/AtomRef.ts
@@ -153,27 +153,21 @@ class ReadonlyRefImpl<A> implements ReadonlyRef<A> {
     return Hash.hash(this.value)
   }
 
-  listeners: Array<(a: A) => void> = []
-  listenerCount = 0
+  listeners: Set<{ readonly f: (a: A) => void }> = new Set()
 
   notify(a: A) {
-    const listeners = this.listeners.slice(0, this.listenerCount)
+    const listeners = Array.from(this.listeners)
     for (let i = 0; i < listeners.length; i++) {
-      listeners[i](a)
+      listeners[i].f(a)
     }
   }
 
   subscribe(f: (a: A) => void): () => void {
-    this.listeners.push(f)
-    this.listenerCount++
+    const listener = { f }
+    this.listeners.add(listener)
 
     return () => {
-      const index = this.listeners.indexOf(f)
-      if (index !== -1) {
-        this.listeners[index] = this.listeners[this.listenerCount - 1]
-        this.listeners.pop()
-        this.listenerCount--
-      }
+      this.listeners.delete(listener)
     }
   }
 
--- a/packages/effect/src/unstable/reactivity/AtomRef.ts
+++ b/packages/effect/src/unstable/reactivity/AtomRef.ts
@@ -157,8 +157,9 @@ class ReadonlyRefImpl<A> implements ReadonlyRef<A> {
   listenerCount = 0
 
   notify(a: A) {
-    for (let i = 0; i < this.listenerCount; i++) {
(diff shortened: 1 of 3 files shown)


packages/effect/src/unstable/reactivity/AtomRef.ts
tim-smart: Would using a Set instead work around this issue?

tim-smart:
Addressed the Set feedback by switching listener storage to a Set while keeping notify snapshotting via Array.from. Snapshotting is still needed because Set iteration is live and a listener that deletes/re-adds itself can otherwise be visited again in the same notification.

PR #2288, opened 2026-05-27
[codex] Add generic HttpApi HTTP security schemes

juliusmarminge:
## Summary

- add `HttpApiSecurity.http({ scheme })` for arbitrary HTTP authorization schemes
- express `HttpApiSecurity.bearer` and `HttpApiSecurity.dpop` as convenience declarations on the generic HTTP scheme
- validate and decode token68 credentials for HTTP authorization schemes and specialized `Basic` authentication
- emit and preserve arbitrary OpenAPI `{ type: "http", scheme }` security schemes in generated `HttpApi` modules
- cover generic runtime extraction, DPoP extraction, invalid/wrong-scheme credentials, OpenAPI output, generator output, and public API typing

## Why

`HttpApi` can currently extract DPoP authorization values only by declaring an API key in the `authorization` header. That exposes the raw header and documents the security scheme incorrectly in OpenAPI. DPoP is not the only HTTP authorization scheme that needs this capability, so the API now models HTTP authentication directly:

```ts
HttpApiSecurity.http({ scheme: "Signature" })
HttpApiSecurity.dpop // equivalent convenience declaration for scheme "DPoP"
HttpApiSecurity.bearer // equivalent convenience declaration for scheme "bearer"
```

HTTP authentication declarations must also enforce their declared scheme. Without validation, a `Bearer` authorization value can be sliced into an apparent DPoP credential because the prefixes have the same length. The decoder now validates the authentication scheme case-insensitively and requires token68 credential syntax before passing a value to middleware; invalid inputs yield the existing empty-credential result for middleware rejection.

`HttpApiSecurity.basic` remains specialized rather than being an alias for `http({ scheme: "basic" })`, because it decodes the authorization parameter into `{ username, password }`. Generic HTTP schemes expose their authorization parameter as a redacted value.

## Usage

A protected DPoP resource declares `HttpApiSecurity.dpop` once, then uses the same middleware to require the companion `DPoP` proof header and delegate RFC 9449 validation to an application service. The verification service implementation is intentionally omitted here: token validation, JWS verification, replay storage, nonce policy, and request URI policy belong to the application.

```ts
import { Context, Effect, Layer, Redacted, Schema } from "effect"
import { HttpServerRequest } from "effect/unstable/http"
import {
  HttpApi,
  HttpApiBuilder,
  HttpApiEndpoint,
  HttpApiError,
  HttpApiGroup,
  HttpApiMiddleware,
  HttpApiSecurity
} from "effect/unstable/httpapi"

interface RelayClient {
  readonly subject: string
}

class CurrentRelayClient extends Context.Service<CurrentRelayClient, RelayClient>()("CurrentRelayClient") {}

class VerifyDPoPProtectedResource extends Context.Service<VerifyDPoPProtectedResource, {
  readonly verify: (input: {
    readonly accessToken: Redacted.Redacted
    readonly proofJwt: Redacted.Redacted
    readonly request: HttpServerRequest.HttpServerRequest
  }) => Effect.Effect<RelayClient, HttpApiError.Unauthorized>
}>()("VerifyDPoPProtectedResource") {}

class RelayDPoPAuth extends HttpApiMiddleware.Service<RelayDPoPAuth, {
  provides: CurrentRelayClient
  requires: VerifyDPoPProtectedResource
}>()("RelayDPoPAuth", {
  error: HttpApiError.UnauthorizedNoContent,
  security: {
    dpop: HttpApiSecurity.dpop
  }
}) {}

const RelayDPoPAuthLive = Layer.succeed(RelayDPoPAuth)({
  dpop: Effect.fnUntraced(function*(handler, { credential }) {
    const request = yield* HttpServerRequest.HttpServerRequest
    const proofJwt = request.headers.dpop
    if (proofJwt.length === 0) {
      return yield* Effect.fail(new HttpApiError.Unauthorized({}))
    }

    const verifier = yield* VerifyDPoPProtectedResource
    const client = yield* verifier.verify({
      accessToken: credential,
      proofJwt: Redacted.make(proofJwt),
      request
    })

    return yield* Effect.provideService(handler, CurrentRelayClient, client)
  })
})

class RelayApi extends HttpApiGroup.make("relay")
  .add(HttpApiEndpoint.get("status", "/status", {
    success: Schema.Struct({ subject: Schema.String })
  }))
  .middleware(RelayDPoPAuth)
{}

class Api extends HttpApi.make("api").add(RelayApi) {}

const RelayApiLive = HttpApiBuilder.group(Api, "relay", (handlers) =>
  handlers.handle("status", Effect.fnUntraced(function*() {
    const client = yield* CurrentRelayClient
    return { subject: client.subject }
  }))
)

const ApiLive = HttpApiBuilder.layer(Api).pipe(
  Layer.provide(RelayApiLive),
  Layer.provide(RelayDPoPAuthLive)
)
```

`credential` is the access token extracted from `Authorization: DPoP <access-token>`. `ApiLive` requires an application layer for `VerifyDPoPProtectedResource`, which verifies the presented access token together with the proof JWT.

The proof header should not be declared as another security entry alongside `dpop`, because security entries are alternatives; checking it in the same middleware expresses that both token and proof are required.

## Validation

- `pnpm lint-fix`
- `pnpm test packages/platform-node/test/HttpApi.test.ts packages/platform-node/test/OpenApi.test.ts packages/tools/openapi-generator/test/OpenApiGenerator.test.ts`
- `pnpm test-types packages/effect/typetest/unstable/httpapi/HttpApiMiddleware.tst.ts`
- `pnpm docgen` from `packages/effect`
- `pnpm docgen` from `packages/tools/openapi-generator` (no docgen modules configured)
- `pnpm exec tsgo -p packages/effect/tsconfig.json --noEmit`
- `pnpm exec tsgo -p packages/tools/openapi-generator/tsconfig.json --noEmit`
- `pnpm check:tsgo` (exited successfully while reporting existing diagnostics in `packages/effect/typetest/Effect.tst.ts`)

### DIFF
--- a/packages/effect/src/unstable/httpapi/OpenApi.ts
+++ b/packages/effect/src/unstable/httpapi/OpenApi.ts
@@ -730,23 +730,18 @@ const makeSecurityScheme = (security: HttpApiSecurity): OpenAPISecurityScheme =>
         scheme: "basic"
       }
     }
-    case "Bearer": {
-      const format = Context.getOption(security.annotations, Format).pipe(
-        Option.map((format) => ({ bearerFormat: format })),
-        Option.getOrUndefined
-      )
+    case "Http": {
+      const format = security.scheme.toLowerCase() === "bearer"
+        ? Context.getOption(security.annotations, Format).pipe(
+          Option.map((format) => ({ bearerFormat: format })),
+          Option.getOrUndefined
+        )
+        : undefined
       return {
         ...meta,
         type: "http",
-        scheme: "bearer",
+        scheme: security.scheme,
         ...format
-      }
-    }
-    case "DPoP": {
-      return {
-        ...meta,
-        type: "http",
-        scheme: "DPoP"
       }
     }
     case "ApiKey": {
--- a/packages/effect/src/unstable/httpapi/OpenApi.ts
+++ b/packages/effect/src/unstable/httpapi/OpenApi.ts
@@ -742,6 +742,13 @@ const makeSecurityScheme = (security: HttpApiSecurity): OpenAPISecurityScheme =>
         ...format
       }
     }
+    case "DPoP": {
+      return {
+        ...meta,
+        type: "http",
+        scheme: "DPoP"
+      }
+    }
     case "ApiKey": {
       return {
         ...meta,
@@ -954,7 +961,7 @@ export interface OpenAPIComponents {
 }
 
 /**
- * Generated OpenAPI HTTP security scheme, such as bearer or basic authentication.
+ * Generated OpenAPI HTTP security scheme, such as bearer, DPoP, or basic authentication.
  *
  * @category models
  * @since 4.0.0
--- a/.changeset/dpop-security-scheme.md
+++ b/.changeset/dpop-security-scheme.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Add `HttpApiSecurity.dpop` for decoding DPoP-bound access tokens and generating the matching OpenAPI HTTP security scheme.
--- a/.changeset/dpop-security-scheme.md
+++ b/.changeset/dpop-security-scheme.md
@@ -3,4 +3,4 @@
 "@effect/openapi-generator": patch
 ---
 
-Add `HttpApiSecurity.dpop` for validating and decoding DPoP-bound access tokens, generating the matching OpenAPI HTTP security scheme, and preserving that scheme in generated `HttpApi` modules.
+Add `HttpApiSecurity.http` for arbitrary HTTP authorization schemes and `HttpApiSecurity.dpop` as a DPoP convenience declaration, with matching runtime decoding, OpenAPI generation, and generated `HttpApi` support.
--- a/.changeset/dpop-security-scheme.md
+++ b/.changeset/dpop-security-scheme.md
@@ -1,5 +1,6 @@
 ---
 "effect": patch
+"@effect/openapi-generator": patch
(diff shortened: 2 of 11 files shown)


packages/effect/src/unstable/httpapi/OpenApi.ts
```diff
@@ -742,6 +742,13 @@ const makeSecurityScheme = (security: HttpApiSecurity): OpenAPISecurityScheme =>
         ...format
       }
     }
+    case "DPoP": {
+      return {
+        ...meta,
+        type: "http",
+        scheme: "DPoP"
```
tim-smart: I don't think DPoP is a valid value here.

I think instead we can add a "prefix" option to the ApiKey security type.

juliusmarminge: it should be valid? 

From https://learn.openapis.org/specification/security.html#http-authentication

> The authorization scheme should be defined in the IANA registry for HTTP Authentication, but API providers can extend this to use custom values if required.

IANA Registry: https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

tim-smart: OK seems like we should have a parent `Http` type that allows any scheme type then.

tim-smart:
https://github.com/Effect-TS/effect-smol/pull/2291

tim-smart:
Superseded by a couple PRs

PR #2302, opened 2026-05-29
fix(vitest): preserve subproperties on chained access through it proxy

### LINKED ISSUE #6348: `@effect/vitest`: `it.describe.each` is not a function

4.0.0-beta.70

```typescript
import { assert, it } from "@effect/vitest"

it.describe("it.describe.each reproduction", () => {
  it.describe.each(["foo", "bar"] as const)("%s", (text) => {
    it.effect("runs an Effect test", () =>
      Effect.sync(() => {
        assert.include(["foo", "bar"], text)
      }))
  })
})
```

Code above should result in successful test run.

### What do you see instead?

`TypeError: it.describe.each is not a function`

### Additional information

_No response_

truffle-dev:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Closes Effect-TS/effect#6348.

`makeItProxy` in `packages/vitest/src/internal/internal.ts` returned `value.bind(target)` whenever a non-override property access produced a function. Binding strips the function's own properties, so chained access like `it.describe.each(...)`, `it.skip.each(...)`, or `it.only.each(...)` resolved the chained step to `undefined` and threw `TypeError: it.describe.each is not a function`.

`Reflect.get(target, property, receiver)` already returns the underlying vitest functions correctly; vitest's chainable APIs do not depend on `this === V.it`. Removing the bind restores chained access without changing direct-call behavior.

Adds a regression case in `packages/vitest/test/index.test.ts` covering `it.describe.each`, `it.skip.each`, and `it.only.each`. With the bind removed the suite reports `26 passed | 1 expected fail | 6 skipped`; reverting only the source change (test in place) reproduces the reporter's error against `it.skip.each`.

### DIFF
--- a/.changeset/fix-vitest-it-proxy-subproperties.md
+++ b/.changeset/fix-vitest-it-proxy-subproperties.md
@@ -1,0 +1,5 @@
+---
+"@effect/vitest": patch
+---
+
+Fix `it.describe.each`, `it.skip.each`, `it.only.each` (and similar chained access on the `it` proxy) by removing an unnecessary `value.bind(target)` from `makeItProxy`'s `get` trap. The bind stripped each function's own properties (`.each`, `.skip`, `.only`, ...), so `it.describe.each` resolved to `undefined` and threw `TypeError: it.describe.each is not a function`.
--- a/packages/vitest/src/internal/internal.ts
+++ b/packages/vitest/src/internal/internal.ts
@@ -64,8 +64,7 @@ const makeItProxy = <Methods extends object>(
       if (property in overrides) {
         return Reflect.get(overrides, property)
       }
-      const value = Reflect.get(target, property, receiver)
-      return typeof value === "function" ? value.bind(target) : value
+      return Reflect.get(target, property, receiver)
     }
   })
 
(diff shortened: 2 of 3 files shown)


truffle-dev:
Hi @gcanti, polite nudge in case this slipped past triage. The fix is one line in `packages/vitest/src/internal/internal.ts` — swap `value.bind(target)` for `Reflect.get(target, property, receiver)` so chained access like `it.describe.each` / `it.skip.each` / `it.only.each` survives the proxy. Happy to rebase if anything has moved.

thife455:
Bump, this is very useful when nesting layers

truffle-dev:
Second user hitting this now (thanks @thife455). Still a one-line change in `packages/vitest/src/internal/internal.ts`: `Reflect.get(target, property, receiver)` instead of `value.bind(target)`, so chained access through the `it` proxy (`it.describe.each`, `it.skip.each`, `it.only.each`) keeps its subproperties. Rebased and green. @tim-smart @gcanti happy to adjust if the proxy has moved since.

IMax153:
Hi @truffle-dev,

The GitHub issues linked to this PR have also been migrated to the `effect` repository:
- [Effect-TS/effect#6348](https://github.com/Effect-TS/effect/issues/6348)

Please use these migrated issue links in the replacement PR.

PR #2306, opened 2026-05-29
fix(HttpApi): strip space delimiter when decoding bearer/http credentials

### LINKED ISSUE #2305: Http Bearer token has extra space

beta.73 regressed Bearer by adding a prefix " " on the token

When users use >= beta.73, our Http API fails Bearer token authorization because either the client or server is injecting a `" "` in the parsed Bearer token

```ts
// Repro: HttpApiSecurity bearer credential gains a leading space.
//
//   bun add effect@4.0.0-beta.74   # bad  (also beta.73)
//   bun add effect@4.0.0-beta.72   # good
//   bun effect-bearer-repro.ts
//
// Regressed in Effect-TS/effect-smol@361ca30 (PR #2291), released in
// effect@4.0.0-beta.73: HttpApiSecurity.http sets schemeLength = scheme.length,
// and securityDecode does authorization.slice(schemeLength). The header is
// "<scheme> <credential>", so the slice must be scheme.length + 1 to skip the
// delimiter space. As-is, every bearer/http credential decodes with a leading
// space (basic is unaffected — its old `basicLen = "Basic ".length` survived).
import * as Effect from "effect/Effect";
import * as Redacted from "effect/Redacted";
import * as HttpClientRequest from "effect/unstable/http/HttpClientRequest";
import * as HttpServerRequest from "effect/unstable/http/HttpServerRequest";
import * as HttpApiBuilder from "effect/unstable/httpapi/HttpApiBuilder";
import * as HttpApiSecurity from "effect/unstable/httpapi/HttpApiSecurity";

const TOKEN = "abc123";

// --- CLIENT: build the request exactly as a real client does ---
const clientRequest = HttpClientRequest.get("http://localhost/").pipe(
  HttpClientRequest.bearerToken(TOKEN),
);
const wireHeader = clientRequest.headers["authorization"];
console.log("client sends Authorization:", JSON.stringify(wireHeader)); // "Bearer abc123"

// --- SERVER: decode it with the library's own securityDecode ---
const serverRequest = HttpServerRequest.fromWeb(
  new Request("http://localhost/", { headers: { authorization: wireHeader! } }),
);

const credential = Redacted.value(
  Effect.runSync(
    HttpApiBuilder.securityDecode(HttpApiSecurity.bearer).pipe(
      Effect.provideService(HttpServerRequest.HttpServerRequest, serverRequest),
      Effect.provideService(HttpServerRequest.ParsedSearchParams, {}),
    ),
  ),
);

console.log("server decoded credential:", JSON.stringify(credential));
console.log("schemeLength:", (HttpApiSecurity.bearer as any).schemeLength);

if (credential !== TOKEN) {
  throw new Error(
    `REGRESSION: decoded ${JSON.stringify(credential)} !== sent ${JSON.stringify(TOKEN)}. ` +
      `securityDecode sliced schemeLength=${(HttpApiSecurity.bearer as any).schemeLength} ` +
      `but "Bearer " is 7 chars — it drops one char short and keeps the delimiter space.`,
  );
}
console.log("OK: credential matches");

```

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

sam-goodwin:
Closes #2305

`securityDecode` slices `schemeLength` off the `Authorization` header, but the header is `"<scheme> <credential>"` — so the single space separating the scheme from the credential stays attached to the decoded value. Every `bearer`/`http` credential decodes with a leading space (`basic` is unaffected — it still uses the `basicLen = "Basic ".length` constant).

```diff
- (request) => Redacted.make((request.headers.authorization ?? "").slice(self.schemeLength)) as any
+ (request) => Redacted.make((request.headers.authorization ?? "").slice(self.schemeLength + 1)) as any
```

```ts
// before:  "Bearer abc123".slice("Bearer".length /* 6 */) === " abc123"
// after:   "Bearer abc123".slice("Bearer".length + 1 /* 7 */) === "abc123"
```

### Regression

Introduced in 361ca30eb6e134feece547d6e00f82be4cb23f75 (#2291, "Add HttpApiSecurity.http for passing custom schemes"), first released in `effect@4.0.0-beta.73`. That change replaced the hard-coded ``const bearerLen = `Bearer `.length`` (7, which included the delimiter space) with `schemeLength: options.scheme.length` (6) and `slice(self.schemeLength)`, dropping the `+ 1`. Bisection: `beta.72` good, `beta.73` bad.

### Impact

Any HTTP API guarded by `HttpApiSecurity.bearer` (or a custom `HttpApiSecurity.http({ scheme })`) rejects valid credentials, because the server-side decoded token is one byte longer than what the client sent.

### Test

Added `HttpApiSecurity.test.ts` exercising `securityDecode` for both `bearer` and a custom `http` scheme. Verified it fails on `main` (`Received: " abc123"`) and passes with the fix.

### DIFF
--- a/packages/effect/test/unstable/httpapi/HttpApiSecurity.test.ts
+++ b/packages/effect/test/unstable/httpapi/HttpApiSecurity.test.ts
@@ -1,0 +1,35 @@
+import { describe, it } from "@effect/vitest"
+import { strictEqual } from "@effect/vitest/utils"
+import { Effect, Redacted } from "effect"
+import { HttpClientRequest, HttpServerRequest } from "effect/unstable/http"
+import { HttpApiBuilder, HttpApiSecurity } from "effect/unstable/httpapi"
+
+const decode = (security: HttpApiSecurity.HttpApiSecurity, authorization: string) =>
+  HttpApiBuilder.securityDecode(security).pipe(
+    Effect.provideService(
+      HttpServerRequest.HttpServerRequest,
+      HttpServerRequest.fromWeb(new Request("http://localhost/", { headers: { authorization } }))
+    ),
+    Effect.provideService(HttpServerRequest.ParsedSearchParams, {})
+  )
+
+describe("HttpApiSecurity", () => {
+  describe("securityDecode", () => {
+    it.effect("decodes a bearer token without a leading space", () =>
+      Effect.gen(function*() {
+        const token = "abc123"
+        // build the header exactly as a client does
+        const { headers } = HttpClientRequest.get("http://localhost/").pipe(
+          HttpClientRequest.bearerToken(token)
+        )
+        const credential = yield* decode(HttpApiSecurity.bearer, headers.authorization!)
+        strictEqual(Redacted.value(credential), token)
+      }))
+
+    it.effect("decodes a custom http scheme without a leading space", () =>
+      Effect.gen(function*() {
+        const credential = yield* decode(HttpApiSecurity.http({ scheme: "Token" }), "Token abc123")
+        strictEqual(Redacted.value(credential), "abc123")
+      }))
+  })
+})
--- a/packages/effect/test/unstable/httpapi/HttpApiSecurity.test.ts
+++ b/packages/effect/test/unstable/httpapi/HttpApiSecurity.test.ts
@@ -4,31 +4,34 @@ import { Effect, Redacted } from "effect"
 import { HttpClientRequest, HttpServerRequest } from "effect/unstable/http"
 import { HttpApiBuilder, HttpApiSecurity } from "effect/unstable/httpapi"
 
-const decode = (security: HttpApiSecurity.HttpApiSecurity, authorization: string) =>
-  HttpApiBuilder.securityDecode(security).pipe(
-    Effect.provideService(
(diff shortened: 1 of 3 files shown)


IMax153 (dismissed):

IMax153:
@sam-goodwin - looks like there's a type error in CI.

packages/effect/test/unstable/httpapi/HttpApiSecurity.test.ts
```diff
@@ -0,0 +1,35 @@
...
+const decode = (security: HttpApiSecurity.HttpApiSecurity, authorization: string) =>
+  HttpApiBuilder.securityDecode(security).pipe(
+    Effect.provideService(
+      HttpServerRequest.HttpServerRequest,
+      HttpServerRequest.fromWeb(new Request("http://localhost/", { headers: { authorization } }))
+    ),
+    Effect.provideService(HttpServerRequest.ParsedSearchParams, {})
+  )
+
+describe("HttpApiSecurity", () => {
```
IMax153, line 7: We should also have a test that validates that credentials from a different scheme than the one specified are rejected.

See the tests in Julius' PR #2315

sam-goodwin:
Thanks @tim-smart for driving it home.

PR #2309, opened 2026-05-31
[Draft/RFC] Tracer.instrumenting — zero-ceremony service span tracing

damienmeur:
# [DRAFT / RFC] `Tracer.instrumenting` — zero-ceremony service span tracing

> **This is a draft to start a discussion, not a merge request.** I'd love the
> team's eyes on the approach (especially concurrency and the composition
> question below) — happy to reshape the API, location, and naming entirely.

## Problem

When a failure happens deep in a service stack, the rendered cause shows runtime
internals but not _which_ business operation failed. The fix today is to
hand-wrap every method in `Effect.fn("Service.method")` — per-method ceremony on
every method of every service you want to see in a trace.

## What this adds

A single constructor, `Tracer.instrumenting`, that returns a `Tracer` which
wraps the methods of _matched_ services in spans automatically:

```ts
import { Effect, Tracer } from "effect"

program.pipe(
  Effect.provide(appLayer),
  Effect.withTracer(Tracer.instrumenting({ match: (key) => key.startsWith("@app/") }))
)
```

On failure the business chain is now present (innermost first) via the
`Cause.StackTrace` annotation — e.g. `@app/Db.exec` → `@app/Repo.persistMany` →
`@app/Importer.persist` — with **zero per-method ceremony**.

## How it works

It hooks the **public `Tracer` `context` seam** — the same one `OtlpTracer` uses
— and nothing else:

- When a service whose `Context` tag key satisfies `match` is resolved, the
  tracer returns a span-wrapped copy of its implementation (cached per instance
  in a `WeakMap`). Each `Effect`-returning method runs inside
  `withSpan(` ${key}.${method}`)`.
- Wrapping at **resolution** captures dependencies grabbed at make-time
  (`const repo = yield* Repo`): install the tracer _outside_ `Effect.provide`, the
  layer build runs under it, and the make-time grab resolves the _wrapped_ repo —
  so the whole nested chain is traced, not just the entry point.
- Span creation is delegated to a base tracer (native by default; pass `tracer`
  to compose with another backend's `span` and `context`).

No monkey-patching and no `node:inspector` — only public API — so it behaves the
same on Node, Deno, Bun, and the browser.

### Note on the implementation (cycle avoidance)

`Tracer.ts` sits early in the module-load graph (`internal/effect.ts` imports it),
so it cannot import the `Effect` barrel — doing so reorders init and breaks the
build. `instrumenting` instead imports `withSpan` directly from
`internal/effect.ts` and the exit/effect predicates from `internal/core.ts`, and
references `withSpan` only inside the returned closure (never at module-init), so
no eager cycle is introduced. (Flagging this in case you'd prefer a different
home for the implementation.)

## What works (tested)

`packages/effect/test/TracerInstrumenting.test.ts` — all green:

- a matched service method runs inside a span named `key.method`
- **make-time deps**: `Svc` grabs `Repo` at make-time → the cause's
  `Cause.StackTrace` chain is `@svc/Repo.find` → `@svc/Svc.load`
- unmatched services are passed through untouched
- **parallel fibers**: 3 concurrent
  `Effect.forEach(..., { concurrency: "unbounded" })` branches each carry their
  _own_ `@svc/Repo.find` → `@svc/Svc.load` chain — no cross-fiber span bleeding

## Performance

`packages/effect/benchmark/tracer/instrumenting.ts`. The honest comparison is vs
hand-written `Effect.fn`, not vs an untraced no-op (a span costs what a span
costs). Measured on Node 26 / Bun 1.3:

- **Bare spans, no I/O**: `instrumenting` is ~1.2× the per-call cost of `Effect.fn`
  on Node, ~1.0× on Bun. (Versus a do-nothing `Effect.succeed` baseline it's
  ~100×, but that's ~100× of ~0.06µs — a span is ~3.6µs absolute.)
- **Realistic nested request** (nested services + ~1ms DB round-trips): **~1.03×
  vs untraced** on both runtimes — lost in I/O latency.
- Installing any `context`-hook tracer adds a small per-primitive tax (~1.05–1.12×)
  on all fiber steps — which is why this is opt-in, never on by default.

## Recovering file:line (node-only — food for thought)

This PR's core delivers **span names + correct nesting**, which is what drives
the chain. It does **not** put source `file:line` in core: auto-wrapping happens
away from the method's definition site, so the span's own stack points at the
wrapper, not your code.

The location can still be recovered — node-only — by reading the **original
function's** `[[FunctionLocation]]` via the V8 inspector (it points at the real
method def, not the wrapper). The tracer keeps a `Map<spanName, originalFn>` at
wrap time (refs only, ~free); then for the methods in a failing chain:

```ts
const session = new Session()
session.connect()
const urls = new Map<string, string>()
session.on("Debugger.scriptParsed", (m) => urls.set(m.params.scriptId, m.params.url))
await session.post("Debugger.enable") // replays scriptParsed for loaded scripts

;(globalThis as any).__probe = originalFn
const { result } = await session.post("Runtime.evaluate", { expression: "globalThis.__probe" })
const { internalProperties } = await session.post("Runtime.getProperties", { objectId: result.objectId })
const loc = internalProperties?.find((p) => p.name === "[[FunctionLocation]]")?.value?.value
// loc = { scriptId, lineNumber, columnNumber } -> urls.get(loc.scriptId) + line/col
```

Two ways to time it (per @mikearnaldi's suggestion) — measured in
`benchmark/tracer/locations.ts`, Node 26; both yield identical locations:

| strategy | cost | when |
| --- | --- | --- |
| **lazy** — transient session, resolve the failing chain | **~2 ms** | per failure |
| **pre-warm** — resolve all registered methods up front | **~10 ms** | once at boot |
| **pre-warmed** — `Map.get` on the error path | **~0 ms** (sub-µs) | per failure |

So pre-warming (e.g. from a `Context.Service` at startup, as suggested) amortizes
a ~10 ms one-time cost and makes every failure's lookup free; lazy avoids the
fixed cost but pays ~2 ms per failure. Either way it's `node:inspector`, so it
stays out of the runtime-agnostic core — it'd belong in `platform-node`/userland.

## Known limitations

There is exactly **one interception seam**: the fiber-loop `context` hook
(`internal/effect.ts`), which fires only when a service `Key` is *resolved*. It
wraps the resolved value and hands it back as the result — it never writes the
wrapped copy back into the context. So the boundaries are precisely the call
paths that don't pass through that seam (thanks @mikearnaldi for surfacing
these). All four cases below are pinned by tests.

- **internal self-calls via `this`** — *now wrapped.* Each method is bound to the
  wrapped copy, so `this.other()` resolves the sibling off the wrapped receiver
  and that inner span is captured.
- **internal self-calls via a closure** — *not wrapped, and unfixable at
  runtime.* When a method calls a sibling through a closure-captured local
  (`make = () => { const m2 = ...; const m1 = () => m2(); return { m1, m2 } }`),
  the call is a lexical reference — not a property access and not `this` — so
  there is no interception point at all. Recompiling the method via
  `Function.prototype.toString` can't help either: `new Function` loses the
  captured scope. Closeable only at build time or by opting into `Effect.fn`.
- **`Context.get(ctx, Tag)`** — *not wrapped.* `Context.get` is a synchronous
  `mapUnsafe.get(key)` read that never touches the fiber, so the resolve-time
  hook never runs and the raw impl comes back. Closing this would mean wrapping
  at *store* time (a `Layer`-level transform that re-provides wrapped services)
  rather than at resolve time — a different, eager, placement-sensitive surface,
  not this `Tracer` seam. (Idiomatic `yield* Tag` is wrapped; `Context.get` is
  the manual escape hatch.)
- **service shape** — it assumes services are objects of `Effect`-returning
  methods accessed via `yield*`. Common in practice but not guaranteed; other
  shapes (bare functions, getters, prototype methods) won't be wrapped.

## Open questions for the team

1. **Composition with `OtlpTracer`.** `Tracer.Tracer` is a single reference, and
   both this and OTLP want the one `context` hook. `instrumenting` accepts a base
   `tracer` and chains to its `context` for non-matched primitives — is that the
   right shape, or should `context` hooks compose more formally so auto-instrument
   and export can coexist cleanly?
2. **Where should this live, and is the API right?** `Tracer.instrumenting`? a
   `Layer` helper? Naming, the `match` predicate, the `spanName` override — all
   open.

## Files

- `packages/effect/src/Tracer.ts` — `instrumenting` export (+ JSDoc)
- `packages/effect/test/TracerInstrumenting.test.ts` — tests incl. parallel fibers
- `packages/effect/benchmark/tracer/instrumenting.ts` — tinybench benchmark
- `packages/effect/benchmark/tracer/locations.ts` — node:inspector file:line bench (lazy vs pre-warm)
- `.changeset/add-tracer-instrumenting.md`

### DIFF
--- a/.changeset/add-tracer-instrumenting.md
+++ b/.changeset/add-tracer-instrumenting.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+Add `Tracer.instrumenting` to auto-instrument matched services with one span per method, surfacing the business call chain in failures with zero per-method ceremony. It hooks the public `Tracer` `context` seam only — no monkey-patching, no `node:inspector` — so it works on every runtime.
--- a/packages/effect/benchmark/tracer/instrumenting.ts
+++ b/packages/effect/benchmark/tracer/instrumenting.ts
@@ -1,0 +1,123 @@
+import { Context, Effect, Layer, Tracer } from "effect"
+import { Bench } from "tinybench"
+
+// Measures `Tracer.instrumenting` against the idiomatic manual way to span a
+// method (`Effect.fn`) and against no tracing — both with bare, do-nothing
+// method bodies (where the per-span cost dominates) and with a realistic nested
+// request that ends in I/O (where it is lost against latency).
+//
+// Run: tsx packages/effect/benchmark/tracer/instrumenting.ts
+
+interface RepoShape {
+  readonly find: (id: number) => Effect.Effect<number>
+}
+class Repo extends Context.Service<Repo, RepoShape>()("@svc/Repo") {}
+
+interface SvcShape {
+  readonly load: (id: number) => Effect.Effect<number>
+}
+class Svc extends Context.Service<Svc, SvcShape>()("@svc/Svc") {}
+
+const matchTracer = Tracer.instrumenting({ match: (key) => key.startsWith("@svc/") })
+
+// --- bare spans, no I/O: 200 calls/run ---
+
+const RepoBare = Layer.succeed(Repo, { find: (id) => Effect.succeed(id * 2) })
+const SvcBare = Layer.effect(
+  Svc,
+  Effect.gen(function*() {
+    const repo = yield* Repo
+    return { load: (id: number) => repo.find(id) }
+  })
+).pipe(Layer.provide(RepoBare))
+
+const RepoBareFn = Layer.succeed(Repo, {
+  find: Effect.fn("Repo.find")(function*(id: number) {
+    return id * 2
+  })
+})
+const SvcBareFn = Layer.effect(
+  Svc,
+  Effect.gen(function*() {
+    const repo = yield* Repo
+    return {
+      load: Effect.fn("Svc.load")(function*(id: number) {
+        return yield* repo.find(id)
+      })
+    }
+  })
+).pipe(Layer.provide(RepoBareFn))
+
+const bareLoop = Effect.gen(function*() {
+  const svc = yield* Svc
+  let total = 0
+  for (let i = 0; i < 200; i++) {
+    total += yield* svc.load(i)
+  }
+  return total
+})
+
+const barePlain = bareLoop.pipe(Effect.provide(SvcBare))
+const bareManual = bareLoop.pipe(Effect.provide(SvcBareFn))
+const bareAuto = barePlain.pipe(Effect.withTracer(matchTracer))
+
+// --- realistic nested request ending in I/O: 20 requests/run ---
+
+interface DbShape {
+  readonly query: (id: number) => Effect.Effect<number>
+  readonly exec: (id: number) => Effect.Effect<number>
+}
+class Db extends Context.Service<Db, DbShape>()("@svc/Db") {}
+
+const ioWork = (id: number) =>
+  Effect.gen(function*() {
+    yield* Effect.sleep("1 millis")
+    return id
+  })
+
+const DbLayer = Layer.succeed(Db, { query: ioWork, exec: ioWork })
+const ReqRepo = Layer.effect(
+  Repo,
+  Effect.gen(function*() {
+    const db = yield* Db
+    return { find: (id) => Effect.flatMap(db.query(id), (n) => db.exec(n)) }
+  })
+).pipe(Layer.provide(DbLayer))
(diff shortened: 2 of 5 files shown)


IMax153:
Hi @damienmeur,

PR #2270, opened 2026-05-25
EFF-845: Add HttpApi streaming responses

IMax153:
## Summary

- Adds HttpApi streaming success response support for SSE and Uint8Array streams.
- Wires streaming response encoding/decoding through HttpApiBuilder, HttpApiClient, OpenAPI generation, and the OpenAPI generator tooling.
- Adds runtime tests, type tests, and changesets for the streaming response APIs.
- Addresses review feedback by removing the SuccessSchema alias, making getSuccessSchemas include stream schemas, using getSuccessSchemas in OpenAPI success extraction, and filtering streams from reflected buffered success maps.

## Validation

- pnpm lint-fix
- pnpm test packages/effect/test/unstable/httpapi/HttpApiEndpoint.test.ts packages/effect/test/unstable/httpapi/HttpApiBuilder.test.ts packages/effect/test/unstable/httpapi/HttpApiClient.test.ts packages/effect/test/unstable/httpapi/OpenApi.test.ts
- pnpm test-types packages/effect/typetest/unstable/httpapi/HttpApiEndpoint.tst.ts packages/effect/typetest/unstable/httpapi/HttpApiClient.tst.ts packages/effect/typetest/unstable/httpapi/HttpApiBuilder.tst.ts packages/effect/typetest/unstable/httpapi/HttpApiSchema.tst.ts packages/effect/typetest/unstable/httpapi/HttpApiMiddleware.tst.ts
- pnpm check:tsgo

PR #2315, opened 2026-06-01
Fix HTTP API security credential decoding

juliusmarminge:
## Summary

- restore validated HTTP authorization token decoding for generalized `HttpApiSecurity.http` schemes
- strip the scheme separator before passing credentials to security middleware
- preserve scheme validation for `Basic` decoding and add direct regression tests

## Root Cause

When `HttpApiSecurity.http` generalized bearer-style authorization schemes, `securityDecode` changed from parsing the authorization token to slicing at `scheme.length`. For a standard header such as `Authorization: Bearer token`, the middleware therefore received `" token"` rather than `"token"`. It also no longer rejected a token supplied with a different authorization scheme.

## Validation

- `DPRINT_MAX_THREADS=1 pnpm lint-fix`
- `pnpm test packages/effect/test/unstable/httpapi/HttpApiBuilder.test.ts`
- `pnpm check:tsgo`

`DPRINT_MAX_THREADS=1` was used because the unrestricted full-tree `dprint fmt` subprocess was killed by the local OS while formatting this checkout.

### DIFF
--- a/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
@@ -446,7 +446,7 @@ export const securityDecode = <Security extends HttpApiSecurity.HttpApiSecurity>
     case "Http": {
       return Effect.map(
         HttpServerRequest,
-        (request) => Redacted.make((request.headers.authorization ?? "").slice(self.schemeLength)) as any
+        (request) => Redacted.make(decodeAuthorizationToken(request.headers.authorization, self.scheme)) as any
       )
     }
     case "ApiKey": {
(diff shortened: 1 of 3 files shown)


juliusmarminge:
@tim-smart

IMax153:
@juliusmarminge - see #2270 - not sure the regex parsing is necessary if we use the simpler implementation from Sam's PR

packages/effect/src/unstable/httpapi/HttpApiBuilder.ts
```diff
@@ -521,7 +523,12 @@ export const securitySetCookie = (
 // Internal
 // -----------------------------------------------------------------------------
 
-const basicLen = `Basic `.length
+const authorizationToken = /^([^ ]+) +([-A-Za-z0-9._~+/]+=*)$/
```
IMax153, line 526: If we do elect to go with a regex-based decoder, doesn't the Basic Authorization schema _require_ a single space? This regex would allow multiple if I'm not mistaken.

tim-smart:
Duplicate of #2306

PR #2325, opened 2026-06-03
Add layer to configure otlp with standard environment variables

fubhy:


packages/effect/src/unstable/observability/internal/otlpEnv.ts
```diff
@@ -0,0 +1,58 @@
...
+  Schema.decode({
+    decode: SchemaGetter.transform((_: ReadonlyArray<string>) => _.filter((_) => _ !== "")),
+    encode: SchemaGetter.passthrough()
+  })
+)
+
+const HeadersRecord = Config.Record(Schema.String, Schema.String)
+
+export const headers = (signal: Signal) =>
+  Config.make((provider) =>
```
tim-smart, line 26: This is sloppy

PR #2338, opened 2026-06-04
fix(ai-openai-compat): handle null function.name on streamed tool-call fragments

boozedog:
## What

Two coupled fixes so streamed tool-call arguments aren't lost when an OpenAI-compatible provider sends `function.name: null` on continuation fragments.

## Root cause

Some providers (e.g. Fireworks/Kimi) send the tool name only on the **first** streamed `tool_calls` fragment and `function.name: null` on the continuation fragments that carry the argument deltas. Two things broke:

1. `ChatCompletionToolFunctionDelta.name` was `Schema.optionalKey(Schema.String)` (non-nullable). Chunk validation (`Schema.is`) rejected every `name: null` continuation, and `decodeChatCompletionSseData` silently dropped those chunks — discarding their argument deltas. Result: one empty `tool-params-start`, zero `tool-params-delta`, an empty/partial tool call.
2. With the schema fixed (chunks no longer dropped), the streaming decoder's tool-name accumulator used `providerToolName !== undefined`, so a `null` continuation overwrote the accumulated name with a placeholder → the assembled `tool-call` failed the toolkit's tool-name schema (`Expected "X", got null`).

## Fix

- `OpenAiClient.ts`: `name: Schema.optionalKey(Schema.NullOr(Schema.String))`.
- `OpenAiLanguageModel.ts`: adopt the provider tool name only when it's a string (`typeof providerToolName === "string"`); otherwise keep the name accumulated from the first fragment.

## Test

New `streamText` test that streams arguments across `function.name: null` continuation fragments and asserts the assembled `tool-call.params` is complete. Fails on `main` (`Invalid output: Missing key`), passes with the fix.

## Notes

- The underlying fragility — `decodeChatCompletionSseData` silently dropping any chunk that fails validation — is raised separately in Effect-TS/effect#6346. This PR is the narrow, low-risk fix.
- Changeset included (`@effect/ai-openai-compat` patch).

## Validation

`pnpm check` clean · full `ai-openai-compat` suite (30) passing · `oxlint` + `dprint` clean.


packages/ai/openai-compat/src/OpenAiLanguageModel.ts
```diff
@@ -1234,8 +1234,11 @@ const makeStreamResponse = Effect.fnUntraced(
             const toolIndex = deltaTool.index ?? indexInChunk
             const activeToolCall = activeToolCalls[toolIndex]
             const toolId = activeToolCall?.id ?? deltaTool.id ?? `${event.id}_tool_${toolIndex}`
+            // Providers may send `function.name: null` on continuation
+            // fragments; keep the name already accumulated for this tool call
+            // rather than overwriting it with a placeholder.
             const providerToolName = deltaTool.function?.name
-            const toolName = providerToolName !== undefined
+            const toolName = typeof providerToolName === "string"
```
IMax153: I think I would rather use the `Predicate` module here.


```suggestion
            const toolName = Predicate.isNotNullish(providerToolName)
```

packages/ai/openai-compat/src/OpenAiLanguageModel.ts
```diff
@@ -1234,8 +1234,11 @@ const makeStreamResponse = Effect.fnUntraced(
             const toolIndex = deltaTool.index ?? indexInChunk
             const activeToolCall = activeToolCalls[toolIndex]
             const toolId = activeToolCall?.id ?? deltaTool.id ?? `${event.id}_tool_${toolIndex}`
+            // Providers may send `function.name: null` on continuation
+            // fragments; keep the name already accumulated for this tool call
+            // rather than overwriting it with a placeholder.
```
IMax153:
```suggestion
```

IMax153 (changes requested):

boozedog:
looking at the broken node test now

IMax153:
> looking at the broken node test now

It's just a flaky test - you can ignore it for now.

IMax153 (approved):

IMax153 (approved):

PR #2346, opened 2026-06-07
feat(effect): add Atom.dedupe and Atom.dedupeWith for value-equality emission dedup

front-depiction:
## What

Adds the standard `dedupe` / `dedupeWith` pair (mirroring `Array.dedupe` / `Array.dedupeWith`) as opt-in combinators that switch an atom's emission guard from reference identity (`Object.is`) to value equality. When a written or recomputed value compares equal to the current one, the atom suppresses the notification **and** skips invalidating downstream atoms.

```ts
import { Atom } from "effect/unstable/reactivity"

// Equal.equals (default value equality)
const a = Atom.make({ count: 0 }).pipe(Atom.dedupe)

// a custom Equivalence
const b = Atom.make({ count: 0, meta: "x" }).pipe(
  Atom.dedupeWith((x, y) => x.count === y.count)
)
```

`dedupe` is a unary combinator defaulting to `Equal.equals`; `dedupeWith` is a `dual` taking a required `Equivalence` — exactly the `dedupe` / `dedupeWith` convention used across `Array`/`Chunk`/`Iterable`. The default behavior of every atom is unchanged: the `setValue` guard only consults the equivalence when one of these combinators set it, otherwise it stays `Object.is`.

This PR is the **conservative, opt-in route** — backward-compatible, default unchanged.

## Behavior (covered by tests)

- **Source set dedup:** a value-equal `set` on a deduped atom suppresses the notification (default `Object.is` still emits).
- **Map dedup is free from source dedup:** a suppressed source never invalidates its children.
- **Derived-output dedup is per-node:** because every atom emits through the same guarded `setValue`, applying `dedupe`/`dedupeWith` to a derived/combining atom dedupes *its own* output even when inputs changed — a value-stable hop terminates the chain tail.
- **Custom `Equivalence`:** `dedupeWith` with a partial equivalence suppresses writes that differ only in ignored fields, while a change to a compared field still notifies.

## Implementation

- `AtomRegistry` `NodeImpl.setValue`: guard becomes `atom.equivalence ? atom.equivalence(prev, next) : Object.is(prev, next)`.
- `Atom`: an optional `equivalence?(self, that): boolean` interface member (a method signature, to keep `Atom<A>` covariant in `A`), and the `dedupe` (unary, `Equal.equals`) + `dedupeWith` (`dual`, custom `Equivalence`) combinators sharing a private `setEquivalence` helper. No casts.

## Relationship to #2347

This is the **opt-in** option (default stays `Object.is`). Its sibling **#2347** is the **surgical alternative** that simply restores `Equal.equals` as the *default* in one line (on the view that the `Object.is` switch in #853 was a slip). Pick whichever fits: opt-in combinators (this) or default restore (#2347). See also the v3-side proposal `tim-smart/effect-atom#415`.

Draft (the conservative fallback); #2347 is the primary proposal.


IMax153:
Hi @front-depiction,

PR #2347, opened 2026-06-07
fix(effect): restore value-equality emission dedup as the Atom default (regression from v3)

front-depiction:
## What

A surgical one-line change: in `AtomRegistry` `NodeImpl.setValue`, the emission guard goes from reference identity to value equality —

```diff
-    if (Object.is(this._value, value)) {
+    if (Equal.equals(this._value, value)) {
```

(plus the `Equal` import). No new API. With this, writing or recomputing a value that is `Equal.equals` to the current one no longer notifies subscribers or invalidates downstream atoms — restoring the v3 default.

## Why — this is a silent regression from v3 `@effect-atom`

The v3 standalone `@effect-atom/atom` deduped atom emissions on **value equality** (`registry.ts` `setValue`: `if (Equal.equals(this._value, value))`). In the v4-core rewrite the guard became `Object.is` — **reference identity** — so a fresh-reference-but-value-equal write (which derived atoms produce on every recompute) now re-emits and cascades a re-render through the whole derived chain.

The change was unintentional, not a reasoned design decision:

- **Origin:** the `Object.is` guard was born in commit `00a9ef20` / **#853 "add Atom modules"** — the bulk port (43 files, ~8k insertions) that *created* `AtomRegistry.ts`. It rode in mechanically.
- **No changeset, no test** in that commit pins reference-equality dedup as intended behavior.
- **intra-commit inconsistency:** the *same* commit's `AtomRef.ts` kept `Equal.equals` (`if (Equal.equals(value, this.value))`, and `AtomRef` implements `Equal.Equal`). Within one commit the sibling `Ref` preserved value-equality while only `NodeImpl.setValue` regressed — characteristic of an accidental narrowing during the port.
- **Maintainer intent:** in `tim-smart/effect-atom#218` the maintainer treats value-equality dedup as the correct/expected behavior.

## Safety

The full `reactivity/` test suite passes under the flipped default (104/104) — no existing test relied on reference-identity emission. New tests assert the default now suppresses a fresh-reference-but-`Equal.equals` set (and skips downstream invalidation) while a genuinely different value still emits.

## Relationship to #2346

This is the **surgical "restore the default"** option. Its sibling **#2346** instead adds opt-in `Atom.dedupe` / `Atom.dedupeWith` combinators and leaves the default as `Object.is`. The two are independent:

- If you agree this was a slip → **this PR** restores the v3 default in one line.
- If you'd rather keep `Object.is` as the default and make value-equality opt-in → **#2346** is the conservative alternative.

(See also the v3-side proposal `tim-smart/effect-atom#415`.)

### DIFF
--- a/.changeset/atom-value-equality-default.md
+++ b/.changeset/atom-value-equality-default.md
@@ -1,0 +1,9 @@
+---
+"effect": minor
+---
+
+AtomRegistry: dedupe atom writes by value equality (`Equal.equals`) instead of reference identity (`Object.is`).
+
+When an atom is set to a fresh-reference value that is `Equal.equals` to its current value, the registry no longer treats it as a change: no subscribers are notified and no downstream atoms are invalidated. A genuinely different value still notifies and invalidates as before.
+
+This restores the value-equality dedup behavior of the standalone v3 `@effect-atom/atom` package, which used `Equal.equals` in its `setValue` guard. The guard narrowed to `Object.is` during the bulk port of the Atom subsystem into core, while the sibling `AtomRef` kept `Equal.equals` — an inconsistency this change resolves. Atoms holding structurally-equal values (e.g. `Data` instances, `Option`, `Result`) now behave as expected without an opt-in combinator.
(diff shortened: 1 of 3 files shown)


tim-smart:
This was intentional, as Equal.equals now use deep equality checks for everything which would be a perf footgun here.

front-depiction:
> This was intentional, as Equal.equals now use deep equality checks for everything which would be a perf footgun here.

Perhaps something like #2346 makes sense then as an opt in?

PR #2351, opened 2026-06-08
Added module for Machine

SandroMaglione:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added `Machine` module with Atom adapter:
- `Machine` ("state chart")
- `AtomMachine` integration for effect-atom

The feature is exposed under `effect/unstable/machine`.

```ts
import { Machine } from "effect/unstable/machine"

class Cart extends Schema.TaggedClass<Cart>("Cart")("Cart", {
  orderId: Schema.String
}) {}

class Editing extends Schema.TaggedClass<Editing>("Editing")("Editing", {}) {}
class Reviewing extends Schema.TaggedClass<Reviewing>("Reviewing")("Reviewing", {}) {}

class Checkout extends Schema.TaggedClass<Checkout>("Checkout")("Checkout", {
  orderId: Schema.String
}) {}

class Payment extends Schema.TaggedClass<Payment>("Payment")("Payment", {
  status: Schema.Literals(["pending", "paid"])
}) {}

class Delivery extends Schema.TaggedClass<Delivery>("Delivery")("Delivery", {
  status: Schema.Literals(["label", "shipped"])
}) {}

class Review extends Schema.TaggedClass<Review>("Review")("Review", {}) {}

class Submit extends Schema.TaggedClass<Submit>("Submit")("Submit", {
  orderId: Schema.String
}) {}

const OrderStates = Machine.defineStates({
  cart: {
    schema: Cart,
    initial: "editing",
    states: {
      editing: Editing,
      reviewing: Reviewing
    }
  },
  checkout: {
    schema: Checkout,
    type: "parallel",
    states: {
      payment: Payment,
      delivery: Delivery
    }
  }
})

const orderMachine = Machine.make({
  id: "Order",
  states: OrderStates.states,
  events: [Review, Submit],
  initial: () =>
    OrderStates.initial.cart(
      new Cart({ orderId: "order-1" }),
      (cart) => cart.editing(new Editing({}))
    )
}).handle({
  cart: {
    states: {
      editing: {
        on: {
          Review: ({ target }) => target.local.reviewing(new Reviewing({}))
        }
      },
      reviewing: {
        on: {
          Submit: ({ event, target }) =>
            target.full.checkout(
              new Checkout({ orderId: event.orderId }),
              (checkout) =>
                checkout
                  .payment(new Payment({ status: "pending" }))
                  .delivery(new Delivery({ status: "label" }))
            )
        }
      }
    }
  }
})
```

Feel free to push back on any of the API decisions or to propose or ask for any change.

## Related

This work is inspired by [`xstate`](https://github.com/statelyai/xstate), and it builds on top of #1945.


mikearnaldi:
quick feedback on naming, I wouldn't use `Actor / ActorSystem` due to potential conflicts in cluster I'd try to scope it more to semantically be tied to `Machine`

SandroMaglione:
> quick feedback on naming, I wouldn't use `Actor / ActorSystem` due to potential conflicts in cluster I'd try to scope it more to semantically be tied to `Machine`

I would rename `Actor` to `Process`, which is how [Erlang](https://www.erlang.org/doc/system/ref_man_processes.html) calls a similar abstraction. And also possibly rename `StateMachine` to `Machine`.

Since `Actor` is not only for state machines, I would avoid naming to be too tight between the two.

The problem is that `effect/unstable/process` is already taken by `ChildProcess`. Would it be okay to move `ChildProcess` in a separate path and use `/process` for this new module?

Another options would be `Interpreter`, but it sounds less good.

thife455:
One thing I ran into while experimenting with a similar API is that, for model-based testing, it is very useful to have the possible transitions available as data, not only as handlers.

Without some declaration/introspection of the valid { state, event } -> next state transitions, it becomes hard to derive a model that can be used by something like fast-check’s fc.commands API. In practice, the commands need to know which events are currently valid for a given model state so they can generate meaningful event sequences instead of mostly generating invalid/no-op cases.

I’m not sure what the right API shape would be here, but it would be very nice if the state machine definition could expose enough transition metadata to support model-based testing. For example, something along the lines of deriving/generating fc.Commands from the machine, or at least making the valid transitions inspectable.

SandroMaglione:
@thife455 the current API has a few method to inspect the machine without running the transition:
- `enabled`: gives you the list of enabled events for a given state
- `plan`: given a state+event returns the next expected state and the list of actions

Is this what you were looking for? Is this enough to implement your use case?

thife455:
Not quite, but enabled and plan do sound like useful building blocks.

What I was trying to do was closer to model-based testing, similar to Hypothesis state machines or fast-check’s fc.commands API:

https://fast-check.dev/docs/advanced/model-based-testing/

The idea is to generate arbitrary sequences of valid events, run them against the machine, and assert invariants over the transitions.

For example, I was experimenting with an API like this:
```ts
it.effect.prop(
  "Description",
  [Machine.commands(WalkmanMachine, { minLength: 10, maxLength: 30 })],
  ([cmds]) =>
    Machine.check(WalkmanMachine, cmds, {
      state: {
        Playing: ({ previous, event, next }) => {
          assert(WalkmanMachine.is.events.Play(event))
        }
      }
    })
)
```

So I think the missing piece is a higher-level way to derive/generate valid sequences of commands from the machine. Maybe enabled + plan are enough as lower-level primitives, assuming the event schemas can also be used to generate arbitrary event values.

thife455:
XState had a very nice approach for generating tests based on the state machine https://stately.ai/docs/xstate-test.

SandroMaglione:
Converting this to draft since I am planning of adding nested/parallel states to the state machine (state charts), and this may require a considerable model change

Meanwhile, looking for feedback on the overall API

tim-smart:
Ideally any state machine primitive we add would integrate cleanly with cluster or rpc. I would add that to the design requirements.

mikearnaldi:
> > quick feedback on naming, I wouldn't use `Actor / ActorSystem` due to potential conflicts in cluster I'd try to scope it more to semantically be tied to `Machine`
> 
> I would rename `Actor` to `Process`, which is how [Erlang](https://www.erlang.org/doc/system/ref_man_processes.html) calls a similar abstraction. And also possibly rename `StateMachine` to `Machine`.
> 
> Since `Actor` is not only for state machines, I would avoid naming to be too tight between the two.
> 
> The problem is that `effect/unstable/process` is already taken by `ChildProcess`. Would it be okay to move `ChildProcess` in a separate path and use `/process` for this new module?
> 
> Another options would be `Interpreter`, but it sounds less good.

That would be even more confusing given that the Erlang model is implemented by Cluster, if this PR has nothing to do with Cluster it should only be scoped to Machine (a self contained module)

IMax153:
Hi @SandroMaglione,

SandroMaglione:
Opened [PR inside `effect`](https://github.com/Effect-TS/effect/pull/6429) main repository

PR #2360, opened 2026-06-10
perf(Scheduler): flush fiber tasks on the microtask queue

mikearnaldi:
## Scheduler

`MixedScheduler` dispatches task flushes on the microtask queue instead of arming a `setImmediate` per flush. A global budget (2048 consecutive microtask flushes across all dispatchers) forces a macrotask fallback so timers and I/O keep running under continuous fiber load; once the budget is exhausted, every dispatcher arms macrotasks until one runs. Dispatchers stay per-fiber, so task callbacks keep capturing each fiber's own `AsyncLocalStorage` context.

Yield-heavy workloads get 3-20x faster (medians, see `benchmark/effect/comprehensive.ts`): request pipeline 443→72µs, producer/consumer 786→47µs, connection pool 1502→65µs, full application 1008→320µs.

Two new context references expose runtime execution state (same pattern as `MaxOpsBeforeYield` / `PreventSchedulerYield`, re-exported from `References`):

- `Scheduler.MacrotaskDispatch`: mutable `{ current: boolean }`; while `true`, a fiber's task flushes are dispatched with `setImmediate`. The dispatcher reads it through the fiber (`fiber.currentMacrotaskDispatch`, cached in `setContext`).
- `Scheduler.AsyncActivity`: mutable `{ resumptions: number }`; incremented by the fiber when it resumes from a suspension on the async primitive (the basis of `Effect.callback`, `Effect.promise`, and all other async operations).

## Fiber

Async resumptions no longer refill the operation budget: `runLoop` does not reset `currentOpCount` on resume; the budget is refilled in the `yieldNow` task when the fiber passes through the scheduler. A fiber looping over async work (`Effect.forever(Effect.promise(...))`) previously never reached the budget and starved the event loop (timers never fired, the fiber was uninterruptible — reproducible on `main`); it now yields after `maxOpsBeforeYield` cumulative operations. Covered by a regression test.

## TestClock

`adjust`/`setTime` set `MacrotaskDispatch.current` for their duration, so fibers run with macrotask dispatch while virtual time advances. Between time steps the clock yields repeatedly until a yield produces no `AsyncActivity` resumptions: async work in flight on the host (e.g. the `crypto.subtle.digest` in workflow id derivation) settles at the current virtual timestamp before time moves. Previously this interleaving was an artifact of per-flush `setImmediate` dispatch; it is now explicit and deterministic. `layer` provides both states, scoped to the clock.

## Sharding

`notifyLocal` waits for the local entity manager only for incoming messages; outgoing notifications are persisted without the registration wait. Previously, completing a `DurableDeferred` for a workflow not registered locally blocked indefinitely once its shard was assigned to the local runner.

## Validation

effect test suite passes (sequential and concurrent mode); remaining monorepo failures are the pre-existing `better-sqlite3` ABI issue, identical on `main`. Changeset included.

### DIFF
--- a/.changeset/scheduler-microtask-dispatch.md
+++ b/.changeset/scheduler-microtask-dispatch.md
@@ -1,0 +1,13 @@
+---
+"effect": patch
+---
+
+Scheduler: flush fiber tasks on the microtask queue instead of `setImmediate`, falling back to a macrotask after a global budget of consecutive microtask flushes so timers and I/O are not starved. Dispatchers remain per-fiber, preserving `AsyncLocalStorage` context propagation. This significantly reduces scheduling latency for yield-heavy workloads (3-20x faster on realistic application benchmarks).
+
+Fiber: async resumptions no longer refill the operation budget; it is refilled when the fiber yields through the scheduler. Fibers that loop over asynchronous work (e.g. `Effect.forever(Effect.promise(...))`) now yield after `maxOpsBeforeYield` cumulative operations instead of monopolizing the event loop.
+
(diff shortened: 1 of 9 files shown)


tim-smart:
Not too sure about this as it makes async things like the HttpServer slower.

IMax153:
Hi @mikearnaldi,

PR #2367, opened 2026-06-10
Fix HttpClient rate limiter Retry-After queueing

### LINKED ISSUE #2311: HttpRequest.withRateLimiter allows already-queued requests to bypass the limit given by Retry-After.

4.0.0-beta.71

```ts
import { BunHttpServer } from "@effect/platform-bun";
import { Clock, Duration, Effect, Layer } from "effect";
import { FetchHttpClient, HttpClient, HttpServer, HttpServerRequest, HttpServerResponse } from "effect/unstable/http";
import { RateLimiter } from "effect/unstable/persistence";
import { BunRuntime } from "@effect/platform-bun";

const serverEffect = Effect.gen(function* () {
	const start = yield* Clock.currentTimeMillis;
	let prev: number | null = null;

	return HttpServer.serve(
		Effect.succeed(HttpServerResponse.empty()).pipe(
			Effect.andThen((r) =>
				Effect.gen(function* () {
					const req = yield* HttpServerRequest.HttpServerRequest;
					const nowAbs = yield* Clock.currentTimeMillis;
					const now = nowAbs - start;

					let outcome = "200";
					if (prev && now - prev < 2000) {
						outcome = "429";
						r = r.pipe(HttpServerResponse.setStatus(429), HttpServerResponse.setHeaders({ "Retry-After": "2" }));
					}

					console.log({
						msg: "received request",
						url: req.url,
						outcome,
						time: now,
						prev,
						delta: prev ? now - prev : null,
					});
					prev = now;

					return r;
				}),
			),
		),
	);
});

	const base = yield* HttpClient.HttpClient;
	const limiter = yield* RateLimiter.RateLimiter;
	const client = base.pipe(
		HttpClient.withRateLimiter({
			limiter,
			key: "some-key",
			disableResponseInspection: false,
			limit: 1,
			window: Duration.millis(1000),
		}),
	);

	const requestCount = 10;

	yield* Effect.forEach(Array(requestCount).fill(0), (_, i) => client.get(`/${i+1}`), {
		concurrency: "unbounded",
		discard: true,
	});
}).pipe(
	Effect.provide(HttpServer.layerTestClient),
	Effect.provide(FetchHttpClient.layer),
	Effect.provide(RateLimiter.layer),
	Effect.provide(RateLimiter.layerStoreMemory),
	Effect.provide(Layer.unwrap(serverEffect), { local: true }),
	Effect.provide(BunHttpServer.layerTest),
	Effect.timeout("5 seconds"),
);

BunRuntime.runMain()(program);
```

TLDR: the above code sets up a server that returns a 429 whenever a request is within 2 seconds of the previous request. The client has a base 1 request per second limit, but has response inspection enabled. 10 requests are "queued" together on startup.

I would expect that the rate limiter would prevent any requests being sent before the "Retry-After" period sent by the server has completed. i.e. the first request would hit the server immediately and get a 200, the second would hit the server after 1 second and get a 429 with a 2 second Retry-After and the server should not receive another request until those 2 seconds have elapsed.

### What do you see instead?

```
// The first request gets a 200.
{
  msg: "received request",
  url: "/1",
  outcome: "200",
  time: 15,
  prev: null,
  delta: null,
}
// The second request gets a 429, since it was 1s between requests, not 2.
{
  msg: "received request",
  url: "/2",
  outcome: "429",
  time: 1006,
  prev: 15,
  delta: 991,
}
// The third request fires after the initial delay it was given - it hasn't been updated by the Retry-After header.
{
  msg: "received request",
  url: "/3",
  outcome: "429",
  time: 2006,
  prev: 1006,
  delta: 1000,
}
// The other requests continue like this, and the server will never give a 200.
{
  msg: "received request",
  url: "/4",
  outcome: "429",
  time: 3012,
  prev: 2006,
  delta: 1006,
}
{
  msg: "received request",
  url: "/5",
  outcome: "429",
  time: 4013,
  prev: 3012,
  delta: 1001,
}
```

### Additional information

_No response_

IMax153:
## Summary
- make queued HttpClient.withRateLimiter requests re-check Retry-After blocks before sending
- preserve delay behavior by sleeping and re-checking current limiter state instead of precomputing stale queued delays
- preserve disabled response inspection behavior
- add regression coverage for queued response and HttpClientError 429 paths

## Notes
The previous implementation used RateLimiter.consume with onExceeded: "delay", which reserved delayed requests up front. That allowed already queued requests to keep an old delay even after a later response supplied Retry-After. This PR intentionally switches the wrapper to fail/sleep/retry acquisition so delayed requests loop back through the latest Retry-After and limiter state before sending.

## Testing
- pnpm lint-fix
- pnpm test packages/effect/test/unstable/http/HttpClient.test.ts
- pnpm check:tsgo

Closes: #2311

### DIFF
--- a/.changeset/fix-http-client-ratelimiter-retry-after.md
+++ b/.changeset/fix-http-client-ratelimiter-retry-after.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `HttpClient.withRateLimiter` so already queued requests respect newly observed `Retry-After` response headers.
--- a/.changeset/fix-http-client-ratelimiter-retry-after.md
+++ b/.changeset/fix-http-client-ratelimiter-retry-after.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-Fix `HttpClient.withRateLimiter` so already queued requests respect newly observed `Retry-After` response headers.
+Fix `HttpClient.withRateLimiter` so already queued requests re-check current limiter state and respect newly observed `Retry-After` response headers. Repository-provided `RateLimiterStore` implementations now store `Retry-After` backoff so clients sharing a limiter key can observe the same server-imposed delay.
--- a/.specs/README.md
+++ b/.specs/README.md
@@ -1,4 +1,5 @@
 # Specifications
 
+- [HttpClient RateLimiter Retry-After Persistence](./http-client-rate-limiter-retry-after.md) — Fixes queued `HttpClient.withRateLimiter` requests by storing `Retry-After` backoff through `RateLimiterStore` top-level `setRetryAfter` / `getRetryAfter` methods without changing the `RateLimiter` interface.
 - [Effect SQL UniqueViolation SqlError Reason](./effect-sql-unique-violation.md) — Adds a `UniqueViolation` SQL error reason with a `constraint` property and updates driver classification for UNIQUE constraint violations.
 - [Effect Platform Crypto Service](./effect-platform-crypto.md) — Adds a platform-agnostic `Crypto` service for cryptographic random bytes, native UUIDv4 generation, and digest operations.
--- a/.specs/http-client-rate-limiter-retry-after.md
+++ b/.specs/http-client-rate-limiter-retry-after.md
@@ -1,0 +1,733 @@
+# HttpClient RateLimiter Retry-After Persistence Specification
+
+## Summary
+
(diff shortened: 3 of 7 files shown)


tim-smart:
Not sure about this as it is in-memory and can't be distributed.

IMax153:
@tim-smart good call - I've re-worked the implementation to persist the retry after information by utilizing the `RateLimiterStore`, if present. 

Let me know what you think of the revised implementation.

PR #2368, opened 2026-06-10
Encode HTTP API path params

IMax153:
## Summary
- encode generated HTTP API client path parameters with encodeURIComponent
- add urlBuilder and generated client regression coverage for slash-containing params
- add a patch changeset

## Validation
- pnpm test packages/effect/test/unstable/httpapi/HttpApiClient.test.ts
- pnpm lint-fix
- pnpm check:tsgo

### DIFF
--- a/packages/effect/src/unstable/httpapi/HttpApiClient.ts
+++ b/packages/effect/src/unstable/httpapi/HttpApiClient.ts
@@ -620,28 +620,25 @@ export const urlBuilder = <Api extends HttpApi.Any>(api: Api, options?: {
 
 // ----------------------------------------------------------------------------
 
-const paramsRegExp = /:(\w+)\??/g
+const paramsRegExp = /(\/?):(\w+)(\?)?/g
 
 const compilePath = (path: string) => {
-  const segments = path.split(paramsRegExp)
-  const len = segments.length
-  if (len === 1) {
+  if (!paramsRegExp.test(path)) {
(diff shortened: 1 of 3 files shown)


packages/effect/src/unstable/httpapi/HttpApiClient.ts
```diff
@@ -634,7 +634,8 @@ const compilePath = (path: string) => {
       if (i % 2 === 0) {
         url += segments[i]
       } else {
-        url += params[segments[i]]
+        const value = params[segments[i]]
+        url += value === undefined ? value : encodeURIComponent(value)
```
gcanti: If `value` is `undefined`, this is effectively:

```ts
url += undefined
// appends "undefined"
```

If optional path params are not supported, I'd make the invariant explicit:

```ts
const value = params[segments[i]]
if (value === undefined) {
  throw new Error(`Missing path parameter: ${segments[i]}`)
}
url += encodeURIComponent(value)
```

IMax153: Implemented in cf50d186: missing path params now throw an explicit `Missing path parameter: <name>` error before encoding. Re-ran `pnpm test packages/effect/test/unstable/httpapi/HttpApiClient.test.ts`, `pnpm lint-fix`, and `pnpm check:tsgo`.

PR #2373, opened 2026-06-11
Swallow reader.cancel rejection in Stream.fromReadableStream finalizer

### LINKED ISSUE #2372: Stream.fromReadableStream: finalizer dies with the raw error when the source stream has errored, clobbering the typed onError failure
When the source `ReadableStream` errors, the typed failure from `onError` gets replaced by a defect carrying the raw error. The mapped error is lost entirely:

```ts
import { Effect, Exit, Stream } from "effect"

const exit = await Effect.runPromiseExit(
  Stream.fromReadableStream({
    evaluate: () =>
      new ReadableStream<number>({
        start(controller) {
          controller.error(new Error("boom"))
        }
      }),
    onError: () => new Error("mapped")
  }).pipe(Stream.runDrain)
)

if (Exit.isFailure(exit)) {
  console.log(exit.cause.reasons.map((r) => r._tag))
  // actual:   ["Die"]  (defect is the raw "boom" error)
  // expected: ["Fail"] (the mapped error)
}
```

The culprit is the finalizer:

https://github.com/Effect-TS/effect-smol/blob/98630b7c8f679c352ba6796636c85688fa009d8d/packages/effect/src/Stream.ts#L1414-L1420

Per spec, `reader.cancel()` on an already-errored stream rejects with the stored error, and `Effect.promise` turns that rejection into a defect. So the same error that was just surfaced through `onError` comes back a second time as a `Die` from the finalizer and clobbers the typed failure.

How we ran into it: we proxy LLM SSE responses (response bodies come in via `HttpClientResponse.stream`, which sits on top of `fromReadableStream`) and guard the body with `Stream.timeoutOrElse` as an idle timeout. When a provider drops the connection mid-stream, this defect doesn't halt `timeoutOrElse`'s merged watchdog promptly — the consumer sees nothing until the idle timeout fires much later, and the real upstream error is misreported as a timeout. With the cancel rejection swallowed, the error propagates immediately.

That second part (a lone `Die` out of a finalizer not stopping the merge in `timeoutOrElse`) might be worth a look on its own — happy to file a separate issue with a repro if useful.

Fix is a one-liner since the error has already been surfaced through `onError` — PR incoming.

Reproduced on current main (98630b7).

StarpTech:
Fixes #2372

`reader.cancel()` rejects with the stored error when the `ReadableStream` has already errored. The finalizer runs it through `Effect.promise`, so the rejection came back as a defect and clobbered the typed failure that `onError` had already produced — `Stream.runDrain` on an errored source ended with `["Die"]` instead of `["Fail"]`.

The finalizer now swallows the cancel rejection. The error has already been surfaced through `onError` at that point, so there's nothing to lose.

Added a regression test asserting an errored source fails with only the mapped error, plus a changeset.

### DIFF
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -1416,7 +1416,10 @@ export const fromReadableStream = <A, E>(
       scope,
       options.releaseLockOnEnd
         ? Effect.sync(() => reader.releaseLock())
-        : Effect.promise(() => reader.cancel())
+        // Cancelling the reader of an already-errored stream rejects with the
+        // stored error - swallow it so the finalizer does not die with a
+        // defect after the error was already surfaced through `onError`.
+        : Effect.promise(() => reader.cancel().catch(constVoid))
     )
     return Effect.flatMap(
       Effect.tryPromise({
--- a/packages/effect/src/Stream.ts
+++ b/packages/effect/src/Stream.ts
@@ -1416,9 +1416,6 @@ export const fromReadableStream = <A, E>(
       scope,
       options.releaseLockOnEnd
         ? Effect.sync(() => reader.releaseLock())
-        // Cancelling the reader of an already-errored stream rejects with the
-        // stored error - swallow it so the finalizer does not die with a
-        // defect after the error was already surfaced through `onError`.
         : Effect.promise(() => reader.cancel().catch(constVoid))
     )
     return Effect.flatMap(
--- a/packages/effect/test/Stream.test.ts
+++ b/packages/effect/test/Stream.test.ts
@@ -238,6 +238,28 @@ describe("Stream", () => {
         assert.deepStrictEqual(result, [2, 4])
         assert.strictEqual(yield* Ref.get(releases), 2)
       }))
+
+    it.effect("fromReadableStream - errored streams fail with the mapped error, not a finalizer defect", () =>
+      Effect.gen(function*() {
+        // Cancelling the reader of an already-errored ReadableStream rejects
+        // with the stored error. The cancel in the finalizer must swallow that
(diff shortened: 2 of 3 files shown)


packages/effect/src/Stream.ts
```diff
@@ -1416,7 +1416,10 @@ export const fromReadableStream = <A, E>(
       scope,
       options.releaseLockOnEnd
         ? Effect.sync(() => reader.releaseLock())
-        : Effect.promise(() => reader.cancel())
+        // Cancelling the reader of an already-errored stream rejects with the
+        // stored error - swallow it so the finalizer does not die with a
+        // defect after the error was already surfaced through `onError`.
+        : Effect.promise(() => reader.cancel().catch(constVoid))
```
tim-smart:
```suggestion
        : Effect.promise(() => reader.cancel().catch(constVoid))
```

packages/effect/test/Stream.test.ts
```diff
@@ -238,6 +238,28 @@ describe("Stream", () => {
         assert.deepStrictEqual(result, [2, 4])
         assert.strictEqual(yield* Ref.get(releases), 2)
       }))
+
+    it.effect("fromReadableStream - errored streams fail with the mapped error, not a finalizer defect", () =>
+      Effect.gen(function*() {
+        // Cancelling the reader of an already-errored ReadableStream rejects
+        // with the stored error. The cancel in the finalizer must swallow that
+        // rejection so the typed `onError` failure is the only cause.
+        const exit = yield* Stream.fromReadableStream({
```
tim-smart, line 244:
```suggestion
      Effect.gen(function*() {
        const exit = yield* Stream.fromReadableStream({
```

PR #2379, opened 2026-06-11
feat: openrouter audio input parts

walln:
## Summary

@effect/ai-openrouter currently converts every non-image file part in a user message into a generic file content block (the document/PDF path). OpenRouter does not accept audio this way — audio must be sent as `input_audio` content parts containing base64 data and a format identifier ([OpenRouter multimodal audio docs](https://openrouter.ai/docs/guides/overview/multimodal/audio)). As a result, audio prompt attachments are rejected or mishandled by upstream providers, even though the generated wire schema (`ChatMessageContentItemAudio`) already supports `input_audio`.

This PR corrects prepareMessages to emit `input_audio` blocks for audio/* file parts:

- Maps audio media types to OpenRouter's documented format set (aac, aiff, flac, m4a, mp3, ogg, pcm16, pcm24, wav), including common aliases (audio/mpeg → mp3, audio/x-wav → wav, audio/mp4 → m4a, etc.). Raw PCM maps from the IANA-registered audio/L16 / audio/L24 types.
- Uint8Array data is base64-encoded; string data reuses getBase64FromDataUrl so both raw base64 and data: URLs work.
- URL data fails with AiError (InvalidUserInputError) — OpenRouter requires base64 and does not fetch audio URLs.
- Unsupported audio/* media types fail with AiError (InvalidUserInputError) listing the supported formats, following the same convention as the Anthropic provider's unsupported-media-type handling.
- cache_control handling is consistent with the sibling text/image/file branches.

Non-audio, non-image files (e.g. PDFs) are unaffected and still use the file block.

Also adds the @effect/ai-openrouter path aliases to the root tsconfig.json, since this package previously had no test directory.

## Test plan

Added packages/ai/openrouter/test/OpenRouterLanguageModel.test.ts (infrastructure modeled on the @effect/ai-openai tests), covering:

- Uint8Array audio → input_audio with base64 data and correct format
- data: URL string audio → prefix stripped, format derived from mediaType
- media type → format mapping across all supported formats
- unsupported audio media type → InvalidUserInputError
- URL audio data → InvalidUserInputError
- PDF file parts still produce file blocks (regression)

Ran pnpm lint-fix, pnpm check:tsgo, and pnpm vitest run across all packages/ai/* test suites (130 passing).

### DIFF
--- a/packages/ai/openrouter/src/OpenRouterLanguageModel.ts
+++ b/packages/ai/openrouter/src/OpenRouterLanguageModel.ts
@@ -765,7 +765,7 @@ const prepareMessages = Effect.fnUntraced(
                       method: "prepareMessages",
                       reason: new AiError.InvalidUserInputError({
                         description: `Detected unsupported media type for audio file: '${part.mediaType}' ` +
-                          "- OpenRouter supports aac, aiff, flac, m4a, mp3, ogg, pcm16, pcm24, and wav audio"
+                          `- OpenRouter supports ${supportedAudioFormats} audio`
                       })
                     })
                   }
@@ -1820,6 +1820,8 @@ const audioFormats: Record<string, string> = {
   "audio/wave": "wav",
   "audio/x-wav": "wav"
 }
+
+const supportedAudioFormats = Array.from(new Set(Object.values(audioFormats))).join(", ")
 
 const getMediaType = (dataUrl: string, defaultMediaType: string): string => {
   const match = dataUrl.match(/^data:([^;]+)/)
--- a/packages/ai/openrouter/src/OpenRouterLanguageModel.ts
+++ b/packages/ai/openrouter/src/OpenRouterLanguageModel.ts
@@ -756,6 +756,45 @@ const prepareMessages = Effect.fnUntraced(
                   break
                 }
 
+                if (part.mediaType.startsWith("audio/")) {
+                  const format = audioFormats[part.mediaType.toLowerCase()]
+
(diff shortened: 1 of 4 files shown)


packages/ai/openrouter/src/OpenRouterLanguageModel.ts
```diff
@@ -756,6 +756,45 @@ const prepareMessages = Effect.fnUntraced(
...
+                if (part.mediaType.startsWith("audio/")) {
+                  const format = audioFormats[part.mediaType.toLowerCase()]
+
+                  if (Predicate.isUndefined(format)) {
+                    return yield* AiError.make({
+                      module: "OpenRouterLanguageModel",
+                      method: "prepareMessages",
+                      reason: new AiError.InvalidUserInputError({
+                        description: `Detected unsupported media type for audio file: '${part.mediaType}' ` +
+                          "- OpenRouter supports aac, aiff, flac, m4a, mp3, ogg, pcm16, pcm24, and wav audio"
```
IMax153: To avoid the list of supported audio formats going stale, we should derive and dedupe them from the `audioFormats` record below. This way we only need to maintain them in one place.

IMax153 (changes requested):

packages/ai/openrouter/src/OpenRouterLanguageModel.ts
walln: Ah good point. Should be fixed now!

IMax153 (approved):

PR #2382, opened 2026-06-12
Fix searchParam history state handling

jacobiajohnson:
## Summary
- Preserve existing `history.state` metadata when `Atom.searchParam` pushes a debounced URL update, including TanStack Router `__TSR_index` / key fields.
- Discard pending writes when the route changes before the debounce flush or when the user navigates via `popstate`.
- Avoid pushing no-op updates.

## Verification
- `pnpm vitest run packages/effect/test/reactivity/Atom.test.ts`


jacobiajohnson:
Closing this for now while I continue validating whether these upstream changes are needed.

PR #2399, opened 2026-06-12
Fail with typed SqlError when sqlite-bun statement preparation throws

### LINKED ISSUE #2385: `@effect/sql-sqlite-bun` mishandles driver errors

4.0.0-beta.80

Run with `bun --bun run reproduction.ts`  
```typescript
import { SqliteClient } from '@effect/sql-sqlite-bun';
import * as Effect from 'effect/Effect';
import { SqlClient } from 'effect/unstable/sql/SqlClient';

	const sql = yield* SqlClient;
	yield* sql`insert into nonexistent (id) values (1)`.pipe(Effect.result);
}).pipe(Effect.provide(SqliteClient.layer({ filename: ':memory:' })));

try {
	await Effect.runPromise(program.pipe(Effect.result));
	console.log('Executed as expected');
} catch (e) {
	console.log('Error thrown:');
	console.error(e);
}
```

Expected to log `Executed as expected`

### What do you see instead?

```
Error thrown:
63 |     yield* Effect.addFinalizer(() => Effect.sync(() => db.close()));
64 |     if (options.disableWAL !== true) {
65 |       db.run("PRAGMA journal_mode = WAL;");
66 |     }
67 |     const run = (sql, params = []) => Effect.withFiber(fiber => {
68 |       const statement = db.query(sql);
     ^
SQLiteError: no such table: nonexistent
      errno: 1,
 byteOffset: -1,
```

### Additional information

Moving `try` in `run`, `runValues` to capture `db.query(sql)` resolves the issue

truffle-dev:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Closes #2385.

In `@effect/sql-sqlite-bun`, `run` and `runValues` call `db.query(sql)` before the `try` block. Bun's `db.query` prepares the statement eagerly, so a prepare-time driver error (missing table, syntax error) throws synchronously inside `Effect.withFiber` and escapes as a defect instead of landing in the error channel as a typed `SqlError` — it blows past `Effect.result` and crashes `runPromise`, as the issue repro shows.

The sibling clients already handle this: `sqlite-node` wraps `db.prepare` in `Effect.try` with a `"Failed to prepare statement"` / `"prepare"` classification, and `sqlite-wasm` / `sqlite-do` wrap the whole run. This change closes the asymmetry by moving statement preparation (including the `safeIntegers` call) into a guarded step that fails with `classifySqliteError(cause, { message: "Failed to prepare statement", operation: "prepare" })`, mirroring the node client's semantics. Execute-phase classification is unchanged.

Verified with the issue's repro under `bun` against this branch: before, the raw `SQLiteError: no such table` is thrown at `SqliteClient.ts:134`; after, `Effect.result` captures `SqlError` with `operation: "prepare"`. Also verified the happy path and that an execute-phase unique violation still classifies as `UniqueViolation` / `operation: "execute"`. No CI-runnable regression test added because the test matrix runs Node and Deno only, where `bun:sqlite` cannot load (the package's test file is a stub for the same reason).

`pnpm --filter @effect/sql-sqlite-bun check`, `oxlint`, and `dprint check` are clean; changeset included.

### DIFF
--- a/.changeset/sqlite-bun-prepare-error-channel.md
+++ b/.changeset/sqlite-bun-prepare-error-channel.md
@@ -1,0 +1,5 @@
+---
+"@effect/sql-sqlite-bun": patch
+---
+
+Fail with a typed `SqlError` when Bun SQLite statement preparation throws (for example a missing table or a syntax error), instead of letting the driver error escape as a defect, closes #2385.
--- a/packages/sql/sqlite-bun/src/SqliteClient.ts
+++ b/packages/sql/sqlite-bun/src/SqliteClient.ts
@@ -139,14 +139,8 @@ export const make = (
       ) =>
         Effect.withFiber<Array<any>, SqlError>((fiber) => {
           const useSafeIntegers = Context.get(fiber.context, Client.SafeIntegers)
-          let statement: ReturnType<typeof db.query>
           try {
-            statement = prepare(sql, useSafeIntegers)
-          } catch (cause) {
-            return Effect.fail(new SqlError({ reason: classifyError(cause, "Failed to prepare statement", "prepare") }))
-          }
-          try {
-            return Effect.succeed((statement.all(...(params as any)) ?? []) as Array<any>)
+            return Effect.succeed((prepare(sql, useSafeIntegers).all(...(params as any)) ?? []) as Array<any>)
           } catch (cause) {
             return Effect.fail(new SqlError({ reason: classifyError(cause, "Failed to execute statement", "execute") }))
           }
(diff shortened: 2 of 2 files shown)


truffle-dev:
This has been open a week with no CI run — fork PRs sit behind the approve-and-run gate. One thing worth flagging before anyone clicks it: the checks won't exercise this path anyway, since the matrix is Node + Deno and `bun:sqlite` can't load there (the package's own test file is a stub for the same reason). So the before/after `bun` repro in the description is the regression evidence, not CI. The branch is still clean against current main; happy to rebase or reshape the `prepare` classification wording if you'd prefer it differently.

PR #2415, opened 2026-06-15
Fix `Effect.try` thunk usage and `Effect.tryPromise` mapper and signal handling defects, closes #2406

gcanti:


IMax153 (dismissed):
This looks good to me - I would make `Effect.try` consistent though with the `Effect.tryPromise` overloads (if we can).

PR #2428, opened 2026-06-17
Add isOpen method to Latch to query latch position

MrGovindan:
## Type
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adds `isOpen` to Latch to allow querying of latch position

### DIFF
--- a/packages/effect/src/Latch.ts
+++ b/packages/effect/src/Latch.ts
@@ -11,14 +11,6 @@ import type * as Effect from "./Effect.ts"
 import * as internal from "./internal/effect.ts"
 
 /**
- * Represents the position of a `Latch`, which can be either `"open"` or `"closed"`.
- *
- * @category models
- * @since 4.0.0
- */
-export type LatchPosition = "open" | "closed"
-
-/**
  * A reusable coordination primitive that lets fibers wait until they are
  * released by the latch.
  *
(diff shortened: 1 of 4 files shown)


packages/effect/src/Latch.ts
```diff
@@ -113,6 +121,15 @@ export interface Latch {
...
   whenOpen<A, E, R>(self: Effect.Effect<A, E, R>): Effect.Effect<A, E, R>
+
+  /**
+   * Polls the latch to check its state.
+   *
+   * ** When to use**
+   *
+   * Use to check the state of the latch without suspending or changing its state.
+   */
+  poll(this: Latch): LatchPosition
```
tim-smart: I think `isOpen(): boolean` is a better api

JesseGovindan: I've changed the api to match.

packages/effect/test/Latch.test.ts
```diff
@@ -18,4 +18,16 @@ describe("Latch", () => {
 
       assert.isDefined(waiter.pollUnsafe())
     }))
+
+  it.effect("poll reflects the state of the latch", () =>
```
tim-smart:
```suggestion
  it.effect("isOpen reflects the state of the latch", () =>
```

PR #2435, opened 2026-06-18
Fix Postgres runner storage lock expiration

hsyntax:
Fixes the Postgres shard lock expiration SQL and covers the expiration behavior in the platform-node Postgres suite.

*I used Codex to diagnose and raise this PR*


hsyntax:
Apologies - the core bug that I was dealing with actually came from a local config mismatch. Closing this PR.

Issue #2427, opened 2026-06-17
Redis PersistedQueueStore `offer`  is pushing in the wrong order


julia-script:
### What version of Effect is running?

4.0.0-beta.83

### What steps can reproduce the bug?

```ts
import { NodeRedis } from "@effect/platform-node";
import { Effect, Layer, Schema } from "effect";
import { PersistedQueue, Redis } from "effect/unstable/persistence";
import * as PersistedQueueFactory from "effect/unstable/persistence/PersistedQueue";

const redisStore = PersistedQueue.layerStoreRedis().pipe(Layer.provide(redis));
const factory = PersistedQueueFactory.layer.pipe(Layer.provide(redisStore));
const layer = Layer.mergeAll(redis, factory).pipe(Layer.provide(redis));
const program = Effect.gen(function* () {
  const redis = yield* Redis.Redis;
  yield* redis.send("FLUSHALL");

  const queue = yield* PersistedQueue.make({
    name: "queue",
    schema: Schema.Number,
  });
  yield* queue.offer(1);
  yield* queue.offer(2);
  yield* queue.offer(3);
  const job = yield* queue.take(Effect.succeed);
  yield* Effect.log(job); // logs "1" with sqlite and memory, but logs "3" with redis
 
}).pipe(Effect.provide(layer));

await Effect.runPromise(program);
```

### What is the expected behavior?

Redis is operating as LIFO

### What do you see instead?

Both memory and sqlite PersistedQueueStore operates as FIFO, so I imagine Redis is the wrong one out of the three.

### Additional information

I suspect the problem is with is calling RPUSH instead of a LPUSH here, so the head is always the oldest item (or pop the last item on `take` )
https://github.com/Effect-TS/effect-smol/blob/2ba316bd15fcbf1c50626500d44a2c9b3bec19f5/packages/effect/src/unstable/persistence/PersistedQueue.ts#L511

as a side note I wonder if it wouldn't make sense to support both LIFO and FIFO?

Edit: after looking at it a bit more, I noticed that when Ids are specified, it uses the offer lua script, which does actually RPUSH as expected
https://github.com/Effect-TS/effect-smol/blob/2ba316bd15fcbf1c50626500d44a2c9b3bec19f5/packages/effect/src/unstable/persistence/PersistedQueue.ts#L600

so this code actually logs `1`
```ts
  yield* queue.offer(1,{
    id: '1'
  });
  yield* queue.offer(2, {
    id: '2'
  });
[...]

PR #2436, opened 2026-06-18
[Redis][Queue]:change default operation in redis from LPUSH TO RPUSH

MohanedMashaly:
## Type
- ✅  Bug Fix

## Description
As reported in the below issue, the default behavior in case of not passing a custom ID to persistedQueue using redis as storage layer is LIFO basis which is against the default behavior, this PR make the default operation FIFO, we can add a flag to select wether a queue should add on LIFO or FIFO as an option.

## Related

- Related Issue : https://github.com/Effect-TS/effect-smol/issues/2427

### DIFF
--- a/.changeset/every-olives-burn.md
+++ b/.changeset/every-olives-burn.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+change default operation in redis from LPUSH TO RPUSH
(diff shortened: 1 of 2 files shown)


.changeset/every-olives-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": minor
```
tim-smart:
```suggestion
"effect": patch
```

Issue #2437, opened 2026-06-18
Broken Config definition for openai-compat LanguageModel


aedrondouren:
### What version of Effect is running?

4.0.0-beta.84

### What steps can reproduce the bug?

```ts
import { OpenAiLanguageModel } from '@effect/ai-openai-compat';

const model = OpenAiLanguageModel.model('<MODEL>', { });
//         No auto-completion when providing config ^
```

### What is the expected behavior?

Auto-completion when providing default fields like `temperature`, `top_p` and `reasoning`.

### What do you see instead?

Auto-completion with every known symbol in the project.

### Additional information

```ts
export class Config extends Context.Service<
  Config,
  Simplify<
    & Partial<
      Omit<
        CreateResponse,
        "input" | "tools" | "tool_choice" | "stream" | "text"
      >
    >
    & {
      readonly fileIdPrefixes?: ReadonlyArray<string> | undefined
      readonly text?: {
        readonly verbosity?: "low" | "medium" | "high" | undefined
      } | undefined
      readonly strictJsonSchema?: boolean | undefined
      readonly [x: string]: unknown
    }
  >
>()("@effect/ai-openai-compat/OpenAiLanguageModel/Config") {}
```

The `readonly [x: string]: unknown` key breaks auto-completion. Removing it solves the issue, but I assume this was added to allow custom fields for other OpenAI compatible model providers.

aedrondouren:
Considering how TypeScript works, I believe it might be best to have a dictionary for those, otherwise I think auto-completion becomes impossible.

IMax153:
I believe this may have been fixed by Tim's work here: https://github.com/Effect-TS/effect-smol/pull/2608

The Effect v4 codebase has now been merged back into `effect` and we are working to move relevant issues / pull requests over to the `effect` repository.

Please re-open this issue in the `effect` repository (https://github.com/Effect-TS/effect) if this issue has not been fully addressed.

PR #2438, opened 2026-06-19
Restore autocomplete for openai-compat model config

aedrondouren:
## Type

* [x] Bug Fix

## Description

Remove the top-level string index signature from `OpenAiLanguageModel.Config`.

This restores IntelliSense/autocomplete for known config fields in editors.

## Notes

This change focuses only on the typing/autocomplete issue and does not address the broader question of how arbitrary
provider-specific fields should be represented.

## Related

* Related Issue: #2437


packages/ai/openai-compat/src/OpenAiLanguageModel.ts
```diff
@@ -112,7 +112,6 @@ export class Config extends Context.Service<
        * Defaults to `true`.
        */
       readonly strictJsonSchema?: boolean | undefined
-      readonly [x: string]: unknown
```
tim-smart, line 115: You should be allowed to pass any field here to support provider custom config.

aedrondouren: Yeah that's what I assumed was the goal, but with the nature of typescript as mentioned in my original [issue](https://github.com/Effect-TS/effect-smol/issues/2437#issuecomment-4745635702), this might need to be broken out in a dedicated field on the Config object to preserve autocompletion.

What do we think of a dedicated `custom: Record<string, unknown>` field on `Config`?

aedrondouren:
I've noticed the same thing with the embedding model config [here](https://github.com/Effect-TS/effect-smol/blob/main/packages/ai/openai-compat/src/OpenAiEmbeddingModel.ts#L49-L62), we could use the same approach if you guys like it.

tim-smart:
https://github.com/Effect-TS/effect-smol/pull/2608#issuecomment-4953179957

PR #2443, opened 2026-06-19
Add Svelte Atoms package

thomasfosterau:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds an `@effect/atom-svelte` package for using `Atom` within Svelte, equivalent to `@effect/atom-react`, `@effect/atom-vue` or `@effect/atom-solid`.

This uses Svelte’s `createSubscriber()` function from the [`svelte/reactivity`](https://svelte.dev/docs/svelte/svelte-reactivity) module, which has been available since Svelte 5.7.0. It would be possible to implement this directly with Svelte 5’s runes but it adds build-time complexity for little gain. This implementation is still fully compatible with runes.

Currently this PR does not use the [Hydratable API](https://svelte.dev/docs/svelte/hydratable) which Svelte introduced in 5.44.0. I’m not sure if this is the right place to be using it. Ditto for the [`getAbortSignal()`](https://svelte.dev/docs/svelte/svelte#getAbortSignal) function. Happy to have a go at including either of these if desired.

Tests pass but Svelte’s compiler throws a bunch of warnings; as far as I can see they’re false positives. An unrelated test for one of the SQL packages is failing on my machine, possibly as I was running in a container.

Disclosure: this was written using Claude Code, but has been reviewed by myself before opening a PR.

## Related

- Related Issue # N/A
- Closes # N/A


packages/atom/svelte/src/RegistryContext.ts
```diff
@@ -0,0 +1,117 @@
...
+import * as Hydration from "effect/unstable/reactivity/Hydration"
+import { getContext, setContext } from "svelte"
+
+/**
+ * The Svelte context key under which the `AtomRegistry` is stored.
+ *
+ * @category registry
+ * @since 4.0.0
+ */
+export const registryKey: symbol = Symbol.for("@effect/atom-svelte/registryKey")
```
thomasfosterau, line 20: Svelte's `createContext` API would actually be better here — will update.

packages/atom/svelte/src/RegistryContext.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Reads the `AtomRegistry` from Svelte context, falling back to
+ * {@link defaultRegistry}. Must be called during component initialisation.
+ *
+ * @category registry
+ * @since 4.0.0
+ */
+export const injectRegistry = (): AtomRegistry.AtomRegistry =>
+  getContext<AtomRegistry.AtomRegistry | undefined>(registryKey) ?? defaultRegistry
```
thomasfosterau, line 65: Switching to `createContext` above would require changes here because getting uninitialised context would throw.

IMax153:
Hi @thomasfosterau,

PR #2444, opened 2026-06-19
fix: support frozen intrinsics when adjusting Error.stackTraceLimit

bweis:
Ports [Effect-TS/effect#6279](https://github.com/Effect-TS/effect/pull/6279) to `effect-smol`.

## Type

- [x] Bug Fix

## Description

In hardened/deterministic JavaScript environments (SES "frozen intrinsics", Temporal-style sandboxes), `Error` is frozen and `Error.stackTraceLimit` is read-only. Effect mutates `stackTraceLimit` in several internal spots to capture short or empty stack traces cheaply, using the pattern:

```ts
const limit = Error.stackTraceLimit
Error.stackTraceLimit = 2
const err = new Error()
Error.stackTraceLimit = limit
```

In these environments the assignments throw, breaking Effect entirely.

This PR adds `packages/effect/src/internal/stackTraceLimit.ts`, which checks writability once at module load (mirroring [Node's internal guard](https://github.com/nodejs/node/blob/e77694631f1642c302f664703197b5aabc65b482/lib/internal/errors.js#L246)) and exposes `getStackTraceLimit()` / `setStackTraceLimit()` helpers where `set` is a silent no-op when the property can't be modified. All raw mutations are replaced with these guarded calls, with `new Error()` kept inline at each call site so the captured stack trace still points at the real caller rather than the helper.

Call sites updated: `Context.ts`, `Layer.ts`, `LayerMap.ts`, `internal/effect.ts` (cause pretty-errors, `Effect.fn`), `internal/tracer.ts`, `unstable/httpapi/HttpApiMiddleware.ts`, `unstable/rpc/RpcMiddleware.ts`.

### Notes for review (vs. the upstream `effect` PR)

The `effect-smol` layout differs from `Effect-TS/effect`, so this is a port rather than a cherry-pick:

- Internals are consolidated (`internal/effect.ts`, public `Context.ts`/`Layer.ts`/`LayerMap.ts`) rather than split across `cause`/`core-effect`/`runtime`/`layer`/`context`; there is no `Micro.ts`.
- The new helper reuses the existing `ErrorWithStackTraceLimit` interface and `.ts` import extensions.
- `unsafeSecureJsonParse` (`unstable/ai/Tool.ts`) imports the internal helper directly instead of duplicating the guard, because `ai` lives in the same package here (unlike `@effect/ai` upstream).

Behavior in normal (writable) environments is unchanged. A changeset and a dedicated test (`packages/effect/test/StackTraceLimit.test.ts`, covering both the writable and frozen paths) are included; typecheck and lint pass and the affected suites (Cause, Tracer, Context, LayerMap, Effect) are green.

> [!NOTE]
> The original change was stood up by @arlyon. This description and PR were aided via Claude Code.

### DIFF
--- a/.changeset/frozen-intrinsics-stack-trace-limit.md
+++ b/.changeset/frozen-intrinsics-stack-trace-limit.md
@@ -1,0 +1,7 @@
+---
+"effect": patch
+---
+
+Avoid throwing when `Error.stackTraceLimit` is non-writable (frozen intrinsics / SES / deterministic sandboxes such as Temporal).
+
+Effect manipulates `Error.stackTraceLimit` in several internal spots to capture short or empty stack traces cheaply. In hardened environments where `Error` is frozen and `stackTraceLimit` is read-only, assigning to it throws, which broke Effect entirely. Stack-trace-limit manipulation is now best-effort and silently no-ops when the property cannot be modified, mirroring Node's own internal guard. Behavior in normal (writable) environments is unchanged.
--- a/packages/effect/src/Context.ts
+++ b/packages/effect/src/Context.ts
@@ -17,7 +17,7 @@ import { dual, type LazyArg } from "./Function.ts"
 import * as Hash from "./Hash.ts"
 import type { Inspectable } from "./Inspectable.ts"
 import { exitSucceed, PipeInspectableProto, withFiber } from "./internal/core.ts"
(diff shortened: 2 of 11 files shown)


bweis:
effect-smol equivalent of https://github.com/Effect-TS/effect/pull/6279

PR #2451, opened 2026-06-22
Preserve external span sampling in OpenTelemetry

StarpTech:
Generic `Tracer.externalSpan` values carry their sampling decision in `span.sampled`. The OpenTelemetry adapter currently defaults to `SAMPLED` when no `OtelTraceFlags` annotation is present, so an external span created with `sampled: false` becomes sampled during conversion.

A common case is an Effect HTTP service that extracts an incoming `traceparent` header into a generic external span, then exports through the OpenTelemetry SDK with a parent-based sampler. If the upstream service chose not to sample the trace, the adapter currently turns that remote parent into a sampled OTel context. The parent-based sampler then correctly follows the wrong input and records a child span that should have been dropped.

The same issue applies to custom propagators and integrations that construct `Tracer.externalSpan` directly rather than using the OpenTelemetry-specific constructor.

This change uses `span.sampled` as the fallback while keeping explicit OpenTelemetry trace flags authoritative. It also adds a regression test covering an unsampled generic external parent.


StarpTech:
@tim-smart ptal thanks!

tim-smart (approved):

PR #2454, opened 2026-06-22
Strip response metadata from HTTP server span failures

StarpTech:
## Problem

When a streamed HTTP response fails after its headers have been sent, the Node server keeps the original `HttpServerResponse` in the failure cause. This lets the platform finish the request with the response that was already started.

The HTTP tracer currently ends the server span with that full cause. OpenTelemetry can then pick the response as the span status message and render the whole object, including headers and body metadata, instead of showing the stream error. We hit this with an SSE endpoint: the request correctly had HTTP status 200, but its failed span displayed a serialized `HttpServerResponse` as the error.

## Fix

Use `causeResponseStripped` when ending the HTTP server span as well as when reading response attributes. The response still supplies status and header attributes, while the span failure only contains the underlying application or stream error. If the response was the only cause entry, the span ends successfully.

## Test

The regression test models a stream failure after headers were sent and checks that:

- the span still records HTTP status 200
- the span remains failed with the original stream error
- the sent response is not part of the span exit


StarpTech:
@tim-smart ptal thanks!

tim-smart (approved):

PR #2457, opened 2026-06-22
Scope unknown request tag failures to the request instead of the connection

P0lip:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The serialized server run-loop rejected an unknown rpc tag with a connection-level defect (sendDefect, no requestId). The client turns a connection-level defect into clearEntries, which fails every in-flight request and stream on the shared multiplexed connection. A single unroutable tag therefore tore down all other live requests.

Fix: parse the requestId before the group.requests.get(tag) lookup so the unknown-tag branch can fail just that request via sendRequestDefect instead of the whole connection.

Test multiplexes a live stream and an unroutable request over one connection and asserts the routing miss fails alone while the stream keeps emitting.


.changeset/fix-rpc-unknown-tag-isolation.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix `RpcServer` tearing down unrelated in-flight requests when a single request used an unknown tag. Previously an unrecognized request tag was rejected with a connection-level defect, which the client turned into `clearEntries` and failed every concurrent request and stream sharing the multiplexed connection. The unknown tag now fails only its own request, leaving other live requests and streams on the connection untouched.
```
tim-smart:
```suggestion
Localize missing rpc method errors to the provided request id
```

PR #2477, opened 2026-06-24
Add opt-in Undici request body release

StarpTech:
## Motivation

Large uploads followed by long-lived streaming responses can retain the completed request body for the full response lifetime when using the Undici client.

In that workload, retained memory grows with:

```text
active streaming responses * request body size
```

This PR adds an opt-in way to release that memory.

## Change

```ts
client.execute(request).pipe(
  Effect.provideService(NodeHttpClient.RetainResponseRequestBody, false)
)
```

The option defaults to `true`, so this is not a breaking change. Existing users keep the current behavior and `response.request.body` still contains the original body.

When disabled:

- successful responses keep request method, URL fields, and headers
- `response.request.body` is empty
- the caller-owned request is unchanged
- transport errors before response headers still keep the original request

For byte-array uploads of at least 64 KiB, the Undici adapter uses a one-shot `Readable` and drops its byte-array reference after upload. Smaller bodies keep the existing direct path. The opt-in response API is the same for both sizes.

## Benchmark

Node 24.15.0, local Undici server, five alternating runs per variant. Responses remained open after upload; retained memory was measured after diagnostic GC.

**Main result:** the opt-out removes almost exactly the uploaded payload total from live memory: about **100 MiB** for 100 x 1 MiB uploads and **160 MiB** for 40 x 4 MiB uploads.

### 100 concurrent 1 MiB uploads

| Metric | Default | Opt-out | Gain |
| --- | ---: | ---: | ---: |
| Throughput | 1,451 req/s | 1,451 req/s | unchanged |
| Retained ArrayBuffers | 108.1 MiB | 8.1 MiB | **100 MiB / 92.5% lower** |
| Retained external | 132.5 MiB | 32.5 MiB | **100 MiB / 75.5% lower** |
| Retained RSS | 306.7 MiB | 286.6 MiB | **20.1 MiB / 6.5% lower** |

### 40 concurrent 4 MiB uploads

| Metric | Default | Opt-out | Gain |
| --- | ---: | ---: | ---: |
| Throughput | 509 req/s | 498 req/s | 2.0% lower |
| Retained ArrayBuffers | 168.1 MiB | 8.1 MiB | **160 MiB / 95.2% lower** |
| Retained external | 192.5 MiB | 32.5 MiB | **160 MiB / 83.1% lower** |
| Retained RSS | 373.7 MiB | 368.2 MiB | **5.5 MiB / 1.5% lower** |

RSS falls less than live external memory because V8 and the native allocator keep freed pages available for reuse. The uploaded payloads are no longer live and cannot accumulate with active responses.

A separate small-body benchmark found that wrapping every 1 KiB upload in a `Readable` reduced throughput by about 10%, so bodies below 64 KiB stay on the current fast path. Default behavior has no new overhead.

## Scope / open question

This only covers the Node.js Undici client. Fetch and native `node:http` have different retention paths and should be measured separately.

I am not sure whether this policy belongs upstream as an opt-in, or whether Effect should instead document the application-level workaround for long-lived streaming responses.

Without the new API, an application can avoid retaining the uploaded bytes by using a one-shot request stream and clearing its source after the HTTP client pulls it:

```ts
const encoder = new TextEncoder()

const releasingBody = (input: string) =>
  Stream.fromEffect(
    Effect.sync(() => {
      const bytes = encoder.encode(input)
      input = ""
      return bytes
    })
  )

const body = JSON.stringify(payload)
const request = HttpClientRequest.post(url).pipe(
  HttpClientRequest.bodyStream(releasingBody(body), {
    contentType: "application/json",
    contentLength: Buffer.byteLength(body)
  })
)
```

This releases the source string after upload, but `response.request.body` still contains the stream object. The proposed option handles this centrally and also keeps request metadata without retaining the body.

## Validation

- `pnpm lint-fix`
- `pnpm test packages/platform-node/test/NodeHttpClient.test.ts` (27 passed)
- `pnpm --filter @effect/platform-node build:tsgo`

Tests cover default compatibility, opt-in behavior for small and large bodies, exact payload delivery while the response stays open, preserved metadata, and scoped server cleanup.

### DIFF
--- a/.changeset/fix-node-http-response-body-retention.md
+++ b/.changeset/fix-node-http-response-body-retention.md
@@ -1,5 +1,5 @@
 ---
-"@effect/platform-node": patch
+"@effect/platform-node": minor
 ---
 
-Avoid retaining large uploaded byte-array request bodies for the lifetime of Undici HTTP client responses. Successful responses preserve request metadata but expose an empty request body.
+Add `NodeHttpClient.RetainResponseRequestBody`, which can be disabled to avoid retaining large uploaded byte-array bodies for the lifetime of Undici responses. The existing response API remains unchanged by default; when disabled, successful responses preserve request metadata but expose an empty request body.
--- a/.changeset/fix-node-http-response-body-retention.md
+++ b/.changeset/fix-node-http-response-body-retention.md
@@ -1,0 +1,5 @@
+---
+"@effect/platform-node": patch
+---
+
+Avoid retaining large uploaded byte-array request bodies for the lifetime of Undici HTTP client responses. Successful responses preserve request metadata but expose an empty request body.
--- a/packages/platform-node/src/NodeHttpClient.ts
+++ b/packages/platform-node/src/NodeHttpClient.ts
@@ -21,10 +21,10 @@ import type * as Scope from "effect/Scope"
 import * as Stream from "effect/Stream"
 import * as Cookies from "effect/unstable/http/Cookies"
 import * as Headers from "effect/unstable/http/Headers"
-import type * as Body from "effect/unstable/http/HttpBody"
+import * as Body from "effect/unstable/http/HttpBody"
 import * as Client from "effect/unstable/http/HttpClient"
 import * as Error from "effect/unstable/http/HttpClientError"
-import type { HttpClientRequest } from "effect/unstable/http/HttpClientRequest"
+import { type HttpClientRequest, makeWith } from "effect/unstable/http/HttpClientRequest"
 import * as Response from "effect/unstable/http/HttpClientResponse"
 import type { HttpClientResponse } from "effect/unstable/http/HttpClientResponse"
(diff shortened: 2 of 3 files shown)


mikearnaldi:
Feels too ad-hoc, you can use `HttpClient.withScope` to control the lifecycle of a request explicitely:

```ts
import { HttpClient, HttpClientRequest } from "effect/unstable/http"

export const program = Effect.gen(function*() {
  const client = HttpClient.withScope(yield* HttpClient.HttpClient)

  const response = yield* Effect.scoped(
    client.execute(HttpClientRequest.get("..."))
      .pipe(Effect.flatMap((_) => _.text))
  )

  console.log(response)
})
```

IMax153:
Hi @StarpTech,

StarpTech:
@mikearnaldi this doesn't fix it for long-lived responses. I benchmarked it. withScope is useful for cleanup of completed or abandoned responses. It is not a solution for releasing an uploaded body while preserving a long-lived response.

PR #2479, opened 2026-06-25
EFF-875 Add custom errors to Effect.fromOption

tim-smart:
## Summary
- Add optional custom onNone error support to Effect.fromOption in data-first and data-last forms.
- Preserve the default NoSuchElementError behavior when onNone is omitted.
- Add runtime and type-level coverage, including higher-order callback usage.

## Validation
- pnpm lint-fix
- pnpm test packages/effect/test/Effect.test.ts
- pnpm test-types packages/effect/typetest/Effect.tst.ts
- pnpm check:tsgo

### DIFF
--- a/packages/effect/test/Effect.test.ts
+++ b/packages/effect/test/Effect.test.ts
@@ -176,22 +176,22 @@ describe("Effect", () => {
         Effect.runPromise
       ))
 
-    it.effect("from a none with a custom error", () => {
-      const error = new Error("Missing value")
-      return Effect.fromOption(Option.none(), () => error).pipe(
-        Effect.flip,
-        Effect.tap((cause) => Effect.sync(() => assert.strictEqual(cause, error)))
-      )
-    })
+    it.effect("from a none with a custom error", () =>
+      Effect.gen(function*() {
+        const error = new Error("Missing value")
+        const cause = yield* Effect.fromOption(Option.none(), () => error).pipe(Effect.flip)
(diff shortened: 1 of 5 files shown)


packages/effect/test/Effect.test.ts
```diff
@@ -175,6 +175,25 @@ describe("Effect", () => {
         Effect.tap((error) => Effect.sync(() => assert.ok(error instanceof Cause.NoSuchElementError))),
         Effect.runPromise
       ))
+
+    it("from a none with a custom error", () => {
```
tim-smart: These tests should use it.effect and remove the .runPromise

packages/effect/test/Effect.test.ts
```diff
@@ -176,22 +176,20 @@ describe("Effect", () => {
...
       ))
 
-    it("from a none with a custom error", () => {
+    it.effect("from a none with a custom error", () => {
       const error = new Error("Missing value")
       return Effect.fromOption(Option.none(), () => error).pipe(
         Effect.flip,
-        Effect.tap((cause) => Effect.sync(() => assert.strictEqual(cause, error))),
-        Effect.runPromise
+        Effect.tap((cause) => Effect.sync(() => assert.strictEqual(cause, error)))
```
tim-smart: Instead of Effect.tap, you should instead just use normal variable assignment `const cause = yield* ...`

PR #2482, opened 2026-06-25
feat: add graph set operators

lloydrichards:
## Type

- [x] Refactor
- [x] Feature

## Description

Have added 4 basic and 3 advanced set operators to the Graph module to be able to build and explore for interesting structures. These include:

- `Graph.compose` - $(G1 ∪ G2 = { V1 ∪ V2, E1 ∪ E2 })$
  merges two graphs by node identity, combining nodes and edges from both. Overlapping data from that wins.
- `Graph.intersection` - $(G1 ∩ G2 = { V1 ∩ V2, E1 ∩ E2 })$
  keeps only nodes and edges present in both graphs.
- `Graph.difference` - $(G1 \ G2 = { V1, E1 \ E2 })$
  keeps all nodes from self, but removes edges that also exist in that.
- `Graph.symmetricDifference` - $(G1 Δ G2 = { V1 ∪ V2, (E1 ∪ E2) \ (E1 ∩ E2) })$
  keeps edges that exist in exactly one graph, removing shared edges.
- `Graph.complement` - $(G' = { V, (V × V) \ E })$
  keeps the same nodes, but adds every missing non-self edge.
- `Graph.neighborhood` - $(Nᵣ(v) = G[{ u ∈ V | distance(v, u) ≤ r }])$
  builds an induced subgraph around a node within a radius.
- `Graph.sum` - $(G1 + G2 = { V1 ⊔ V2, E1 ⊔ E2 })$
  creates a disjoint union of two graphs without merging equal nodes.

Most of the implementation was done similar to `NetworkX` when it came to making decisions around when `self` or `that` would win on conflict and the main decision on node equality is left to the implementor where they need to pass a nodeId callback to be able to identify each node.

One additional refactor (separate commit) was the additional logic to `Graph.bfs` and `Graph.dfs` to allow for a `maxDepth`.  This was important to be able to reuse the BFS in the `neighbordhood` op and create a custom version like I did in my utils experiment.

Some fun and exciting things can be done with this functionality.

Screenshot 2026-06-23 at 22 26 49

## For Reviewing

To make things easier to review I've broken up the three main parts so they can be reviewed separately and left off the type tests till last. Any feedback or requests I will initial add as commits on top of the current tree and then rebase together to clean up once approved

## Related

Came up as a request in Slack with a bit of back and forth discussion with @fubhy.  I then spend a couple of days building [similar functions in my own repo](https://github.com/lloydrichards/edu_effect-okf/blob/main/apps/cli/src/lib/graph-utils.ts) to be sure of the api as well as test out their implementation with some real examples.

- Slack [thread](https://discord.com/channels/795981131316985866/1518570933864894485)

### DIFF
--- a/.changeset/fair-jobs-like.md
+++ b/.changeset/fair-jobs-like.md
@@ -1,0 +1,5 @@
+---
+"effect": minor
+---
+
+add a `maxDepth` option to `Graph` search configuration, allowing `dfs` and `bfs` traversals to limit returned nodes by depth from the configured start nodes.
--- a/.changeset/hip-friends-kiss.md
+++ b/.changeset/hip-friends-kiss.md
@@ -1,0 +1,10 @@
+---
+"effect": minor
+---
+
+added graph set operations for combining and comparing graphs
+
+- `Graph.compose` - union of two graphs, merging nodes by identity
+- `Graph.intersection` - intersection of two graphs, keeping only common nodes and edges
+- `Graph.difference` - difference of two graphs, removing edges present in the second graph
+- `Graph.symmetricDifference` - symmetric difference of two graphs, keeping edges present in exactly one graph
--- a/.changeset/six-pumas-take.md
+++ b/.changeset/six-pumas-take.md
@@ -1,0 +1,9 @@
+---
+"effect": minor
+---
+
+add advanced graph set operations for deriving related graph structures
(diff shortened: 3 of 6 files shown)


IMax153:
Hi @lloydrichards,

PR #2489, opened 2026-06-25
Add a DuckDB adapter for Effect SQL

arthberman:
## What

Adds a new `@effect/sql-duckdb` package — an Effect SQL adapter for [DuckDB](https://duckdb.org), backed by [`@duckdb/node-api`](https://www.npmjs.com/package/@duckdb/node-api) — and registers the `duckdb` dialect in the core SQL statement compiler.

It follows the existing `packages/sql/*` adapter conventions (compare `sqlite-node` / `libsql` / `pg`): a dialect-specific `DuckdbClient` service plus the generic `SqlClient`, a `DuckdbMigrator`, and `layer` / `layerConfig` / `layerFrom` constructors.

## Features

- Three constructors: `make` (managed instance), `fromInstance`, and `fromConnection`, with optional process-local instance caching (`cache: true`) and read-only mode.
- Configurable result conversion: `js` (default), `native`, or `json`.
- DuckDB-specific helpers exposed on `DuckdbClient`: typed parameters (`sql.typed`), appenders, prepared statements, scalar-function registration, query `interrupt`/`progress`, and `getTableNames`.
- Error classification mapping DuckDB messages to the shared `SqlError` reasons (unique/constraint/syntax/serialization/lock/timeout).

## Design notes

- **Transactions**: DuckDB has no SQL savepoints (and rejects nested `BEGIN`), so nested transactions are modeled by marking the whole outer transaction for rollback when a nested scope fails.
- **Streaming**: outside a transaction, `.stream` runs on a **dedicated connection** that is closed when the stream finishes or is interrupted. A DuckDB streaming result is bound to its connection and is silently truncated if another statement runs on that connection before it is fully consumed; using a dedicated connection isolates it from concurrent queries (including per-row queries during consumption). Inside a transaction — or when no instance is available to open a connection — the result is materialized up front to preserve transactional visibility.

## Core change

`Statement.ts` gains `"duckdb"` in the `Dialect` union and as a key on `onDialect` / `onDialectOrElse`; the existing `pg` `onDialect` test is updated accordingly.

## Testing

`pnpm --filter @effect/sql-duckdb test` (in-memory DuckDB, no container required) covers basic queries, compiler helpers, typed/binary params, streaming (including isolation from concurrent connection use), JSON result mode, name transforms, appender/scalar-function/table-names, prepared statements, error classification, the constructor variants, transactions (commit/rollback/nested), and migrations.

A changeset is included.

### DIFF
--- a/.changeset/duckdb-sql-adapter.md
+++ b/.changeset/duckdb-sql-adapter.md
@@ -1,0 +1,6 @@
+---
+"effect": patch
+"@effect/sql-duckdb": patch
+---
+
+Add a DuckDB adapter for Effect SQL, backed by `@duckdb/node-api`.
--- a/ai-docs/package.json
+++ b/ai-docs/package.json
@@ -16,6 +16,7 @@
     "@effect/platform-node": "workspace:*",
     "@effect/sql-clickhouse": "workspace:*",
     "@effect/sql-d1": "workspace:*",
+    "@effect/sql-duckdb": "workspace:*",
     "@effect/sql-libsql": "workspace:*",
     "@effect/sql-mssql": "workspace:*",
     "@effect/sql-mysql2": "workspace:*",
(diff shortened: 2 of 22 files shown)


IMax153:
Hi @arthberman,

PR #2495, opened 2026-06-28
Preserve SSE data schema identifier in OpenAPI output

### LINKED ISSUE #2496: HttpApiSchema.StreamSse({ data }) steals the data schema's OpenAPI identifier
## Summary

When generating an OpenAPI spec from an `HttpApi` that uses `HttpApiSchema.StreamSse({ data })`, the JSON-string transport wrapper claims the data schema's `identifier`, pushing the actual decoded schema to a suffixed name.

## Reproduction

```ts
import { Schema } from "effect"
import { HttpApi, HttpApiEndpoint, HttpApiGroup, HttpApiSchema, OpenApi } from "effect/unstable/httpapi"

const Event = Schema.Struct({
  kind: Schema.String,
  payload: Schema.String
}).annotate({ identifier: "MyEvent" })

const Api = HttpApi.make("Api").add(
  HttpApiGroup.make("test").add(
    HttpApiEndpoint.get("stream", "/stream", {
      success: [HttpApiSchema.StreamSse({ data: Event })]
    })
  )
)

const spec = OpenApi.fromApi(Api)
console.log(Object.keys(spec.components?.schemas ?? {}))
// [ 'MyEvent1', 'MyEvent' ]
```

## Actual

- `MyEvent` resolves to the JSON-string wrapper (`{ type: "string", contentSchema: { $ref: "#/components/schemas/MyEvent1" } }`)
- `MyEvent1` resolves to the real decoded object schema

The `data` field is encoded as a JSON string via `Schema.fromJsonString(options.data)`, and that wrapper shares the inner schema's identifier. Because the wrapper is emitted first, it takes the bare name and the decoded schema is deduplicated to `MyEvent1`.

## Expected

The decoded data schema should keep its identifier (`MyEvent`), and the JSON-string wrapper should get a distinct name.

## Impact

OpenAPI codegen consumers generate inverted types: e.g. `type MyEvent = string` with the real union/object renamed to `MyEvent1`. This was hit in the wild by opencode (their SDK ended up with `V2Event = string`), forcing a local `patches/` override of Effect.

## Fix

PR #2495 annotates the wrapper with its own `${identifier}Stream` name when the data schema has an identifier, leaving the data schema's identifier intact.

kitlangton:
## What

`HttpApiSchema.StreamSse({ data })` encodes the SSE `data` field as a JSON string (`Schema.fromJsonString(data)`). In the generated OpenAPI spec, that string wrapper was claiming the data schema's `identifier`, pushing the decoded schema to a suffixed name.

For a data schema with `identifier: "MyEvent"`:

```
Before:  MyEvent  = { type: "string", ... }      ← the JSON wrapper (wrong)
         MyEvent1 = { type: "object", ... }      ← the actual decoded schema

After:   MyEvent       = { type: "object", ... } ← decoded schema keeps its name
         MyEventStream = { type: "string", contentSchema: { $ref: "#/components/schemas/MyEvent" } }
```

This inverts the names for OpenAPI codegen consumers: generated clients end up with `MyEvent = string` and the real type renamed to `MyEvent1`, breaking typed access to event payloads.

## Fix

`StreamSse` now annotates the JSON-string transport wrapper with its own `${identifier}Stream` name (when the data schema has an identifier), leaving the data schema's identifier intact. This matches how identifiers are resolved/assigned elsewhere in the OpenAPI generator (`SchemaAST.resolveIdentifier`).

The fix is scoped to the `StreamSse` constructor. Other `fromJsonString` usages (`encoding/Sse`, `HttpApiClient`, `HttpApiBuilder`) are runtime decode/encode codecs that are never rendered into an OpenAPI document, so they don't share the collision.

## Validation

- `pnpm lint-fix`
- `pnpm test packages/effect/test/unstable/httpapi/OpenApi.test.ts` (added a regression test)
- `pnpm check:tsgo`

A changeset is included.

Closes #2496


packages/effect/src/unstable/httpapi/HttpApiSchema.ts
```diff
@@ -422,6 +422,14 @@ export const StreamSse: {
   })
 }
 
+const sseDataJsonSchema = (data: Schema.Constraint) => {
+  const json = Schema.fromJsonString(data)
```
tim-smart, line 426: Potentially the real issue is with `Schema.fromJsonString` transforming the schema but keeping the identifier.

I'll wait for @gcanti to weight in here.

gcanti: > Potentially the real issue is with `Schema.fromJsonString` transforming the schema but keeping the identifier.

Agreed, we could add a synthetic identifier to the encoded side: https://github.com/Effect-TS/effect-smol/pull/2512

PR #2500, opened 2026-06-29
fix(PersistedQueue): Fix Redis-backed PersistedQueue script handling

hsyntax:
## Summary

- Fix Redis-backed `PersistedQueue` reset handling so `HGETALL` results are read as field/value pairs from Redis' flat array response.
- Fix the failed-item Redis script key count so exhausted queue items can be moved to the failed queue instead of failing with a missing `KEYS[3]`.
- Add NodeRedis/Testcontainers regression coverage for locked pending reset behavior and exhausted failed-item handling.

## Why

`DurableQueue` workers rely on `PersistedQueue.layerStoreRedis` for durable handoff. Two Redis script bugs could break that path:

- `resetQueueRedis` treated `HGETALL` as if Lua yielded key/value pairs directly, causing locked pending items to be requeued incorrectly as both the item id and payload.
- `failedRedis` declared only two keys even though the script reads `KEYS[3]`, so exhausted items errored before reaching the failed list.

## Validation

- `pnpm lint-fix`
- `pnpm test packages/platform-node/test/NodeRedis.test.ts packages/effect/test/unstable/workflow/DurableQueue.test.ts`
- `pnpm check:tsgo`

_This PR was raised with the help of Codex ( GPT-5.5-xhigh )_

### DIFF
--- a/packages/effect/src/unstable/persistence/PersistedQueue.ts
+++ b/packages/effect/src/unstable/persistence/PersistedQueue.ts
@@ -613,7 +613,9 @@ local key_pending = KEYS[2]
 local prefix = ARGV[1]
 
 local entries = redis.call("HGETALL", key_pending)
-for id, payload in pairs(entries) do
+for i = 1, #entries, 2 do
+  local id = entries[i]
+  local payload = entries[i + 1]
   local lock_key = prefix .. id .. ":lock"
   local exists = redis.call("EXISTS", lock_key)
   if exists == 0 then
@@ -673,7 +675,7 @@ redis.call("DEL", key_lock)
 redis.call("HDEL", key_pending, id)
 redis.call("RPUSH", key_failed, payload)
 `,
-    numberOfKeys: 2
+    numberOfKeys: 3
   }
 )
 
--- a/packages/platform-node/test/NodeRedis.test.ts
+++ b/packages/platform-node/test/NodeRedis.test.ts
@@ -1,6 +1,7 @@
 import { NodeRedis } from "@effect/platform-node"
+import { assert, it } from "@effect/vitest"
 import { RedisContainer } from "@testcontainers/redis"
-import { Effect, Layer } from "effect"
+import { Effect, Layer, Schema } from "effect"
 import * as PersistedCacheTest from "effect-test/unstable/persistence/PersistedCacheTest"
 import * as PersistedQueueTest from "effect-test/unstable/persistence/PersistedQueueTest"
 import { PersistedQueue, Persistence } from "effect/unstable/persistence"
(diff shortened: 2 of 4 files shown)


packages/effect/src/unstable/persistence/PersistedQueue.ts
```diff
@@ -673,7 +675,7 @@ redis.call("DEL", key_lock)
 redis.call("HDEL", key_pending, id)
 redis.call("RPUSH", key_failed, payload)
 `,
-    numberOfKeys: 2
```
tim-smart, line 676: Only 2 keys are being passed

packages/platform-node/test/NodeRedis.test.ts
```diff
@@ -29,3 +31,97 @@ PersistedQueueTest.suite(
   "NodeRedis",
   PersistedQueue.layerStoreRedis().pipe(Layer.provide(RedisLayer))
 )
+
+const PersistedQueueRedisLayer = Layer.mergeAll(
```
tim-smart, line 39: There is a seperate test suite for PersistedQueue. Ideally all queue implementations would get reset coverage.

tim-smart (changes requested):

packages/effect/src/unstable/persistence/PersistedQueue.ts
hsyntax: Hey tim, 

Three keys are definitely being passed : 

Image

packages/platform-node/test/NodeRedis.test.ts
hsyntax: Done. I've kept one test here to validate Redis specific behaviour, but the new tests in `PersistedQueueTest.ts` verify both the `numberOfKeys` fix and the script fix

hsyntax:
Hey @tim-smart. Apologies for the delay in getting back to this PR. It's ready for another round of review. 

If approved, would you like me to back-port this patch back to Effect v3?

hsyntax:
There are some additional issues with the Redis implementation of `PersistedQueue` which needs to be fixed. I'm not raising them here to avoid spamming PR's, but happy to raise them once this is merged in and you're happy to accept more PR's on this topic.

For example, Stale lease-holder writes problem : https://github.com/hsubra89/effect-smol/pull/1

packages/effect/src/unstable/persistence/PersistedQueue.ts
tim-smart: Ah yep, I think the github UI was showing the wrong script header and got me confused.

tim-smart:
> Hey @tim-smart. Apologies for the delay in getting back to this PR. It's ready for another round of review.
> 
> If approved, would you like me to back-port this patch back to Effect v3?

Yeah go ahead. If you have more fixes I'm happy to review them when I have a moment.

tim-smart (approved):

PR #2503, opened 2026-06-29
fix: Made field-less `Schema.Struct({})` respect `onExcessProperty: 'ignore' | 'preserve' | 'error'`

nikelborm:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Motivation: you remove one field in the schema; the decoded object gets fewer fields. You remove another; fewer again. You remove the last; suddenly, all the fields come back, regardless of the default `onExcessProperty: "ignore"`. If the user wants to retrieve all fields, they can set `onExcessProperty: "preserve"`.

I tried to defend old behavior with "`{}` means `typeof obj === 'object' && obj !== null` and effect's schema tries to be closer to TypeScript's behaviour, by treating `Schema.Struct({})` as something passing the condition". And this might hold in isolation. But considering the context, by this logic, any fields which are not explicitly mentioned in `Schema.Struct({ ... })` should be preserved in all decoded objects, not only the ones that have `{}` signature, to follow TypeScript's assignability rules. And this is not what happens. The behavior should be consistent between field-less and field-ful structs.

## Related

- https://github.com/Effect-TS/effect/pull/6298
- https://github.com/Effect-TS/effect/pull/6434

### DIFF
--- a/.changeset/happy-facts-jam.md
+++ b/.changeset/happy-facts-jam.md
@@ -1,0 +1,75 @@
+---
+"effect": patch
+---
+
+Made field-less `Schema.Struct({})` respect `onExcessProperty: 'ignore' | 'preserve' | 'error'`
+
+Motivation: you remove one field, the resulting object gets fewer fields. You
+remove another, fewer again. You remove the last; suddenly all the fields come
+back, without respect to the default `onExcessProperty: "ignore"`. If the user
+wanted to get all the fields, they have an option to set
+`onExcessProperty: "preserve"`.
+
+I tried attempting to defend old behavior with "`{}` means
+`typeof obj === 'object' && obj !== null` and effect's schema tries to be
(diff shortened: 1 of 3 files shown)


IMax153:
Hi @nikelborm,

nikelborm:
I had a mirror PR opened against v3 https://github.com/Effect-TS/effect/pull/6298

Should I close it and create a new one, or should I overwrite the changes in that one?

IMax153:
@nikelborm - the Effect v3 codebase still lives on the `v3` branch - all PRs (including yours) that were opened prior to the v4 codebase merger have been re-targeted to the `v3` branch.

So I would suggest opening a companion PR targeting `main` (i.e. the v4 codebase).

nikelborm:
Reopened https://github.com/Effect-TS/effect/pull/6434

PR #2504, opened 2026-06-29
Add `Reloadable` module

fubhy:


packages/effect/src/Reloadable.ts
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * Use when you need to access or provide the `Reloadable` wrapper itself rather
+ * than the current service implementation.
+ *
+ * @category context
+ * @since 4.0.0
+ */
+export const tag = <I, S>(
+  service: Context.Key<I, S>
+): Context.Key<Reloadable<I>, Reloadable<S>> =>
```
tim-smart: I think this should return `Context.Service`

packages/effect/src/Reloadable.ts
```diff
@@ -0,0 +1,241 @@
...
+ * **When to use**
+ *
+ * Use when a caller should trigger a best-effort refresh without waiting for the
+ * rebuild to complete.
+ *
+ * @see {@link reload} for waiting until the reload completes or fails
+ * @category resource management
+ * @since 4.0.0
+ */
+export const reloadFork = <I, S>(
```
tim-smart: I think we should remove this api. Prefer composition with the Effect.fork apis.

tim-smart (changes requested):

tim-smart:
The more I look at Reloadable the weirder it gets. I feel like there is a better abstraction. Even the `Resource` module is probably better in most cases.

fubhy:
> The more I look at Reloadable the weirder it gets. I feel like there is a better abstraction. Even the `Resource` module is probably better in most cases.

Sonds good. Want to take this over and think of a better alternative? Feel free to close this.

PR #2505, opened 2026-06-29
accept UrlParams.Input in some UrlParams apis

KhraksMamtsov:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Updated the Input type to include the target type, following the approach used in the `Duration` module, to improve usability.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/packages/effect/src/unstable/http/Headers.ts
+++ b/packages/effect/src/unstable/http/Headers.ts
@@ -163,7 +163,6 @@ export const HeadersSchema: HeadersSchema = Schema.declare(
  * @since 4.0.0
  */
 export type Input =
-  | Headers
   | Record.ReadonlyRecord<string, string | ReadonlyArray<string> | undefined>
   | Iterable<readonly [string, string]>
 
@@ -186,9 +185,6 @@ export const empty: Headers = Object.create(Proto)
  * @since 4.0.0
  */
 export const fromInput: (input?: Input) => Headers = (input) => {
-  if (isHeaders(input)) {
-    return input
-  }
   if (input === undefined) {
     return empty
   } else if (Symbol.iterator in input) {
--- a/packages/effect/src/unstable/http/Headers.ts
+++ b/packages/effect/src/unstable/http/Headers.ts
@@ -163,6 +163,7 @@ export const HeadersSchema: HeadersSchema = Schema.declare(
  * @since 4.0.0
  */
 export type Input =
+  | Headers
   | Record.ReadonlyRecord<string, string | ReadonlyArray<string> | undefined>
   | Iterable<readonly [string, string]>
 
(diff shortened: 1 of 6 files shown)


packages/effect/src/unstable/http/Url.ts
```diff
@@ -321,9 +321,9 @@ export const urlParams = (url: URL): UrlParams.UrlParams => UrlParams.fromInput(
  * @since 4.0.0
  */
 export const modifyUrlParams: {
-  (f: (urlParams: UrlParams.UrlParams) => UrlParams.UrlParams): (url: URL) => URL
-  (url: URL, f: (urlParams: UrlParams.UrlParams) => UrlParams.UrlParams): URL
-} = dual(2, (url: URL, f: (urlParams: UrlParams.UrlParams) => UrlParams.UrlParams) =>
+  (f: (urlParams: UrlParams.Input) => UrlParams.UrlParams): (url: URL) => URL
```
tim-smart: Input in the contravariant position doesn't seem correct

packages/effect/src/unstable/http/Headers.ts
```diff
@@ -163,6 +163,7 @@ export const HeadersSchema: HeadersSchema = Schema.declare(
  * @since 4.0.0
  */
 export type Input =
```
tim-smart, line 165: Were you planning on changing any of the Headers apis too?

tim-smart (changes requested):

packages/effect/src/unstable/http/Url.ts
KhraksMamtsov: rolled back

packages/effect/src/unstable/http/Headers.ts
KhraksMamtsov: I just caught it in the crossfire.
I'm reverting it.

KhraksMamtsov:
@tim-smart 
I'd also suggest either removing this function or moving it from `UrlParams` to` Url`.

The behavior of baseUrl() isn't very intuitive.
```ts

const baseUrl = (): string | undefined => {
  if (
    "location" in globalThis &&
    globalThis.location !== undefined &&
    globalThis.location.origin !== undefined &&
    globalThis.location.pathname !== undefined
  ) {
    return location.origin + location.pathname
  }
  return undefined
}

/**
 * Creates a `URL` safely by appending `UrlParams` and an optional hash to a URL string.
 *
 * **Details**
 *
 * Returns a `Result` that fails with `UrlParamsError` if the URL cannot be
 * constructed.
 *
 * @category converting
 * @since 4.0.0
 */
export const makeUrl = (
  url: string,
  params: UrlParams,
  hash: string | undefined
): Result.Result<URL, UrlParamsError> => {
  try {
    const urlInstance = new URL(url, baseUrl())
    for (let i = 0; i < params.params.length; i++) {
      const [key, value] = params.params[i]
      if (value !== undefined) {
        urlInstance.searchParams.append(key, value)
      }
    }
    if (hash !== undefined) {
      urlInstance.hash = hash
    }
    return Result.succeed(urlInstance)
  } catch (e) {
    return Result.fail(new UrlParamsError({ cause: e }))
  }
}
```

tim-smart (approved):

PR #2509, opened 2026-06-30
feat(sql-tursodatabase): add @effect/sql-tursodatabase driver for the Rust Turso engine

artile:
## What

Adds **`@effect/sql-tursodatabase`**, an `@effect/sql` driver for the in-process **Rust Turso engine** (`@tursodatabase/database`).

It mirrors the existing `@effect/sql-libsql` and `@effect/sql-sqlite-node` drivers:

- `TursoClient` — a `SqlClient` (queries as `Effect`, the connection as a composable `Layer` via `TursoClient.layer` / `layerConfig`, typed `SqlError` via Effect SQL's SQLite error classification).
- `TursoMigrator` — `run` / `layer` over the shared `Migrator`.
- Transaction support with savepoints, `SafeIntegers`, and OpenTelemetry span attributes, consistent with the other SQLite-family drivers.
- Streaming queries are not implemented (same as `sql-libsql` / `sql-sqlite-node`).

## Why

`@tursodatabase/database` is the in-process Rust SQLite-compatible engine (MVCC / `BEGIN CONCURRENT`, CDC). It is the only SQLite-family engine in the ecosystem without a first-party Effect SQL client. This package closes that gap so Effect users can adopt the Rust Turso engine without hand-rolling the `Effect.tryPromise` / tagged-error / `acquireRelease` seam.

It is also the upstream half of drizzle-team/drizzle-orm#5962 (which adds `drizzle-orm/effect-tursodatabase` on top of this client).

## Notes

- The engine's async API types `prepare` as returning a `Promise`, but some published versions return the statement synchronously; the client normalizes both shapes.
- The engine opens databases by filesystem path or `":memory:"` (it does not accept `file:` / `libsql:` / `http:` URLs, unlike libSQL) — documented on `TursoClientConfig.Full.url`.

## Test

`packages/sql/tursodatabase/test/Client.test.ts` runs against a real in-memory Turso engine (queries, raw results, `withTransaction` commit/rollback, nested savepoints) — 6/6 passing. `tsc -b`, `oxlint`, and `dprint` are clean. Changeset included.

### DIFF
--- a/.changeset/add-sql-tursodatabase.md
+++ b/.changeset/add-sql-tursodatabase.md
@@ -1,0 +1,7 @@
+---
+"@effect/sql-tursodatabase": minor
+---
+
+Add `@effect/sql-tursodatabase`, an `@effect/sql` driver for the in-process Rust Turso engine (`@tursodatabase/database`).
+
+Provides `TursoClient` (a `SqlClient` exposing queries as `Effect`, the connection as a composable `Layer`, and typed `SqlError`s via Effect SQL's SQLite error classification) alongside `TursoMigrator`. Mirrors the existing `@effect/sql-libsql` and `@effect/sql-sqlite-node` drivers, including transaction support with savepoints and `SafeIntegers`. Streaming queries are not implemented.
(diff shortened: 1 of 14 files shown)


IMax153:
Hi @artile,

artile:
Migrated to https://github.com/Effect-TS/effect/pull/6522 against Effect-TS/effect:main, as requested. Please close this source PR — the discussion/review history is linked from the new one.

PR #2510, opened 2026-06-30
Machine: sound plan requirements with flat/nested split

patroza:
Builds on top of #2351 (the `Machine` module) and #1945. The diff below includes both #2351's full implementation **and** this PR's additions — only the last commit (`feat(machine): sound plan requirements, flat/nested split`) is new. Reviewers should focus on that commit, diffed against #2351's `sandro/state-charts` branch.

## Problem

PR #2351's `plan` returns `Effect<MacrostepPlan<...>, E | ..., R>` where `R` is the machine's full `R` — the union of **every** handler's service requirements (all event handlers, all entry/exit/always/onDone handlers, all invoke sources). This is sound (no requirement is missed) but imprecise: planning a single state/event transition forces the caller to provide services for handlers that will never run.

```ts
// Every handler's R is in the plan's R, even handlers on unrelated states
const planned = Machine.plan(machine, snapshot, new SignIn({}))
// Effect.Services<typeof planned> = ExitReq | EntryReq | AlwaysReq | UnrelatedReq | ...
// even if the SignIn transition only touches signedOut → signedIn
```

This makes it impossible to provide only the `R` that a specific state/event transition requires.

## Solution

Three layers of type-level narrowing that scope `R` and `E` to the concrete state/event pair, plus a flat/nested constructor split:

### 1. Event handler: leaf-to-root walk

`SnapshotLeafPath` descends compound/parallel snapshot structure to the active leaf. `ChainToRoot` produces the leaf-to-root path chain. This matches the runtime's `getLeafCandidatePaths` (leaf → root walk in `selectEventTransitions`), so a handler nested under `app.auth.signedOut` is correctly resolved — instead of only seeing the root path.

### 2. Lifecycle: source + target chain narrowing

`PlanLifecyclePaths` computes the source chain (exit handlers) unioned with the target chain (entry/always handlers). The target is extracted from `PlanResult` — the handler's return value, which carries a `Target` with the destination path. `LifecycleServicesForChain` scopes entry/exit/always/onDone to these specific paths rather than the union of every handler in the machine.

### 3. Flat/nested constructor split

| | `Machine.make` | `Machine.makeNested` |
|---|---|---|
| State trees | Atomic only (rejects compound/parallel via `ValidateFlat`) | Full `StateSchemas` |
| Plan R | Sound path-narrowed (source + target chains only) | Path-narrowed ∪ coarse `LifecycleR` (all lifecycle) |
| Settle loop | Cannot cascade (no compound subtrees) | Can cascade → `LifecycleR` over-approximation is safe |

A **single** `plan` function auto-selects behavior via `PlanServicesFor` / `PlanErrorsFor` conditional types, keyed on `NestedStateCheck<States[keyof States]>`. There is no `planNested` — the `States` type parameter drives the selection. It is impossible to call the wrong plan.

`NestedStateCheck` distinguishes compound/parallel configs (`{ schema, states }` together) from bare `TaggedSchema` values, which also carry a `states` property from Effect schema internals.

### `planSafe`

Like `plan` but constrains `State` to snapshots with a registered handler for `Event` (via `SnapshotFor`), eliminating `UnhandledEventError` from the error channel.

## Known limitation: raised events

`Runtime.raise(event)` enqueues an event as a side effect and returns `Effect<void, MachineSchemaDecodeError>` — `R = never`. The raised event's own event-handler requirements are not part of the original handler's return type, so `PlanServices` cannot see them. The lifecycle side of the raised transition (entry/exit/always) is covered by `LifecycleR` in nested mode, but the raised event's handler services are invisible. Fixing this requires an API change to encode raised events in the handler's return type, not just additional type-level computation. This limitation is inherited from the upstream module (PR #2351).

## Tests

Four new typetest cases in `Machine.tst.ts`:

1. **Nested event handler**: `Effect.Services<typeof planned>` is exactly `NestedRequirement` from a handler on `app.auth.signedOut` (not the union of all handlers).
2. **Flat lifecycle narrowing**: exit/entry/always from source+target chains are included; `UnrelatedReq` from an unrelated sibling state is **excluded**.
3. **Nested coarse fallback**: same setup as (2) but via `makeNested` — `UnrelatedReq` **is included** (proves the coarse fallback works).
4. **Raised events gap**: documents that `RaisedReq` from a raised event's handler is **not surfaced** (known limitation).

Existing runtime tests (125) and typetests (819) pass. The pre-existing `missingEffectContext` / `missingEffectError` errors in `machineProcess.ts` / `machineRuntime.ts` are from PR #2351's base and are not introduced by this PR.


IMax153:
Hi @patroza,

PR #2514, opened 2026-07-01
feat(openapi-generator): generate binary responses as typed streams with typed errors

patroza:
## What

The OpenAPI generator emitted binary success responses (e.g. `application/pdf`, `type: string` + `format: binary`) as JSON-string clients, so a PDF/download endpoint decoded its body as a JSON string and lost its typed error responses.

This teaches the generator to treat binary success responses as streams:

- **Detection by schema shape**, not just media type: `type: string` with `format: binary` (or `contentEncoding: binary`), and binary media types including `application/pdf` — not only `application/octet-stream`.
- **Typed error channel on the stream method.** An operation whose only success response is binary now generates a single `${id}Stream` method returning `Stream.Stream<Uint8Array, HttpClientError | SchemaError | <typed OpenAPI error responses>>`. `binaryRequest` matches on status (`HttpClientResponse.matchStatus`) so per-status error bodies decode into their typed variants.
- **No redundant JSON method** is emitted for a binary-only operation (previously both a JSON method and the stream method were produced).
- Tracked on `ParsedOperation` via a new `binaryResponseStatusCodes` set; `binaryResponse` semantics generalized from "octet-stream only" to "binary success response".

Applies to both the schema-backed HttpClient transformer and the type-only transformer.

## Test

Added a generator test: a `GET` returning `application/pdf` (200, `format: binary`) with a `400` JSON error produces a typed `getInvoiceStream` (error channel includes the `400` variant), emits `binaryRequest(HttpClientResponse.matchStatus({ "2xx": ..., "400": decodeError(...) }))`, and does **not** emit a JSON `getInvoice` method / `schemaBodyJson`.

All 69 `@effect/openapi-generator` tests pass; `tspc -b` and `dprint check` clean.

### DIFF
--- a/.changeset/openapi-generator-binary-stream-typed-errors.md
+++ b/.changeset/openapi-generator-binary-stream-typed-errors.md
@@ -1,0 +1,7 @@
+---
+"@effect/openapi-generator": patch
+---
+
+Generate binary (e.g. `application/pdf`, `format: binary`) success responses as typed `*Stream` clients with typed OpenAPI error responses, instead of decoding them as JSON strings.
+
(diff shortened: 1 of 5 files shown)


IMax153:
Hi @patroza,

PR #2518, opened 2026-07-01
EFF-897 Move UrlParams.makeUrl to Url.make

tim-smart:
## Summary
- Move safe URL construction from `UrlParams.makeUrl` to `Url.make`.
- Add `Url.UrlError` for URL construction failures and remove the now-unused `UrlParams.UrlParamsError`.
- Update HTTP client request URL construction call sites and tests to use `Url.make` / `Url.UrlError`, plus a changeset.

## Validation
- `pnpm lint-fix`
- `pnpm test packages/effect/test/unstable/http/Url.test.ts`
- `pnpm check:tsgo`

### DIFF
--- a/packages/effect/src/unstable/http/Url.ts
+++ b/packages/effect/src/unstable/http/Url.ts
@@ -9,18 +9,28 @@
  * @since 4.0.0
  */
 import * as Cause from "../../Cause.ts"
+import * as Data from "../../Data.ts"
 import { dual } from "../../Function.ts"
 import * as Redacted from "../../Redacted.ts"
 import * as Result from "../../Result.ts"
 import * as UrlParams from "./UrlParams.ts"
 
 /**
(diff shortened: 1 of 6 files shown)


packages/effect/src/unstable/http/Url.ts
```diff
@@ -14,6 +14,51 @@ import * as Redacted from "../../Redacted.ts"
...
+ * constructed.
+ *
+ * @category constructors
+ * @since 4.0.0
+ */
+export const make = (
+  url: string,
+  params: UrlParams.UrlParams,
+  hash: string | undefined
+): Result.Result<URL, UrlParams.UrlParamsError> =>
```
tim-smart: A `UrlError` should be added instead of using UrlParamsError.

If `UrlParamsError` isn't used anywhere else it can be removed.

PR #2530, opened 2026-07-03
feat(@effect/ai-anthropic): support custom auth header and extra headers in AnthropicClient

polRk:
## Summary
- Add `apiKeyHeader` option to `AnthropicClient.make`/`layer`/`layerConfig` — customize the header name used to send the API key (default `"x-api-key"`)
- Add `apiKeyScheme` option — prepend an auth scheme prefix (e.g. `Bearer`, `OAuth`) to the API key value
- Add `headers` option — merge arbitrary extra HTTP headers into every request

Useful when routing Anthropic requests through a gateway/proxy that expects different auth conventions (e.g. `authorization: Bearer <token>` instead of `x-api-key: <token>`).

This is a port of [Effect-TS/effect#6300](https://github.com/Effect-TS/effect/pull/6300), opened against the v3 `effect` repo, since v3 is currently in feature freeze per maintainer guidance in effect#6224.

## Test plan
- [x] `pnpm lint-fix`
- [x] `tsc -b` (package typecheck)
- [x] `pnpm docgen`
- [x] Manual verification via scratchpad script with a mocked `HttpClient` confirming `authorization: Bearer <key>` and custom header appear in outgoing requests, and that the dynamic header name is correctly redacted in logs

### DIFF
--- a/.changeset/anthropic-client-custom-auth-header.md
+++ b/.changeset/anthropic-client-custom-auth-header.md
@@ -1,0 +1,5 @@
+---
+"@effect/ai-anthropic": minor
+---
+
+Add `apiKeyHeader`, `apiKeyScheme`, and `headers` options to `AnthropicClient.make`/`layer`/`layerConfig`, allowing the API key header name (e.g. `authorization`), its auth scheme prefix (e.g. `Bearer`/`OAuth`), and arbitrary extra request headers to be customized. Useful when targeting a gateway or proxy in front of Anthropic's API that expects different authentication conventions.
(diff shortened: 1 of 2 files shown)


tim-smart:
You can use `transformClient` for all of these things, not sure it is worth the extra options.

IMax153:
@polRk I agree with @tim-smart - I want to try to avoid a combinatorial explosion of new configuration options that will only make things more confusing - for this case I would suggest using the `transformClient` option to configure the underlying `HttpClient` used by the SDK. 

If that's not sufficient for your use case, please let us know.

polRk:
@tim-smart @IMax153 I know about this feature, but it's not very user-friendly (my opinion), and many clients/sdks use this approach, which is why I suggested this PR. If it's rejected, it would be better to describe how to achieve this behavior using the standard transformClient in the jsdoc.

IMax153:
I do not think this PR will be accepted as is to avoid a potential combinatorial explosion of configuration options.

If you're willing to author documentation describing how to modify the underlying client via `transformClient`, that would be much appreciated. Otherwise, we will get to it at some point.

PR #2545, opened 2026-07-07
Isolate cluster entity handler context from the fiber that first wakes it

marbemac:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

A cluster entity's `RpcServer` is built once, lazily, inline on whichever fiber first delivers a message to it — and it captured that fiber's ambient context via `Effect.provideContext` (a merge, not a replace). So a caller/request-scoped service on that fiber — e.g. an in-flight SQL `TransactionContext` — got frozen into the long-lived server (`toHandlers` bakes it into each handler's `entry.context`, and the server captures it again) and reused by every later request, long after the caller's scope had closed. It triggers on a non-persisted first message (the default) or a defect rebuild, and once frozen it poisons every later request, durable ones included.

**Fix:** build the handlers and the server under exactly the registration context plus the per-entity additions, using `updateContext` (set) instead of `provideContext` (merge). Same change in `Entity.makeTestClient`. Sitting in the `writeRef` acquire body, it also covers defect rebuilds. (Alternative: isolate in `ResourceMap`'s lookup wrapper to fix all consumers at once — I didn't, as it's broader and misses the defect-rebuild path, which bypasses the wrapper.)

**Tests:** `ContextBleedEntity` (volatile + persisted RPCs) reads a marker service standing in for the transaction context. The first request wakes the entity providing `CallerId = "A"`; later requests provide nothing and must not see `"A"` — covered through real `Sharding` (volatile and durable) and `makeTestClient`. All fail on `main`, pass after the fix.

### DIFF
--- a/.changeset/fix-cluster-entity-context-bleed.md
+++ b/.changeset/fix-cluster-entity-context-bleed.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Isolate a cluster entity's handler context from the fiber that first wakes it, so a caller/request-scoped service (such as an in-flight SQL transaction) present on that fiber is no longer frozen into the entity's long-lived `RpcServer` and reused by every later request.
--- a/.changeset/fix-cluster-entity-context-bleed.md
+++ b/.changeset/fix-cluster-entity-context-bleed.md
@@ -2,4 +2,4 @@
 "effect": patch
 ---
 
-Isolate a cluster entity's handler context from the fiber that first wakes it, so a caller/request-scoped service (such as an in-flight SQL transaction) present on that fiber is no longer frozen into the entity's long-lived `RpcServer` and reused by every later request.
+Use registration context for cluster entities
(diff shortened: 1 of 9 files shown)


tim-smart (commented):
The problem is that it will also strip away any custom default services, like tracers loggers etc. But I guess the registration context should have those

.changeset/fix-cluster-entity-context-bleed.md
```diff
@@ -0,0 +1,5 @@
+---
+"effect": patch
+---
+
+Isolate a cluster entity's handler context from the fiber that first wakes it, so a caller/request-scoped service (such as an in-flight SQL transaction) present on that fiber is no longer frozen into the entity's long-lived `RpcServer` and reused by every later request.
```
tim-smart:
```suggestion
Use registration context for cluster entities
```

tim-smart (approved):

PR #2551, opened 2026-07-08
Add Schedule.max api

tim-smart:
## Summary
- Remove the public `Schedule.both`, `Schedule.bothLeft`, `Schedule.bothRight`, and `Schedule.bothWith` APIs.
- Add `Schedule.max([...])`, which continues while all schedules recur and outputs the slowest selected duration.
- Rework `Schedule.max` to step all supplied schedules via `effect.all` instead of chaining pairwise intermediary schedules.
- Export schedule type extractors: `Schedule.Output`, `Schedule.Input`, `Schedule.Error`, and `Schedule.Env`.
- Update schedule docs/examples, tests, typetests, and add a changeset.

## Validation
- `pnpm lint-fix`
- `pnpm test packages/effect/test/Schedule.test.ts`
- `pnpm test-types Schedule.tst.ts`
- `pnpm check:tsgo`

### DIFF
--- a/packages/effect/src/Schedule.ts
+++ b/packages/effect/src/Schedule.ts
@@ -219,6 +219,38 @@ export declare namespace Schedule {
   }
 }
 
+/**
+ * Extracts the output type from a `Schedule`.
+ *
+ * @category type extractors
+ * @since 4.0.0
+ */
+export type Output<S> = S extends Schedule<infer Output, any, any, any> ? Output : never
+
+/**
+ * Extracts the input type from a `Schedule`.
+ *
+ * @category type extractors
+ * @since 4.0.0
+ */
+export type Input<S> = S extends Schedule<any, infer Input, any, any> ? Input : never
+
+/**
+ * Extracts the error type from a `Schedule`.
+ *
+ * @category type extractors
+ * @since 4.0.0
+ */
+export type Error<S> = S extends Schedule<any, any, infer Error, any> ? Error : never
+
+/**
+ * Extracts the service requirements from a `Schedule`.
+ *
+ * @category type extractors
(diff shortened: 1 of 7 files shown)


packages/effect/src/Schedule.ts
```diff
@@ -789,287 +790,108 @@ export const andThenResult: {
...
+type SchedulesError<Schedules extends ReadonlyArray<Schedule<any, any, any, any>>> = ScheduleError<Schedules[number]>
+
+type SchedulesEnv<Schedules extends ReadonlyArray<Schedule<any, any, any, any>>> = ScheduleEnv<Schedules[number]>
+
+const maxDuration = <Input, Error, Env, Input2, Error2, Env2>(
+  self: Schedule<Duration.Duration, Input, Error, Env>,
+  other: Schedule<Duration.Duration, Input2, Error2, Env2>
+): Schedule<Duration.Duration, Input & Input2, Error | Error2, Env | Env2> =>
+  fromStep(effect.map(
+    effect.zip(toStep(self), toStep(other)),
```
tim-smart: This could use effect.all with the array of schedules instead of creating all the intermediary schedules in a loop.

packages/effect/src/Schedule.ts
```diff
@@ -789,287 +790,108 @@ export const andThenResult: {
     }
   })))
 
+type ScheduleInput<S> = S extends Schedule<any, infer Input, any, any> ? Input : never
```
tim-smart: These extractors could be exported too

```ts
Schedule.Input
Schedule.Error
// etc
```

And move them underneath the `namespace Schedule`

PR #2555, opened 2026-07-08
fix(effect): preserve atom registry parent links

petar-omni:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Fixes an `AtomRegistry` dependency graph bug where an atom can stop being invalidated by a parent atom after recomputation.

## Problem

`AtomRegistry` tracks dependencies bidirectionally:

- a child atom stores its `parents`
- a parent atom stores its `children`

Those links must stay consistent. In some cases, the child still had a parent in `child.parents`, but the parent no longer had the child in `parent.children`. Once that happens, future updates to the parent do not invalidate the child, so the child keeps returning stale data.

## When It Happens

The bug can happen when a derived atom reads the same parent more than once during one evaluation, then later recomputes and reads that parent fewer times.

Example shape:

```ts
const derived = Atom.make((get) => {
  const value = get(base)

  if (get(flag)) {
    get(base)
  }

  return value
})
```

### DIFF
--- a/.changeset/clean-atom-parent-links.md
+++ b/.changeset/clean-atom-parent-links.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix AtomRegistry parent-child tracking when an atom reads the same parent more than once.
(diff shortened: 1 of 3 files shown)


tim-smart:
Will move to using sets for tracking dependencies #2562

PR #2564, opened 2026-07-08
Implement McpServer Middleware API

naji247:
Hi from the [AgentCat](https://github.com/agentcathq) team :) We're a fellow open source solution for user analytics on top of MCP servers. We had a request from one of our customers to support Effect's MCP server implementation, but in order to do so, Effect needs a clean middleware API for us to properly "plug in". 

Happy to make this better in anyway :)

## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR introduces robust, transforming Middleware support to `effect/unstable/ai/McpServer`.

The new Middleware API (`McpServer.middleware`) safely enables full interception, modification, short-circuiting, and observation of MCP requests and responses. Middlewares have full access to request metadata (including headers,`mcpSessionId`, and `clientInfo`) and smoothly integrate with Effect's dependency injection (`Context`) so that underlying `McpServer` properties (like registered tools) can be queried.

**Key Capabilities:**
    - Seamlessly wrap core handlers (`initialize`, `tools/list`, `tools/call`, `prompts`, `resources`, etc.) using `request.next()`.
    - Rewrite outgoing responses or incoming payloads.
    - Short-circuit operations with immediate success results or valid MCP JSON-RPC errors (like `InvalidParams`).
    - Time exact execution duration of sub-handlers using `Effect.timed`.

This feature establishes a stable foundation for advanced 3rd-party integrations to hook into Effect servers natively empowering use cases like custom telemetry/observability solutions (e.g., AgentCat), and dynamic injection.

Comprehensive test coverage (`packages/effect/test/unstable/ai/McpServer.test.ts`) has been added as well as a `.changeset` is included.

### DIFF
--- a/.changeset/mcp-middleware-api.md
+++ b/.changeset/mcp-middleware-api.md
@@ -1,0 +1,7 @@
+---
+"effect": minor
+---
+
+Add robust, transforming Middleware support for Effect MCP servers. 
+
+The new Middleware API (`McpServer.middleware`) allows intercepting, modifying, short-circuiting, and observing MCP requests and responses seamlessly. It provides full access to request metadata (headers, `mcpSessionId`, `clientInfo`) and seamlessly supports 3rd-party integration needs like advanced telemetry, and schema modification.
--- a/packages/effect/src/unstable/ai/McpServer.ts
+++ b/packages/effect/src/unstable/ai/McpServer.ts
@@ -42,6 +42,7 @@ import * as RpcServer from "../rpc/RpcServer.ts"
 import {
   CallToolResult,
   ClientNotificationRpcs,
+  ClientRequestRpcs,
   ClientRpcs,
   CompleteResult,
   Elicit,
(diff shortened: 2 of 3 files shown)


ebramanti:
@naji247 Good idea here, but the approach needs to be adjusted. Info below:

This introduces a second middleware system alongside the one the RPC layer already has. `McpServer` is built on `RpcServer`, which applies `RpcMiddleware` around every handler automatically — `ClientRpcs` already installs one (`McpServerClientMiddleware` in `McpSchema`), receiving nearly the same metadata `MiddlewareRequest` reconstructs by hand here.

The same capability could be exposed with much less machinery: define an `McpServerMiddleware` via `RpcMiddleware.Service`, attach it to `ClientRpcs` with `.middleware(...)`, and have `McpServer.run` resolve it with `Effect.serviceOption`, falling back to a no-op passthrough. Integrators then supply a single `Layer` with their implementation. This removes `runMiddlewares`, the hand-wrapped handler call sites, the mutable `middlewares` registry, and the ordering question (`Layer.mergeAll` builds concurrently, so registration order isn't guaranteed anyway).

Note `RpcMiddleware` is deliberately observe-only — for analytics that's sufficient (tap the wrapped effect, `Effect.timed`, inspect `rpc._tag` / `payload` / `headers`). Rewriting payloads or injecting synthetic tools bypasses server invariants, and if transforming middleware is genuinely needed it's a change to `RpcMiddleware`'s contract worth its own discussion.

*Written with the assistance of [Claude](https://claude.com/claude-code)*

naji247:
Watching requests passively isn't enough for what ours and other MCP integrations actually need. Our SDK's default features work by supplementing the tools generically before the client sees it — adding a context parameter to each tool's schema so the agent tells us why it's calling. And this isn't just us: MCP gateways like MetaMCP, IBM ContextForge, and Microsoft's mcp-gateway are whole products built around editing tool catalogs. They run as separate proxy processes because servers give them no way to do it natively.

The upcoming spec also changes the picture on "server invariants". The 2026-07-28 RC removes sessions from the protocol (SEP-2567), and its official advice for keeping state is: have a tool hand out an ID, and let the model pass it back as a normal argument on later calls. You could argue this is a great opportunity for transform capable middleware. 

If extending the existing RPC Middleware is what the core maintainers of this repo want, I can rework this PR to do that no problem :) To me it seemed a little low level, though.

I'd suggest taking a look at how the official Go MCP SDK accomplishes Middleware as a reference: https://github.com/modelcontextprotocol/go-sdk/blob/b426fbf6153b1be855a79a6d616bee0796ec75be/mcp/shared.go#L101

IMax153:
Hi @naji247,

Issue #1222, opened 2026-02-12
streaming doesn't work for Anthropic AI provider due to schema mismatch in message usage


skoshx:
### What version of Effect is running?

#86b1ff2

### What steps can reproduce the bug?

```ts
import { Config, Effect, Layer, Stream } from "effect"
import * as FetchHttpClient from "effect/unstable/http/FetchHttpClient"
import { Chat, LanguageModel } from "effect/unstable/ai"
import { AnthropicClient, AnthropicLanguageModel } from "@effect/ai-anthropic"

const program = Effect.gen(function* () {
  const chat = yield* Chat.empty
  const stream = chat.streamText({ prompt: "Say hello in one sentence." })
  yield* Stream.runForEach(stream, (part) =>
    Effect.log("part", part._tag),
  )
})

const AnthropicLive = AnthropicLanguageModel.layer({
  model: "claude-sonnet-4-5-20250929",
  config: { max_tokens: 256 },
}).pipe(
  Layer.provide(
    AnthropicClient.layerConfig({
      apiKey: Config.redacted("ANTHROPIC_API_KEY"),
    }),
  ),
  Layer.provide(FetchHttpClient.layer),
)

program.pipe(
  Effect.provide(AnthropicLive),
  Effect.runPromise,
).catch((e) => {
  console.error("FAILED:", e)
  process.exit(1)
})

```

Running the above yields:

```stderr
❯ bun repro.ts
FAILED: 1204 |  * ```
1205 |  *
1206 |  * @since 1.0.0
1207 |  * @category constructors
1208 |  */
1209 | export const make = params => new AiError(params);
                                     ^
error: AnthropicClient.createMessageStream: Invalid output: Missing key
  at [0]["data"]["message"]["usage"]["iterations"]
     module: "AnthropicClient",
     method: "createMessageStream",
       _tag: "AiError",
 ~effect/unstable/ai/AiError/AiError: "~effect/unstable/ai/AiError/AiError",

      at make (/.pnpm/effect@https+++pkg.pr.new+Effect-TS+effect-smol+effect@86b1ff2_032fdbbef8c03380603e9557e692e865/node_modules/effect/dist/unstable/ai/AiError.js:1209:31)
[...]

PR #2568, opened 2026-07-09
fix(ai-anthropic): preserve the string branch of the open `Model` enum

### LINKED ISSUE #2567: @effect/ai-anthropic: response decoding fails for any model id not in the vendored literal union (e.g. claude-opus-4-8)
## Summary

`packages/ai/anthropic/codegen.yml` applies a JSON patch that deletes the `{"type": "string"}` branch from Anthropic's `Model` schema, turning a deliberately **open** enum into a **closed** literal union. That schema is `$ref`'d by response bodies, so any request made with a model id that isn't in the vendored list fails to decode: the API call succeeds and is billed, then the client throws a `SchemaError` before the caller sees a single token.

Affects `claude-opus-4-8`, `claude-opus-4-7`, `claude-sonnet-5`, `claude-fable-5`, `claude-mythos-5` — all present in Anthropic's current published spec.

Verified on `4.0.0-beta.93` and on `main` (`c74babd0`) — both ship the same 22-id union, so there is no published version where this works.

## Symptom

```
SchemaError(Expected "claude-opus-4-6" | "claude-sonnet-4-6" | … | "claude-3-haiku-20240307",
got "claude-opus-4-8"
  at ["message"]["model"])
```

Streaming fails on the **first** SSE event (`message_start`), so the stream dies before any content arrives. `createMessage` fails when the response body is decoded. In both cases the request was already accepted by Anthropic and the input tokens are billed.

## Reproduction

No API key, no network — a pure decode of a payload the API really sends:

```ts
import * as Generated from "@effect/ai-anthropic/Generated"
import { Effect, Exit, Schema } from "effect"

const usage = {
  input_tokens: 4, output_tokens: 1, cache_creation_input_tokens: 0, cache_read_input_tokens: 0,
  cache_creation: { ephemeral_1h_input_tokens: 0, ephemeral_5m_input_tokens: 0 },
  inference_geo: null, server_tool_use: null, service_tier: null
}

const messageStart = (model: string) => ({
  type: "message_start",
  message: {
    id: "msg_01", type: "message", role: "assistant", model,
    content: [], stop_reason: null, stop_sequence: null,
    container: null, context_management: null, usage
  }
})

for (const id of ["claude-opus-4-6", "claude-opus-4-8", "claude-sonnet-5", "claude-fable-5"]) {
  const exit = Effect.runSyncExit(Schema.decodeUnknownEffect(Generated.BetaMessageStartEvent)(messageStart(id)))
  console.log(id.padEnd(18), Exit.isSuccess(exit) ? "decodes" : "DECODE FAILS")
}
```

```
claude-opus-4-6    decodes
claude-opus-4-8    DECODE FAILS
claude-sonnet-5    DECODE FAILS
claude-fable-5     DECODE FAILS
```

## Root cause

Anthropic's published spec (resolved from the `statsUrl` in `codegen.yml` itself) defines `components.schemas.Model` as an **open** enum:

```json
{
  "title": "Model",
  "anyOf": [
    { "type": "string" },
    { "const": "claude-sonnet-5", "x-stainless-nominal": false },
    { "const": "claude-opus-4-8",  "x-stainless-nominal": false },
    …17 branches total
  ]
}
```

The leading `{"type": "string"}` branch plus `x-stainless-nominal: false` is how Stainless encodes "these are the known ids, but any string is valid."

`packages/ai/anthropic/codegen.yml:8-9` removes exactly that branch:

```yaml
patches:
  # Remove string from Model schema - only keep literal model IDs
  - '[{"op":"remove","path":"/components/schemas/Model/anyOf/0"}]'
```

producing `packages/ai/anthropic/src/Generated.ts:2897`:

```ts
export const Model = Schema.Literals([
  "claude-opus-4-6", "claude-sonnet-4-6", /* …20 more */
])
```

`Model` is `$ref`'d from 16 sites in the spec, including these **response** schemas:

| Response schema | Impact |
| --- | --- |
| `BetaMessage.model` (required) | `Generated.ts:7740`; nested in `BetaMessageStartEvent` (`:8007`), the first event of every stream |
| `Message.model` | non-beta path |
| `CompletionResponse.model` | legacy completions |
| `BetaResponseFallbackHopInfo.model` | **the model that actually served the request** |
| `BetaMessageIterationUsage.model`, `BetaFallbackMessageIterationUsage.model`, `BetaAdvisorMessageIterationUsage.model` | per-iteration usage |

Decoding happens at `AnthropicClient.ts:286-302` (`MessageEvent` union → `Sse.decodeDataSchema` → `SchemaError` → `mapSchemaError`).

Worth noting the `BetaResponseFallbackHopInfo` row: with the fallback feature, a response can legitimately report a model id you did **not** request, so a closed response enum can break a client whose own requested model *is* in the vendored list.

### The request side was never the problem

`AnthropicLanguageModel.layer` / `.model` already accept any string (`AnthropicLanguageModel.ts:643,765`):

```ts
export const layer = (options: { readonly model: (string & {}) | Model; … })
```

and `createMessageStream` sends the payload through `HttpBody.jsonUnsafe` — it is never schema-**encoded**. Only the response decode is over-constrained.

## Why regenerating isn't the fix

The vendored literal list is stale, so a regen would *appear* to fix it. But Anthropic ships models faster than the spec propagates through Stainless → `.stats.yml` → an effect release, so every model launch reintroduces a hard failure for anyone who updates their model id before updating `@effect/ai-anthropic`. The class of bug is "the response schema is stricter than the API contract," and it should be fixed structurally.

Related: `codegen.yml`'s `replacements:` block already carries six separate "API may omit this field" fixes (`citations`, `server_tool_use`, `context_management`, `container`, `BetaUsage.iterations`/`speed`), and #1222 was the same failure mode. Response schemas generated as strict keep breaking in production, one field at a time.

## `@effect/ai-openai` already does this correctly

It hits the identical spec construct and emits the permissive form (`packages/ai/openai/src/Generated.ts:3250`):

```ts
export const ModelIdsShared = Schema.Union([
  Schema.String,
  Schema.Literals(["gpt-5.4", "gpt-5.4-mini", …])
])
```

So the generator already produces the right thing for `anyOf: [{type: string}, …consts]`. The Anthropic package is the outlier purely because of the one JSON patch.

## Proposed fix

Delete the patch at `codegen.yml:8-9` and regenerate. `Model` becomes `Schema.Union([Schema.String, Schema.Literals([…])])`, matching `ModelIdsShared`; responses decode for any model id, and requests are unaffected because they already accept any string.

The one cost is that `export type Model` widens to `string`, which collapses `(string & {}) | Model` in `AnthropicLanguageModel.layer` down to `string` — i.e. losing model-id autocomplete. So the permissive schema needs to be paired with a way to keep the literals in the type.

**Correction (2026-07-09):** an earlier revision of this issue claimed that losing autocomplete matched the DX `@effect/ai-openai` ships today. That was wrong. OpenAI's generated *alias* does collapse, but `OpenAiLanguageModel.ts:42-51` reads the literal branch back out of the union (`Generated.ModelIdsShared.members[1]`), so it gets permissive decoding **and** literal autocomplete. Anthropic derives its `Model` from `typeof Generated.Model.Type` (`AnthropicLanguageModel.ts:54`) instead, which is why the naive fix regresses it.

Happy to send a PR — will open one against this issue.

## Environment

- `@effect/ai-anthropic` `4.0.0-beta.93`; also reproduced against `main` @ `c74babd0`
- Spec: `statsUrl` → `anthropic-sdk-typescript@main/.stats.yml` → `anthropic-506a5ad7…62b54.yml`; `components.schemas.Model.anyOf[0]` is `{"type": "string"}` (the 16 `$ref` sites were enumerated programmatically from that document)

K-Mistele:
Closes #2567.

## Problem

`Model` is an **open enum** in Anthropic's OpenAPI spec — `anyOf: [{ type: "string" }, ...consts]`, with every const marked `x-stainless-nominal: false` — and it is `$ref`'d by response schemas: `Message.model`, `BetaMessage.model` (nested in `message_start`, the first event of every stream), `CompletionResponse.model`, `BetaResponseFallbackHopInfo.model`, and the three `*IterationUsage.model` schemas.

`codegen.yml` deleted the string branch, closing the enum. So any request made with a model id newer than the generated literals failed to decode:

```
SchemaError(Expected "claude-opus-4-6" | … | "claude-3-haiku-20240307", got "claude-opus-4-8"
  at ["message"]["model"])
```

Streaming died on the first SSE event; `createMessage` died decoding the body. In both cases the request had already been accepted and the input tokens billed. This affects `claude-opus-4-8`, `claude-opus-4-7`, `claude-sonnet-5`, `claude-fable-5` and `claude-mythos-5` — all present in the spec today.

Note the `BetaResponseFallbackHopInfo` row in particular: with the fallback feature a response can report a model id you did *not* request, so a closed response enum can break a client whose own requested model **is** in the vendored list.

The request side was never affected and still isn't: `AnthropicLanguageModel.layer`/`.model` already accept `(string & {}) | Model`, and `createMessageStream` sends the payload through `HttpBody.jsonUnsafe` — it is never schema-**encoded**.

## Fix

Drop the patch from `codegen.yml` and let `Model` generate as `Schema.Union([Schema.String, ...literals])`, so responses decode for any model id. `@effect/ai-openai` already generates its `ModelIdsShared` this way (`packages/ai/openai/src/Generated.ts:3250`); the Anthropic package was the outlier because of that one JSON patch.

**Note (post-review):** as @IMax153 points out below, this alone collapses `AnthropicLanguageModel`'s exported `Model` type to `string` (it is derived from `typeof Generated.Model.Type` at `AnthropicLanguageModel.ts:54`), which loses model-id autocomplete. `@effect/ai-openai` avoids that by reading the literal branch back out of the union (`OpenAiLanguageModel.ts:42-43`, `Generated.ModelIdsShared.members[1]`). That is **not** handled on this branch yet — the two candidate approaches are under discussion in the thread below.

I left a comment in `codegen.yml` where the patch was, so it doesn't get reintroduced.

## ⚠️ One judgment call I'd like a maintainer's read on

`Generated.ts` here contains **only the regenerated `Model` type and const**, not a full regeneration. I ran `node packages/tools/ai-codegen/src/bin.ts generate --provider anthropic` and then reduced the result to that one symbol, because a full regen against today's spec produces **+25,545 / −7,797 lines and ~4,000 new exported symbols** — the vendored `Generated.ts` is far behind the spec (the nightly `chore/ai-codegen-update` PR, #1147, has been open since January).

A full regen also currently **fails this package's own tests**:

```
InvalidOutputError: Missing key at ["stop_details"]
```

which is another instance of the same over-strict-response-schema class of bug (cf. the six "API may omit this field" entries already in `codegen.yml`'s `replacements:`, and #1222).

So bundling the spec bump into this fix would have made the fix unreviewable *and* red. The `Model` block in this PR is byte-for-byte what the generator emits for that symbol — happy to switch to a full regen (and fix the `stop_details` fallout separately or here) if you'd rather. Just say which you prefer.

One visible consequence of taking the generator's current `Model` output: the literal list gains the new ids and drops several retired ones (`claude-3-7-sonnet-latest`, `claude-sonnet-4-20250514`, …). That is **not** a breaking change now precisely because of the restored `string` branch — the existing test that uses `claude-sonnet-4-20250514` still passes.

## Tests

Added a regression test to `packages/ai/anthropic/test/AnthropicLanguageModel.test.ts` that streams a response whose `message_start` carries an unknown model id, and asserts both that the text arrives and that the id survives decoding into `response-metadata.modelId`.

Verified it is a real regression test: on `main`'s `Generated.ts` it fails with the exact production error (`Expected "claude-opus-4-6" | … , got "claude-not-a-known-model-id"`); with the fix, it passes.

## Validation

`AGENTS.md` lists `pnpm check:tsgo` for code changes, but that script doesn't exist in this checkout — I ran `pnpm check` instead. Everything below is green:

- `pnpm lint-fix` → 0 warnings, 0 errors
- `pnpm test packages/ai/anthropic/test/AnthropicLanguageModel.test.ts` → 5 passed
- `pnpm check` (package and repo root) → clean

Changeset included (`@effect/ai-anthropic`: patch).


packages/ai/anthropic/src/Generated.ts
```diff
@@ -2872,55 +2872,67 @@ export const BetaStopReason = Schema.Literals([
...
-  "claude-sonnet-4-5",
-  "claude-sonnet-4-5-20250929",
-  "claude-opus-4-0",
-  "claude-opus-4-20250514",
-  "claude-4-opus-20250514",
-  "claude-opus-4-1-20250805",
-  "claude-3-opus-latest",
-  "claude-3-opus-20240229",
-  "claude-3-haiku-20240307"
+export const Model = Schema.Union([
```
IMax153, line 2892: This solution will not work. 

A `Schema.Union` of `Schema.String` + a literal will simply collapse on the type level to `string`.

Instead, the `Model` schema must remain the union of literals to preserve the type. A better solution would be to do `Schema.Union([Schema.String, Model])` where the model type is used within the generated schemas.

packages/ai/anthropic/src/Generated.ts
```diff
@@ -2872,55 +2872,67 @@ export const BetaStopReason = Schema.Literals([
   "model_context_window_exceeded"
 ])
 export type Model =
```
IMax153, line 2874: This solution will not work.

A `string | <literal>` collapses on the type level to `string`.

To get around this in other places in our AI sdks, we use the trick:

```ts
type Model = (string & {}) | <literal>
```

which preserves the literal but also allows strings.

IMax153 (changes requested):

K-Mistele:
Thanks, Max. I dug in a bit more — looks like there are two options here.

**Option A — normalize the open enum in codegen, then do what openai does.**

Anthropic's spec puts a `description` on each model const, so the generator emits a separate annotated `Schema.Literal` per id instead of collapsing them into one `Schema.Literals([...])` the way openai's `ModelIdsShared` comes out. If the `onEnter` hook in `packages/tools/ai-codegen/src/Generator.ts` rewrote `anyOf: [{ type: "string" }, ...consts]` into `anyOf: [{ type: "string" }, { enum: [...] }]`, `Model` would generate in the same shape as openai's, and `AnthropicLanguageModel` could use the same trick as `OpenAiLanguageModel`:

```ts
const ModelIds = Generated.Model.members[1]
export type Model = typeof ModelIds.Encoded
```

`(string & {}) | Model` then keeps the literals and still accepts any string.

*Pros:* both providers end up with the same shape; ~10 lines in the hook plus one line in `AnthropicLanguageModel.ts`; it's generic, so any other provider with an open enum picks it up on its next regen.
*Cons:* the per-const descriptions drop off the generated `Model` (openai has none today), and it touches shared codegen, so other providers' output shifts whenever they're next regenerated.

**Option B — keep `Model` as the literal union and wrap at the use sites**, i.e. `Schema.Union([Schema.String, Model])` wherever the generated schemas reference it, as you suggested.

*Pros:* `typeof Generated.Model.Type` stays the literal union, so `AnthropicLanguageModel.ts` doesn't change at all, and the per-const descriptions survive.
*Cons:* the generator has to know at reference time that a component is an open enum, and the union/literal emission lives in `effect/SchemaRepresentation`, so it's a deeper change; anthropic and openai also end up with different shapes for the same spec construct.

I lean slightly toward A for the parity, but you know the codegen far better than I do — which would you like me to go with? Either way I'll add a type-level test pinning `Model` so it can't quietly regress.

Also still open from the description: do you want the full `Generated.ts` regen in this PR (+25k/−8k, currently failing on `Missing key at ["stop_details"]`), or the scoped `Model`-only hunk that's on the branch now?

IMax153:
I think Option A sounds fine, but I'd need to see / review the code to be sure.

And regarding regenerating the generated schemas, I think that can be tackled separately - let's keep this PR focused on the `Model` type / schema.

K-Mistele:
Pushed Option A in 9a4f9e4.

- `packages/tools/ai-codegen/src/Generator.ts` — the `onEnter` hook now collapses an open enum's const branches into a single `enum` branch, so `anyOf: [{ type: "string" }, { const: "a" }, ...]` generates `Schema.Union([Schema.String, Schema.Literals(["a", ...])])` instead of one `Schema.Literal` member per id.
- `packages/ai/anthropic/src/AnthropicLanguageModel.ts` — `Model` now reads the literal branch back out of the union, the same way `OpenAiLanguageModel` does with `Generated.ModelIdsShared.members[1]`:

```ts
export type Model = (typeof Generated.Model)["members"][1]["Encoded"]
```

  (`Generated` is a type-only import here, so this stays in type position rather than adding a runtime import.)

Responses decode for any model id, `(string & {}) | Model` still accepts custom ids, and the known ids survive for autocomplete.

**Pinned with a type-level assertion** in the test file — reverting `Model` to the old `typeof Generated.Model.Type` derivation fails `pnpm check` with `error TS2578: Unused '@ts-expect-error' directive`, so this can't silently regress.

**On blast radius**, since the hook is generic: I scanned the provider specs for nodes it would rewrite (`anyOf[0]` is a bare `{ type: "string" }` and every remaining branch is a string `const`).

| spec | matching nodes |
| --- | --- |
| openai | 0 |
| openrouter | 0 |
| anthropic | 2 — `Model`, `BetaManagedAgentsModel` |

So it's inert for openai/openrouter (their specs use `enum` already). `BetaManagedAgentsModel` doesn't exist in the committed generated output yet, so it only shows up whenever the full regen lands — which per your note I've left out of this PR. `Generated.ts` here still contains just the `Model` symbol, regenerated.

Unrelated, but worth flagging since I tripped over it: `--provider openai` can't be regenerated on `main` right now — `codegen.yaml`'s `[{"op":"remove","path":"/components/schemas/WebSearchActionSearch/required/1"}]` fails with `Array index out of bounds`. Reproduces without any of my changes.

Validation: `pnpm lint-fix` clean, `pnpm test packages/ai/anthropic/test/AnthropicLanguageModel.test.ts` 6 passed, `pnpm check` clean at the package and repo root. (`pnpm check:tsgo` from `AGENTS.md` doesn't exist in this checkout.)

IMax153 (approved):

PR #2570, opened 2026-07-09
Fix transactional cluster request handling

gptguy:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`ClusterSchema.WithTransaction` currently wraps `RpcServer.write`, but `write` forks the request handler and returns immediately. The storage transaction can therefore commit before the handler and durable reply complete.

This change:

- adds an awaited scalar request path to `RpcServer` while preserving ordinary fire-and-forget and streaming behavior
- keeps transactional entity handlers and durable replies inside the same message-storage transaction
- defers in-process reply delivery until the transaction commits
- releases interrupted transactional request claims so they can be reconstructed
- adds regression coverage for transaction lifetime, reply delivery, interruption, and existing `write` semantics

Validation:

- `npx --yes pnpm@10.17.1 lint-fix`
- `npx --yes pnpm@10.17.1 test packages/effect/test/rpc/RpcServer.test.ts packages/effect/test/cluster/MessageStorage.test.ts packages/effect/test/cluster/Sharding.test.ts packages/effect/test/cluster/ClusterWorkflowEngine.test.ts`
- `npx --yes pnpm@10.17.1 check`

## Related

- None.

### DIFF
--- a/.changeset/fix-cluster-transactional-rpc.md
+++ b/.changeset/fix-cluster-transactional-rpc.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Keep transactional cluster request handlers and their durable replies inside the same storage transaction.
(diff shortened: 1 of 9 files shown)


tim-smart:
Went for a different approach. Thanks for the report :)

https://github.com/Effect-TS/effect-smol/pull/2607

PR #2573, opened 2026-07-10
Prevent MCP tool errors from exposing internals

ebramanti:
## Type
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### Summary

`McpServer.registerToolkit` previously converted every failed tool execution into an MCP `CallToolResult` whose text was `Cause.pretty(cause)`. That diagnostic renderer includes stack traces, nested Cause details, Effect implementation frames, and absolute source paths, so server internals crossed the MCP protocol boundary.

This change sanitizes registered-tool failure responses while preserving the validation feedback a MCP client needs to correct invalid arguments and retry.

### Why

Tool-originated errors belong in `CallToolResult` so an LLM can observe the failure, correct its arguments, and retry. Server diagnostic representations do not belong in that result. In particular, `Cause.pretty` is suitable for logs and debugging, not client-visible protocol payloads.

For an optional string field, passing `null` should continue to produce actionable feedback such as:

```text
Invalid parameters for tool 'example': Expected string | undefined, got null
  at ["signature"]
```

It should not append an Effect error type, stack trace, nested `[cause]` rendering, or local source paths.

### Behavior

| Outcome                                                        | MCP response                                          |
| -------------------------------------------------------------- | ----------------------------------------------------- |
| Successful tool call                                           | Existing successful `CallToolResult`                  |
| `ToolParameterValidationError`                                 | `isError: true` with `reason.message`                 |
| Declared handler failure represented by a schema-valid `Error` | `isError: true` with `Error.message`                  |
| Other `AiError`                                                | `isError: true` with a generic internal-error message |
| Undeclared or non-public handler failure                       | `isError: true` with a generic internal-error message |
| Defect, interruption, empty Cause, or residual Cause           | `isError: true` with a generic internal-error message |
| Unknown tool or malformed MCP request                          | Existing MCP protocol error                           |

## Related

None.

### DIFF
--- a/.changeset/quiet-tools-smile.md
+++ b/.changeset/quiet-tools-smile.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Prevent MCP tool failures from exposing Cause rendering, stack traces, and internal paths while preserving actionable validation messages.
--- a/packages/effect/src/unstable/ai/McpServer.ts
+++ b/packages/effect/src/unstable/ai/McpServer.ts
@@ -39,6 +39,7 @@ import type * as RpcGroup from "../rpc/RpcGroup.ts"
 import * as RpcMessage from "../rpc/RpcMessage.ts"
 import * as RpcSerialization from "../rpc/RpcSerialization.ts"
 import * as RpcServer from "../rpc/RpcServer.ts"
+import * as AiError from "./AiError.ts"
 import {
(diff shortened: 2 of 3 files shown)


IMax153:
Hi @ebramanti,

PR #2583, opened 2026-07-10
Handle large millis passed to sleep

MrGovindan:
## Type
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Fixes how Clock.sleep handles large duration

### DIFF
--- a/packages/effect/src/internal/effect.ts
+++ b/packages/effect/src/internal/effect.ts
@@ -5838,8 +5838,8 @@ class ClockImpl implements Clock.Clock {
     const millis = Duration.toMillis(duration)
     if (millis <= 0) return yieldNow
     return callback((resume) => {
-      if (millis > MAX_TIMER_MILLIS) return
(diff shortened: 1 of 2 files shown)


tim-smart (dismissed):

packages/effect/src/internal/effect.ts
```diff
@@ -5838,8 +5838,8 @@ class ClockImpl implements Clock.Clock {
     const millis = Duration.toMillis(duration)
     if (millis <= 0) return yieldNow
     return callback((resume) => {
-      if (millis > MAX_TIMER_MILLIS) return
-      const handle = setTimeout(() => resume(void_), millis)
+      const continuation = millis > MAX_TIMER_MILLIS ? this.sleep(millis - MAX_TIMER_MILLIS) : void_
```
tim-smart: This will need to be wrapped with Duration.millis

PR #2586, opened 2026-07-11
Fix stripped declaration for SchemaError type identifier

codewithkenzo:
## Summary

When internal declarations are stripped, `SchemaError` retained a reference to `SchemaErrorTypeId` without emitting its declaration.

## Fix

Keep `SchemaErrorTypeId` in generated declarations.

## Tests

- `pnpm test-types packages/effect/typetest/schema/Schema.tst.ts`
- `pnpm --dir packages/effect test --run test/schema/Schema.test.ts`
- `pnpm check`
- `pnpm build` with `stripInternal` enabled
- `pnpm jsdocs`
- `pnpm exec dprint check packages/effect/src/internal/schema/schema.ts packages/effect/typetest/schema/Schema.tst.ts`
- `pnpm exec oxlint packages/effect/src/internal/schema/schema.ts packages/effect/typetest/schema/Schema.tst.ts`

### DIFF
--- a/.changeset/fix-schema-error-type-id-declaration.md
+++ b/.changeset/fix-schema-error-type-id-declaration.md
@@ -1,0 +1,5 @@
+---
+"effect": patch
+---
+
+Fix published `SchemaError` declarations by retaining `SchemaErrorTypeId`.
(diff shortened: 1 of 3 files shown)


codewithkenzo:
I realize this is a quick fix and may not be preferred. If maintainers prefer keeping `SchemaErrorTypeId` internal, I can update the PR so the public declaration uses literal types instead and add a stripInternal regression test.

gcanti:
Thanks for the report. I moved `SchemaError` to internal during a refactoring, and then I forgot to move it into a proper module.

PR #2612, opened 2026-07-13
Fix durable race result replay

gptguy:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Fix replay of successful `DurableDeferred.raceAll` results.

`WorkflowEngine.deferredResult` already returns a decoded `Exit`. The replay path used `Effect.flatten`, causing the successful payload to be executed as an Effect and fail with `Fiber.runLoop: Not a valid effect`.

Yield the persisted `Exit` directly, matching `DurableDeferred.await`.

The regression test forces a second workflow evaluation after the race winner is persisted, covering the replay path.

Validation:

- `pnpm lint-fix`
- `pnpm test packages/effect/test/cluster/ClusterWorkflowEngine.test.ts`
- `pnpm check`

## Related

- Related Issue: None
- Closes: None

### DIFF
--- a/packages/effect/src/unstable/workflow/DurableDeferred.ts
+++ b/packages/effect/src/unstable/workflow/DurableDeferred.ts
@@ -280,7 +280,7 @@ export const raceAll = <
     const engine = yield* EngineTag
     const exit = yield* engine.deferredResult(deferred)
     if (Option.isSome(exit)) {
-      return yield* exit.value as Exit.Exit<any, any>
+      return yield* exit.value
     }
     return yield* into(
       Effect.raceAll(options.effects),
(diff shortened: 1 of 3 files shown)


packages/effect/src/unstable/workflow/DurableDeferred.ts
```diff
@@ -280,7 +280,7 @@ export const raceAll = <
     const engine = yield* EngineTag
     const exit = yield* engine.deferredResult(deferred)
     if (Option.isSome(exit)) {
-      return yield* Effect.flatten(exit.value) as Effect.Effect<any, any, any>
+      return yield* exit.value as Exit.Exit<any, any>
```
tim-smart:
```suggestion
      return yield* exit.value
```

